PR #43, opened 2020-05-14
Handle new paysFee enum

maciejhirsz:
+ **[breaking]** `paysFee` on `system.ExtrinsicSuccess`/`system.ExtrinsicFailed` events can now be a string `"Yes"` or `"No"` on newer runtimes.
+ To provide a stable API, all extrinsics now have a `paysFee` field directly on them that is guaranteed to be `null` (if no success or failed event is present) or a boolean.
+ Fixes the transaction fee amount (#41).
+ Bumped `@polkadot/api` to `1.14.0-beta.20`.

### DIFF
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -86,6 +86,7 @@ export default class ApiHandler {
 				info,
 				events: [] as SantiziedEvent[],
 				success: defaultSuccess,
+				paysFee: false, // override to true if `system.ExtrinsicSuccess` event is present
 			};
 		});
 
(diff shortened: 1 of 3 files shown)


src/ApiHandler.ts
```diff
@@ -86,6 +86,7 @@ export default class ApiHandler {
 				info,
 				events: [] as SantiziedEvent[],
 				success: defaultSuccess,
+				paysFee: false, // override to true if `system.ExtrinsicSuccess` event is present
```
joepetrowski, line 89: Actually, this is part of `DispatchInfo`, which emits as part of `system.ExtrinsicFailure` too. So you should pay a fee even if it fails. Just certain dispatchables have weight but don't charge a fee. So it's a property of the extrinsic, not its success or failure.

PR #46, opened 2020-05-16
Return txVersion with artifacts

joepetrowski:


src/ApiHandler.ts
```diff
@@ -328,20 +331,38 @@ export default class ApiHandler {
...
+			// Swap metadata if tx version is different. This block of code should never execute, as
+			// specVersion always increases when txVersion increases, but specVersion may increase
+			// without an increase in txVersion. Defensive only.
+			if (!this.txVersion.eq(blockTxVersion)) {
+				this.txVersion = blockTxVersion;
+				const meta = await api.rpc.state.getMetadata(hash);
+				const chain = await api.rpc.system.chain();
+
+				api.registry.register(
+					getSpecTypes(api.registry, chain, version.specName, blockSpecVersion)
```
joepetrowski, line 357: Should this be `blockTxVersion`? I think no, but want to make sure.

amaurymartiny: No, `getSpecTypes` only checks spec version, and doesn't care about tx version.

amaurymartiny: Actually, I would remove the `api.registry.register(..)` part, as it's really a no-op, since spec version doesn't change in this case.

Metadata might change, so we can keep the `api.registry.setMetadata(meta);` line.

amaurymartiny: Would also add a `console.warn('txVersion bumped without specVersion bumping')`, as as you said, this code should actually never be executed

PR #47, opened 2020-05-17
Prettier & other minor changes

danforbes:
Prettier, move `parseNumber` from `main.ts` to `utils.ts`, fix a typo, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "tsc",
+    "build": "npm run prettier & tsc",
+    "prettier": "./node_modules/.bin/prettier --write .",
     "start": "tsc && node ./build/main.js",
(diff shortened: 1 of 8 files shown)


joepetrowski:
Since this is mostly partnered up with txwrapper, I think we should use the same style, [eslint](https://github.com/paritytech/txwrapper/blob/master/package.json#L19). At a minimum, let's use single quotes.
```
"singleQuote": true,
"trailingComma": "all",
```

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "tsc",
+    "build": "npm run prettier & tsc",
```
joepetrowski, line 7: I don't think this should be silently implied in the build command.

danforbes:
Closing in favor of the changes that will come along with #61 :rocket:

PR #50, opened 2020-05-19
Update fee calculation logic

### LINKED ISSUE #45: Update fees for refund and polynomial calculation
We need to update fee calculation for two changes:

First, some functions now return weight consumed and issue a refund if the weight consumed was less than that charged at the beginning. Update sidecar to return the actual fee paid in case of a refund. https://github.com/paritytech/substrate/pull/6024

Second, weight calculation will be based on a polynomial, the coefficients of which will be constants in the runtime. Calculate the fee based on runtime constants, the fee multiplier, encoded length, and weight consumed. The logic will look roughly like:
```
// constants
Get byte fee
Get polynomial coefficients (integer, fraction, negative, degree)
Get extrinsic base weight

// per-block variables
Get FeeMultiplier (emits in `onInitialize.transactionPayment`)

// per-transaction
Get encoded length
Get weight consumed

fee = byteFee * encodedLength + FeeMultiplier * weightFee(weightConsumed)
```

danforbes:
Closes #45


src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+		const extrinsicBaseWeight = api.consts.system.extrinsicBaseWeight;
+		const coefficients = api.consts.transactionPayment.weightToFee[0];
+		const coeffFrac = coefficients.coeff_frac;
+		const coeffInt = coefficients.coeff_integer;
+		
+		// TODO: generalize for terms that may have more degrees & may be negative
+		const degree = coefficients.degree;
+		const negative = coefficients.negative;
+
+		for (let idx = 0; idx < block.extrinsics.length; ++idx) {
```
athei, line 138:
```suggestion
		for (extrinsic record of extrinsics) {
```

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+			const extrinsic = block.extrinsics[idx];
+			if (!extrinsic.isSigned) {
```
athei, line 140:
```suggestion
			if (!extrinsic.paysFee) {
```

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
 
+		const byteFee = api.consts.transactionPayment.transactionByteFee;
+		const extrinsicBaseWeight = api.consts.system.extrinsicBaseWeight;
+		const coefficients = api.consts.transactionPayment.weightToFee[0];
+		const coeffFrac = coefficients.coeff_frac;
+		const coeffInt = coefficients.coeff_integer;
+		
+		// TODO: generalize for terms that may have more degrees & may be negative
+		const degree = coefficients.degree;
+		const negative = coefficients.negative;
```
athei, line 136: Please remove unused code. I suggest changing the comment to something like this: "We assume that the polynomial has one positive term of degree one."

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+			const extrinsic = block.extrinsics[idx];
+			if (!extrinsic.isSigned) {
+				continue;
+			}
+
+			try {
+				if (api.runtimeVersion.specName.toString() === 'kusama') {
+					extrinsics[idx].info = await api.rpc.payment.queryInfo(extrinsic.toHex(), parentHash);
+				} else {
+					const parsedXt = extrinsics[idx];
```
athei, line 148: Just remove that line and replace all occurrences of `parsedXt` by `extrinsic`.

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+
+		for (let idx = 0; idx < block.extrinsics.length; ++idx) {
+			const extrinsic = block.extrinsics[idx];
+			if (!extrinsic.isSigned) {
+				continue;
+			}
+
+			try {
+				if (api.runtimeVersion.specName.toString() === 'kusama') {
+					extrinsics[idx].info = await api.rpc.payment.queryInfo(extrinsic.toHex(), parentHash);
```
athei, line 146:
```suggestion
					extrinsic.info = await api.rpc.payment.queryInfo(extrinsic.toHex(), parentHash);
```

I just assume everything is a reference. Not a {java, type, ...}script expert. Maybe you can make use of an continue here to safe one level of indentation for the rest of the function.

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+						continue;
+					}
+
+					const weight = weightInfo.weight;
+					const encodedLength = extrinsic.encodedLength;
+					const multiplier = await api.query.transactionPayment.nextFeeMultiplier.at(parentHash);
+
+					const multiplied = multiplier.mul(weight.mul(coeffFrac)
+                                       .add(weight.mul(coeffInt))
+                                       .add(byteFee).muln(encodedLength));
```
athei, line 176: That is barely readable. Parentheses that start on the same line should close on the same line. I also think that it is wrong (the whole sum was multiplicated with  `encodedLength`)
```suggestion
					const multiplied = multiplier.mul(
					    weight.mul(coeffFrac)
                                               .add(weight.mul(coeffInt))
                                               .add(byteFee.muln(encodedLength))
                                        );
```

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+						continue;
+					}
+
+					const completedData = completedEvent.data;
+					if (!completedData) {
+						parsedXt.info = { error: 'Unable to fetch fee info' };
+						continue;
+					}
+
+					const weightInfo = completedData[completedData.length - 1] as DispatchInfo;
```
amaurymartiny, line 164: Your code is correct, but I would love to see a comment on this line, it's not straightforward imo.
In case of `ExtrinsicSuccess`, the data is `[DispatchInfo]`. In case of `ExtrinsicFailed`, the data is `[DispatchError, DispatchInfo]`. That's why we take the last element of the array.

https://github.com/polkadot-js/api/blob/e3abb9511169022f9feeac432b0397f1b377aa79/packages/metadata/src/Metadata/v11/static-substrate.json#L399-L418

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+		const negative = coefficients.negative;
+
+		for (let idx = 0; idx < block.extrinsics.length; ++idx) {
+			const extrinsic = block.extrinsics[idx];
+			if (!extrinsic.isSigned) {
+				continue;
+			}
+
+			try {
+				if (api.runtimeVersion.specName.toString() === 'kusama') {
```
amaurymartiny, line 145: We're making the assumption that sidecar is only running against kusama and polkadot, right?

src/main.ts
```diff
@@ -42,7 +42,20 @@ function parseNumber(n: string): number {
 }
 
 async function main() {
-	const api = await ApiPromise.create({ provider: new WsProvider(WS_URL) });
+	const api = await ApiPromise.create({
+		provider: new WsProvider(WS_URL),
+		types: {
+			Address: 'AccountId',
+			LookupSource: 'AccountId',
+			WeightToFeeCoefficient: {
```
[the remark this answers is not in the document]
amaurymartiny: (@danforbes I don't know if you've done this procedure already, if not, I can walk you through it or do it for you)

src/ApiHandler.ts
athei: The only thing that is `Perbill` is `coeffFrac`. The rest are integers.

src/ApiHandler.ts
joepetrowski: No, Kusama doesn't have the polynomial WeightToFee, so we still have to (incorrectly) use `queryInfo`. But since this code is mostly generic (assuming 1 coefficient of degree 1) then it will work on any chain built on Substrate master, including Polkadot. Once Kusama gets a runtime upgrade to latest Substrate, then we can pull this out.

athei: > (assuming 1 coefficient of degree 1)

It is not difficult to remove this assumption, though. Just a loop over all coefficients. But that can be a followup PR in order to get this working asap.

src/ApiHandler.ts
```diff
@@ -139,6 +125,69 @@ export default class ApiHandler {
...
+					const multiplier = await api.query.transactionPayment.nextFeeMultiplier.at(parentHash);
+
+					const multiplied = multiplier.mul(weight.mul(coeffFrac)
+                                       .add(weight.mul(coeffInt))
+                                       .add(byteFee).muln(encodedLength));
+
+					parsedXt.info = {
+						weight: weight.toString(),
+						class: weightInfo.class.toString(),
+						partialFee: multiplied.add(extrinsicBaseWeight.mul(coeffFrac))
```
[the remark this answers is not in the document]
joepetrowski: Dispatch info comes out in the ExtrinsicSuccess event, we are matching the data format of `queryInfo`. It's not a very useful format, but it is what users are set up to parse, so we are forcing the new format into the old to avoid last-minute breaking changes. But, once Kusama is upgraded and things are a bit more stable, we'll make a better interface with plenty of warning to users about the breaking change in response.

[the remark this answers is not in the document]
joepetrowski: `query_info` and `partial_fee` are probably the two worst named things in Substrate.

src/ApiHandler.ts
joepetrowski: Correct, I was talking more about removing the `kusama` bit. Once Kusama gets an upgrade and we loop over all coefficients here, it will be a clean solution.

src/ApiHandler.ts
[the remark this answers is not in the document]
joepetrowski: I just stumbled upon `RuntimeDispatchInfo` in the Substrate codebase, sorry, my response was nonsense.

src/ApiHandler.ts
danforbes: Added a comment in an attempt to clarify this.

src/main.ts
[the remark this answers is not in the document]
danforbes: I will give it a try!

src/ApiHandler.ts
danforbes: I still needed access to the "raw" extrinsic from the block in some cases so I was not able to figure out how to adopt this suggestion.

src/ApiHandler.ts
danforbes: Great suggestion about using `continue` to save indentation :pray:

src/ApiHandler.ts
danforbes: +1 for readable code...sorry about that. I do not claim to be a math wizard, so please let me know if I have accurately implemented your calculations :pray:

src/ApiHandler.ts
athei: the only place I can spot where you use it is where you read the `encodedLenght` of the extrinsic. I am surprised that this is not contained in the result. Maybe just add this. Might be an interesting information anyways.

danforbes: I also use it to get the hex of the extrinsic when I follow the old code path for Kusama backward compatibility.

athei: Far enough. Then just leave it the way it is. I am no javascript loop expert :D

danforbes: That makes two of us!

src/main.ts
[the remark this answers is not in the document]
danforbes: Think I could use a bit of help with this here https://github.com/polkadot-js/api/pull/2296

src/ApiHandler.ts
```diff
@@ -139,6 +125,72 @@ export default class ApiHandler {
...
+
+			try {
+				if (api.runtimeVersion.specName.toString() === 'kusama') {
+					extrinsics[idx].info = api.createType('RuntimeDispatchInfo',await api.rpc.payment.queryInfo(block.extrinsics[idx].toHex(), parentHash));
+					continue;
+				}
+
+				const xtEvents = extrinsics[idx].events;
+				const success = 'system.ExtrinsicSuccess';
+				const failure = 'system.ExtrinsicFailed';
```
joepetrowski, line 151: I doubt it's much of a performance change, but maybe move these outside the for loop?

src/ApiHandler.ts
```diff
@@ -139,6 +125,72 @@ export default class ApiHandler {
...
+				// both ExtrinsicSuccess and ExtrinsicFailed events have DispatchInfo types as their final arg
+				const weightInfo = completedData[completedData.length - 1] as DispatchInfo;
+				if (!weightInfo.weight) {
+					extrinsics[idx].info = { error: 'Success or failure event for extrinsic does not specify weight' };
+					continue;
+				}
+
+				const weight = weightInfo.weight;
+				const encodedLength = block.extrinsics[idx].encodedLength;
+				const multiplier = await api.query.transactionPayment.nextFeeMultiplier.at(parentHash);
```
joepetrowski, line 173: This can (and should, since it's an RPC call) also come out of the for loop as it's constant for all extrinsics in the block.

src/main.ts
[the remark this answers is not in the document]
joepetrowski: Bump to `1.15.0-beta.8` should fix this.

src/ApiHandler.ts
danforbes: I would prefer to keep these inside the loop close to where they are being used (this is an important principle of writing good code). Declaring a variable should not have an appreciable effect on performance in this case. This change was made to address the line length issues that Alexander brought up...normally I would have just defined these inline and not pulled them out into a variable.

joepetrowski: But you're also using them in [101](https://github.com/paritytech/substrate-api-sidecar/pull/50/files#diff-fdb5fe31b3fb59e594f5a04dabf5c38eR101) and [105](https://github.com/paritytech/substrate-api-sidecar/pull/50/files#diff-fdb5fe31b3fb59e594f5a04dabf5c38eR105). Why not declare them once and use them consistently?

danforbes: Fair enough...I went ahead and changed that. To answer your question, though, in order to keep the diff as small as possible, I was staying focused on the code blocks that I was writing. From my POV, it's better to keep PRs like this focused on the specific feature being implemented or bug being fixed. There should be _separate_ PRs that go back and refactor the existing code. But that's just my $0.02.

src/ApiHandler.ts
```diff
@@ -124,20 +130,79 @@ export default class ApiHandler {
...
+						error: 'Success or failure event for extrinsic does not specify weight'
+					};
+
+					continue;
+				}
+
+				const weight = weightInfo.weight;
+				const encodedLength = block.extrinsics[idx].encodedLength;
+
+				const weightedCoeffFrac = weight.mul(coeffFrac).div(new BN(1e9));
```
amaurymartiny, line 193: `div(new BN(1e9)) === divn(1e9)`, it's shorter, and I think also more performant.

danforbes: Unfortunately `bn.js` does not like this :tired_face:

src/ApiHandler.ts
```diff
@@ -124,20 +130,81 @@ export default class ApiHandler {
...
-				// we need to calculate fees based on the metadata constants and fee multiplier.
-				// https://github.com/paritytech/substrate-api-sidecar/issues/45
-				extrinsics[idx].info = api.createType(
-					'RuntimeDispatchInfo',
-					await api.rpc.payment.queryInfo(block.extrinsics[idx].toHex(), parentHash)
-				);
+				if (api.runtimeVersion.specName.toString() === 'kusama') {
+					extrinsics[idx].info = api.createType(
+						'RuntimeDispatchInfo',
+						await api.rpc.payment.queryInfo(block.extrinsics[idx].toHex(), parentHash)
```
amaurymartiny, line 156: ah my bad, there was a small misunderstanding.

`api.rpc.payment.queryInfo()` will return a RuntimeDispatchInfo JS class, so there's no need to api.createType here, it's a no-op.

(continued below)

src/ApiHandler.ts
```diff
@@ -124,20 +130,81 @@ export default class ApiHandler {
...
+				const weightedCoeffInt = weight.mul(coeffInt);
+				const byteFee = perByteFee.muln(encodedLength);
+				const multiplied = multiplier.mul(weightedCoeffFrac.add(weightedCoeffInt).add(byteFee));
+
+				const coeffFracWeight = extrinsicBaseWeight.mul(coeffFrac);
+				const coeffIntWeight = extrinsicBaseWeight.mul(coeffInt);
+
+				extrinsics[idx].info = {
+					weight: weight.toString(),
+					class: weightInfo.class.toString(),
```
amaurymartiny, line 205: my comment was to put here:

```ts
extrinsics[idx].info = api.createType('RuntimeDispatchInfo', {wieight, class, partialFee})
```

so that in both cases, it returns the same `RuntimeDispatchInfo` class.

src/ApiHandler.ts
```diff
@@ -124,20 +130,81 @@ export default class ApiHandler {
 			}
 		}
 
+		const perBill = new BN(1e9);
+		const perByteFee = api.consts.transactionPayment.transactionByteFee.mul(perBill);
+		const extrinsicBaseWeight = api.consts.system.extrinsicBaseWeight.mul(perBill);
+		const coefficients = api.consts.transactionPayment.weightToFee[0];
+		const coeffFrac = coefficients.coeffFrac.mul(perBill);
```
amaurymartiny, line 137: I believe this is the one that doesn't need `.mul(perBill)`

src/ApiHandler.ts
danforbes: Very nice I was wondering about this :pray:

src/ApiHandler.ts
```diff
@@ -124,20 +130,91 @@ export default class ApiHandler {
 			}
 		}
 
+		const fixed128 = new BN(10).pow(new BN(18))
+		const perBill = new BN(1e9);
+		const perByteFee = api.consts.transactionPayment.transactionByteFee;
+		const extrinsicBaseWeight = api.consts.system.extrinsicBaseWeight;
+		// Here we assume that `weightToFee` Vec only has 1 element, which is of
+		// degree 1. FIXME Implement polynomial.
+		const coefficients = api.consts.transactionPayment.weightToFee[0];
```
maciejhirsz, line 139: I'm getting `Internal Error: TypeError: Cannot read property '0' of undefined` here. Not latest binary, but latest block.

athei: This field is a recent addition. You need to update your node.

maciejhirsz: @athei still having the same issue on stable Kusama 0.7.33 release, fully synced.

athei: The change is not tagged yet. You need to build master.

amaurymartiny:
Is there a reason why we cannot do the calculations with BNs?

athei:
You can do that sure. But just look at the `PerBill` Rust implementation. Good look replicating that so that it works in all corner cases.

Doing that in wasm is just way easier and faster.

maciejhirsz:
Riot is acting up and it seems I can't join channels. WASM is fine by me.

amaurymartiny:
@athei makes sense. Note that https://github.com/polkadot-js/wasm exists, the idea was to have one repo for wasm so that we don't replicate rust->wasm+js_glue_code tooling. If this is something not sidecar-specific, it might be a good idea to put it there.

athei:
> Riot is acting up and it seems I can't join channels. WASM is fine by me.

I cannot accept nor decline your DM invitation. Something is seriously wrong. Try restarting :D

> @athei makes sense. Note that https://github.com/polkadot-js/wasm exists, the idea was to have one repo for wasm so that we don't replicate rust->wasm+js_glue_code tooling. If this is something not-sidecar-specific, it might be a good idea to put it there.

How would I include the wasm blob into the sidecar then? Via npm.js?

amaurymartiny:
> Via npm.js?

Yes.

athei:
I added the code that uses a rust module to calculate the fees. It works for me and yields the correct result. I also rebased it onto the current master.

CI is currently failing because it does not build the rust package nor does it has the required toolchain installed. I will fix this soon.

calc-fee/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[features]
+debug = ["console_error_panic_hook"]
+
+[dependencies]
+wasm-bindgen = { version = "0.2", default_features = false, features = ["serde-serialize"] }
+serde_derive = { version = "1", default_features = false }
+serde = { version = "1", default_features = false }
+console_error_panic_hook = { version = "0.1", optional = true }
+sp-arithmetic = { git = "https://github.com/paritytech/substrate", default_features = false }
```
athei, line 22: We should replace that by a released version as soon as that is possible. The current release on crates.io (2.0.0-rc2) has a broken Fixed128 impl that does not work with negative coefficients on unsigned Balances.

PR #51, opened 2020-05-20
`locks` for `/balance` should always be an array

danforbes:


danforbes:
This PR is being closed based on this guidance:
> ...stay as close as possible to `@polkadot/api`, and let `locks[i].reasons` just return a string, since the array version is actually legacy...

src/ApiHandler.ts
```diff
@@ -181,7 +181,7 @@ export default class ApiHandler {
 				reserved,
 				miscFrozen,
 				feeFrozen,
-				locks,
+				locks: Array.isArray(locks) ? locks : [ locks ],
```
danforbes, line 184: Noting this is actually incorrect...the desire was to make sure that `locks.reasons` is always an array.

PR #57, opened 2020-05-28
Add fee info endpoint(s)

### LINKED ISSUE #49: Add fee prediction endpoint
Transaction Payment provides a `queryInfo` endpoint to predict current transaction fees given the current fee multiplier. Provide an endpoint to call this and get current fee info.

Alternatively: use the latest fee multiplier and calculate this directly in sidecar.

emostov:
Closes #49

Introduce an endpoint to query for the fee estimate of an extrinsic.

The route is POST `tx/fee-estimate` accepting a body with the form `{ “extrinsic”:”0x…” }`.

The route was previously a GET, but after a consultation with @danforbes, it has been decided that in order to comply with RESTful principles it should be switched to a POST. The premise for this decision is that an extrinsic submitted for a fee estimate is creating a fee estimate and not accessing a (on chain) resource. From a practical standpoint, extrinsics can be of arbitrary length and although HTTP spec sets no upper bound to URL length, it is common to keep them [under 2048 characters.](https://stackoverflow.com/questions/29458445/what-is-a-safe-maximum-length-a-segment-in-a-url-path-should-be/33733386#:~:text=According%20to%20the%20HTTP%20spec,remain%20under%20approximately%202000%20characters.)

I have tested this current draft against Polkadot v0.8.1 in —dev mode and have only tested `balances.transfer` encoded extrinsics that have been created offline and never submitted.

I think the major point to consider for review is if this is that exact signature we want to expose and maintain.

Two smaller things I am looking for review:
- Are the README.md updates to describe the new path appropriate? I tried to closely follow the conventions established by the other POST route descriptor.
- I merged in the latest master in an attempt to deal with the unexpected `yarn.lock` diff. Upon manually inspection it seems like the `package.json` and `yarn.lock` look the same as the current master. However, the version of master the PR diff uses seems to be different (older) - causing a an unexpected diff for `package.json` and `yarn.lock`. Still getting used to submitting PRs so likely I'm missing something obvious.

Below are samples of a what response currently look like
On error:

```
{
    "error": "Unable to fetch fee info",
    "data": {
        "extrinsic": "0x2902841cbd2d43530a44705ad088af313e18f80b53ef16b36177cd4b77b846f2a5f07c01b0bcd3422b88513c2c680a5d18bb79f5a14cf8e71030487be8c7e4d8630ff77346e5e8847fabf20960c13d7f381adb4a7c92f0a03163aca099ce389f479f4584b50200000500d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27da10",
        "block": "0xfcf2f48899793dd848407217912961bfad2322dea0f2054c6499e969f4e76535"
    },
    "cause": "Error: 2: Unable to query dispatch info.: Invalid transaction version"
}
```

On success:

```
{
    "weight": "195000000",
    "class": "Normal",
    "partialFee": "165600000"
}
```


package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "tsc",
     "start": "tsc && node ./build/main.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "debug": "tsc && node --inspect ./build/main.js"
```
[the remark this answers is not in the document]
amaurymartiny: +1, i'd say `--inspect` should be rare enough, we shouldn't need another `debug` command.

btw it's strictly equivalent to `yarn start --inspect`

yarn.lock
```diff
@@ -153,6 +153,18 @@
   resolved "https://registry.yarnpkg.com/@polkadot/wasm-crypto/-/wasm-crypto-1.2.1.tgz#2189702447acd28d763886359576c87562241767"
   integrity sha512-nckIoZBV4nBZdeKwFwH5t7skS7L7GO5EFUl5B1F6uCjUfdNpDz3DtqbYQHcLdCZNmG4TDLg6w/1J+rkl2SiUZw==
 
+"@sindresorhus/is@^0.14.0":
```
amaurymartiny, line 156: which package did you add that modifies yarn.lock? I see no change in package.json.

emostov: The two things I did that might have affected this was that I had `nodemon` for a little bit as a dependency, but later removed from the package.json. I also ran `yarn upgrade`. But I am not exactly sure which would cause yarn.lock to be modified here.

README.md
```diff
@@ -47,7 +47,11 @@ yarn start
 
 - `/tx/artifacts/` fetch artifacts used for creating transactions at latest finalized block.
 
-- `/tx/artifacts/NUMBER` fetch artifacts used for creating transactions at bloch height `NUMBER`.
+- `/tx/artifacts/NUMBER` fetch artifacts used for creating transactions at block height `NUMBER`.
+
+- `/tx/fee/:extrinsic/:number/` fetch predicted fee info for an encoded `EXTRINSIC` in context of the block at height `NUMBER`.
```
joepetrowski, line 52:
```suggestion
- `/tx/fee/EXTRINSIC/NUMBER` fetch predicted fee info for an encoded `EXTRINSIC` in context of the
  block at height `NUMBER`.
```

README.md
```diff
@@ -47,7 +47,11 @@ yarn start
...
+
+-`/tx/fee/:extrinsic/` fetch predicted fee info for an encoded `EXTRINSIC`.
```
joepetrowski, line 54:
```suggestion
-`/tx/fee/EXTRINSIC` fetch predicted fee info for an encoded `EXTRINSIC`.
```

src/main.ts
```diff
@@ -185,6 +185,21 @@ async function main() {
 		return await handler.fetchTxArtifacts(hash);
 	});
 
+	get('/tx/fee/:extrinsic/:number', async (params) => {
```
danforbes, line 188: I think that this should be implemented as a `POST` as opposed to a `GET`. I had a brief conversation with @joepetrowski about this, @emostov, and will do the same with you :rocket:

README.md
```diff
@@ -47,8 +47,21 @@ yarn start
 
 - `/tx/artifacts/` fetch artifacts used for creating transactions at latest finalized block.
 
-- `/tx/artifacts/NUMBER` fetch artifacts used for creating transactions at bloch height `NUMBER`.
+- `/tx/artifacts/NUMBER` fetch artifacts used for creating transactions at block height `NUMBER`.
 
+- `/tx/fee-estimate` submit an extrinsic in order to get back a fee estimation. Expects a string with a hex-encoded extrinsic in a JSON POST.
+  body:
```
joepetrowski, line 53:
```suggestion
- `/tx/fee-estimate` submit an extrinsic in order to get back a fee estimation. Expects a string
  with a hex-encoded extrinsic in a JSON POST body:
```

README.md
```diff
@@ -47,8 +47,21 @@ yarn start
...
+  ```
+  curl localhost:8080/tx/fee-estimate -X POST --data '{"extrinsic": "0x..."}' -H 'Content-Type: application/json'
+  ```
+  Expected result is json with fee information:
```
joepetrowski, line 57:
```suggestion
  Expected result is a JSON with fee information:
```

PR #59, opened 2020-05-30
Rename fee variable

joepetrowski:
Change `extrinsic` to `tx`. Although you could submit an extrinsic here, it wouldn't make much sense as only signed transactions actually pay fees. Also one would expect the same post body, `'{"tx": "0x..."}'`, as submitting the transaction.


danforbes:
As you continue making more changes like this you may want to investigate libraries like [`express-validator`](https://github.com/express-validator/express-validator).

PR #62, opened 2020-06-01
Parse nested calls

### LINKED ISSUE #48: Parse call options
Some calls actually take other calls as their arguments, for example `as_multi` or `batch` transactions. Each inner call has a `callIndex` and `call` to identify the call. `callIndex` refers to the index of the pallet in the metadata, not counting those with `null` calls, and `call` is the index of the call within that module.

Instead of displaying `{ callIndex: '06', call: '05' }`, display `'staking.nominate'` (more generally, `'pallet.methodName'`).

emostov:
Edited. 

Closes #48

 `method: {module}.{method}` is added, and and in an additional block field called `newArgs`, arguments to calls are labelled with parameter names. `newArgs` is added so as not to break the API. Eventually, the plan is to replace the value of `args` with what is now the value of `newArgs`

Two points I was considering for discussion:

  1) Do we want to keep `callIndex`? It could allow for more advanced usage with metadata but also could be cumbersome to maintain support for.
  2) Are there any edge cases where the logic makes incorrect assumptions?

---
See [gist for sample block response](https://gist.github.com/emostov/cad68ffe955f9f8687be4b86cff87720) and [this triple batch gist](https://gist.github.com/emostov/787120b14cacc6a894445a38b8e91c24)

To test:
- Test with reconciliation script


joepetrowski:
Wow, a `batch(batch(batch()))`! Let's also test this on some of the actual batch calls in Kusama.

src/utils.ts
```diff
@@ -10,6 +11,26 @@ export function parseNumber(n: string): number {
 	return num;
 }
 
+// Note: this could be a good place to put any additional call parsing logic or
+// even argument parsing logic.
+// Takes in a call or an array that contains a call and recursively parses all calls.
+export function parseCalls(data: any): any {
+	if (data instanceof GenericCall){
```
danforbes, line 18: I like your idea of continuing to include the old information along with the new information. In fact, if we are able to finagle this change so that it only _adds_ information (as opposed to _replacing_ old information with new information), it will not be considered a breaking API change :rocket:

src/ApiHandler.ts
```diff
@@ -61,11 +63,13 @@ export default class ApiHandler {
 			const { method, nonce, signature, signer, isSigned, tip, args } = extrinsic;
 			const hash = u8aToHex(blake2AsU8a(extrinsic.toU8a(), 256));
 
+			const argsWithParsedCalls = parseCalls(args);
```
danforbes, line 66: May be best to just move the `parseCalls` function directly to line 72...this will keep the diff "smaller".

src/utils.ts
```diff
@@ -10,6 +11,26 @@ export function parseNumber(n: string): number {
 	return num;
 }
 
+// Note: this could be a good place to put any additional call parsing logic or
+// even argument parsing logic.
+// Takes in a call or an array that contains a call and recursively parses all calls.
+export function parseCalls(data: any): any {
```
danforbes, line 17: It may be better to put this function directly inside ApiHandler, since that is probably the only place it will ever be used for. In general, "utility" files are sort of considered an anti-pattern and should be used _very_ carefully. Ref http://ralin.io/blog/oop-anti-patterns-utility-or-helper-classes.html

src/utils.ts
emostov: `callIndex` now back in.

src/ApiHandler.ts
```diff
@@ -501,40 +501,51 @@ export default class ApiHandler {
...
+		// A possible pitfall to look out for here is if a call or array of calls
+		// does not have the parameter name `call` or `calls`.
 		if (labelledArgs.call !== undefined) {
-			const genericCallArg = args.find((a: any): boolean => a instanceof GenericCall);
-			if (genericCallArg instanceof GenericCall) {
+			const genericCallArg = args.find((argument: any): boolean => 
+				argument instanceof GenericCall
+			);
+
+			if (genericCallArg !== undefined) {
```
danforbes, line 527: Do we want to cover cases where `genericCallArg` is `null`? If so this can just be `if (genericCallArg)` but not sure what the requirement is here.

PR #69, opened 2020-06-06
Add specName and chainName to Artifacts

joepetrowski:
`specName` comes from an RPC call but AFAICT, nothing returns the `chainName` for the registry. Used a switch/case block here but I could see many possible approaches to this:

1. ~~Leave as is. The reason I did this is because TxWrapper's registry will only accept `'Polkadot' | 'Kusama' | 'Westend'`, so I explicitly filter for them here.~~
2. ~~Remove the switch/case and just do the `default` block now, which capitalizes the first letter of `specName`. This assumes that all chains and the Polkadot JS API type registry will follow this `specName`/`chainName` pattern.~~
3. ~~Just remove `chainName` from the return entirely and let people deal with this `specName -> chainName` conversion on the outside.~~
4. Some RPC endpoint actually does return `chainName` and this is all pointless.

To release only after https://github.com/paritytech/txwrapper/pull/161 is in a txwrapper release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "polkadot-rpc-proxy",
-  "version": "0.4.0",
+  "version": "0.7.0",
   "description": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


amaurymartiny:
> Some RPC endpoint actually does return chainName and this is all pointless

ahah, actually yeah: https://polkadot.js.org/api/substrate/rpc.html#chain-text

PR #73, opened 2020-06-10
Add doc comments to main

joepetrowski:
Probably should use something like typedoc, but thought I'd get started with this. We should also consider adding `/docs/` endpoints that would fetch the docs from the metadata on anything returned.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -149,6 +149,8 @@ async function main() {
 	//   - `reasons`: If true, then the lock remains in effect even for payment of transaction fees.
 	//
 	// Substrate Reference:
+	// - FRAME System: https://crates.parity.io/frame_system/index.html
+	// - Balances Pallet: https://crates.parity.io/pallet_balances/index.html
 	// - `AccountInfo`: https://crates.parity.io/frame_system/struct.AccountInfo.html
 	// - `AccountData`: https://crates.parity.io/pallet_balances/struct.AccountData.html
 	// - `BalanceLock`: https://crates.parity.io/pallet_balances/struct.BalanceLock.html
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -71,19 +71,87 @@ async function main() {
...
+	// GET balance information for an address.
+	//
+	// Paths:
+	// - `address`: The address to query.
+	// - (Optional) `number`: Block number at which to query. If not provided, queries finalized head.
+	//
+	// Returns:
+	// - `at`: Block number and hash at which the call was made.
+	// - `nonce`: Account nonce.
+	// - `free`: Free balance of the account. Not equivalent to _spendable_ balance. This is the only
```
emostov, line 138: Not necessary, but I think the docs could be more comprehensive by linking to something like this for balances: https://stackoverflow.com/questions/56418100/what-is-the-difference-between-free-balance-reserved-balance-locked-balance-v

src/main.ts
```diff
@@ -99,6 +167,22 @@ async function main() {
...
+	// Paths:
+	// - `address`: The _Stash_ address for staking.
+	// - (Optional) `number`: Block number at which to query. If not provided, queries finalized head.
+	//
+	// Returns:
+	// - `at`: Block number and hash at which the call was made.
+	// - `rewardDestination`: The account to which rewards will be paid. Can be 'Staked' (Stash
+	//   account, adding to the amount at stake), 'Stash' (Stash address, not adding to the amount at
+	//   stake), or 'Controller' (Controller address).
+	// - `bonded`: Controller address for the given Stash.
```
emostov, line 181: Should this mention the controller address can also possibly be the stash address? I think that can get a little confusing.

src/main.ts
```diff
@@ -114,6 +198,32 @@ async function main() {
...
 
+	// GET staking information for an address.
+	//
+	// Paths:
+	// - `address`: The _Controller_ address for staking.
+	// - (Optional) `number`: Block number at which to query. If not provided, queries finalized head.
+	//
+	// Returns:
+	// - `at`: Block number and hash at which the call was made.
+	// - `stash`: The stash account whose balance is actually locked and at stake.
```
emostov, line 209: If we want to address my above comment we could add the vice versa here; Stash could be the same as controller

src/main.ts
joepetrowski: I don't think so, that SO isn't maintained and these should be strictly reference.

src/main.ts
joepetrowski: Maybe I'll just link to the ref doc of the Staking pallet. But I think this should just say what this function does, not try to explain all of substrate.

PR #75, opened 2020-06-11
Add author ID to `/block` response

### LINKED ISSUE #74: Add the field `author: <AccountId>` to `fetchBlock`'s response
For some use cases it may helpful to have an `author` field for the `block` endpoint.

danforbes:
Response from Kusama looks like this (note that when I ran against local dev node the author ID was not defined)
image

Closes #74

### DIFF
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -247,6 +247,9 @@ export default class ApiHandler {
 			}
 		}
 
+		const header = await api.derive.chain.getHeader(hash);
+		const authorId = header?.author;
+
 		return {
 			number,
 			hash,
@@ -257,6 +260,7 @@ export default class ApiHandler {
 			onInitialize,
 			extrinsics,
 			onFinalize,
+			authorId,
 		};
 	}
 
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -66,6 +66,9 @@ export default class ApiHandler {
 
 		const { parentHash, number, stateRoot, extrinsicsRoot } = block.header;
 
+		const header = await api.derive.chain.getHeader(hash);
+		const authorId = header?.author;
+
 		const logs = block.header.digest.logs.map((log) => {
 			const { type, index, value } = log;
 
@@ -247,20 +250,17 @@ export default class ApiHandler {
 			}
 		}
 
-		const header = await api.derive.chain.getHeader(hash);
-		const authorId = header?.author;
-
 		return {
 			number,
 			hash,
 			parentHash,
 			stateRoot,
 			extrinsicsRoot,
+			authorId,
 			logs,
 			onInitialize,
 			extrinsics,
 			onFinalize,
-			authorId,
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


src/ApiHandler.ts
```diff
@@ -232,6 +232,9 @@ export default class ApiHandler {
 			}
 		}
 
+		const header = await api.derive.chain.getHeader(hash);
```
joepetrowski, line 235: Can you not pull the author out of the header [here](https://github.com/paritytech/substrate-api-sidecar/blob/60face540ed5e85254000e78a2557ab516a70dd3/src/ApiHandler.ts#L53)?

Even if not, I think it'd make more sense to put this call [here](https://github.com/paritytech/substrate-api-sidecar/blob/60face540ed5e85254000e78a2557ab516a70dd3/src/ApiHandler.ts#L48)

src/ApiHandler.ts
```diff
@@ -242,6 +245,7 @@ export default class ApiHandler {
 			onInitialize,
 			extrinsics,
 			onFinalize,
+			authorId,
```
joepetrowski, line 248: A nit, but this makes more sense to me by the header info. Right before or right after `logs`.

src/ApiHandler.ts
danforbes: Good looking out here (I totally double-checked), but no I can't get the author ID from _that_ header. Here are the attributes of that header:
```
/** @name Header */
export interface Header extends Struct {
    readonly parentHash: Hash;
    readonly number: Compact<BlockNumber>;
    readonly stateRoot: Hash;
    readonly extrinsicsRoot: Hash;
    readonly digest: Digest;
}
```

I followed [the documented steps from the official Polkadot JS  website](https://polkadot.js.org/api/cookbook/blocks.html#how-do-i-extract-the-block-author). Agree that the placement you suggest is better; all things being equal I usually just add new stuff "to the end" but in this context you're definitely right :+1:

joepetrowski: Interesting, I do wonder if it's possible to get author from `digest`

src/main.ts
```diff
@@ -82,6 +82,7 @@ async function main() {
 	// - `parentHash`: The hash of the parent block.
 	// - `stateRoot`: The state root after executing this block.
 	// - `extrinsicsRoot`: The Merkle root of the extrinsics.
+	// - `authorId`: The account ID of the block author (may be undefined for development chains).
```
joepetrowski, line 85: If we want this to be generic for Substrate chains, then some production chains might not have have account IDs, although for Polkadot and Kusama this statement is true.
```suggestion
	// - `authorId`: The account ID of the block author (may be undefined for some chains).
```

src/ApiHandler.ts
emostov: It looks like `api.derive.chain.getHeader` allows you to extract the author from `digest.logs` with some logic that is dependent on the runtime engine. I think we could probably replicate that logic here if we really wanted to avoid the extra RPC call. I believe we could stick [this exact logic in](https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/type/HeaderExtended.ts#L28). Not sure how great of an idea code duplication is though since if that ever gets changed we would have to change it here, but I think its worth considering for performance

danforbes: This sounds like premature optimization to me. I believe that this naive solution will work in the short-term. Once we get some benchmarking in place, we can evaluate possible bottlenecks and optimize where needed. My $0.02.

PR #77, opened 2020-06-12
Get Block Should Accept Hash

### LINKED ISSUE #76: Get Block Should Accept Hash
A lot of systems, including Sidecar's console log, identify blocks by hash rather than number. the `block/:number` endpoint should be able to accept either an int or a hash and return the block accordingly. 

Sample error from Sidecar's console:
```
Internal Error: Error: Missing extrinsic 2 in block 0x9e49962fbbfca0d6921e422e4fc4e3e5efbf4cfe3318847697c035891e9c0a12
```

danforbes:
Closes #76

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,7 +17,7 @@
 import { ApiPromise } from '@polkadot/api';
 import { WsProvider } from '@polkadot/rpc-provider';
 import type { BlockHash } from '@polkadot/types/interfaces';
-import { hexToNumber, isHex } from '@polkadot/util';
+import { isHex } from '@polkadot/util';
 import * as bodyParser from 'body-parser';
 import * as express from 'express';
 
@@ -95,10 +95,10 @@ async function main() {
 	// GET a block.
 	//
 	// Paths:
-	// - (Optional) `number`: Block hash or number at which to query. If not provided, queries finalized head.
+	// - (Optional) `number`: Block hash or height at which to query. If not provided, queries finalized head.
 	//
 	// Returns:
-	// - `number`: Block number.
+	// - `number`: Block height.
 	// - `hash`: The block's hash.
 	// - `parentHash`: The hash of the parent block.
 	// - `stateRoot`: The state root after executing this block.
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -120,8 +122,13 @@ async function main() {
 	});
 
 	get('/block/:number', async (params) => {
-		const number = parseNumber(params.number);
-		const hash = await api.rpc.chain.getBlockHash(number);
+		let hash: BlockHash;
+		if (isHex(params.number)) {
```
joepetrowski, line 126: So, I know with txwrapper, you can construct an unsigned transaction with the block number in hex, like '0x1234'. Makes me wonder two things:
1. Some users might just have hex everything, in which case this should also check that the hex is 32 bytes.
2. Maybe Polkadot JS already has a nice function for this logic?

danforbes: Yeah, totally good point :+1:

amaurymartiny: Yeah correct. I'm not aware of any helper function in polkadot-js, but 32-byte check should be fine.

There's an edge case where the hex-encoded number is 32-bytes, but this should happen in the very distant future, so no worries. Another approach, without disambiguation, would be to have another endpoint `/blockByHash/:hash`

danforbes:
Converting this to draft so I can rebase on top of the ESLint changes :rocket:

danforbes:
Okay rebased and linted :+1:

src/main.ts
```diff
@@ -454,4 +445,33 @@ async function main() {
...
+	api: ApiPromise,
+	blockId: string
+): Promise<BlockHash> {
+	try {
+		const isHexStr = isHex(blockId);
+		if (isHexStr && blockId.length === 66) {
+			return api.createType('BlockHash', blockId);
+		} else if (isHexStr) {
+			throw {
+				error: `Cannot get block hash for ${blockId}. Hex string block IDs must be 66-characters (32-bytes) in length.`,
```
joepetrowski, line 458: I think it'd be good to convert this to a number if some people's systems use hex. There's no way block number collides with a hash that's 32 bytes.

emostov: Not sure I totally understand, but would [the conversion](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/utils.ts#L8) be surfacing an additional point of failure/ misdirection that could obscure the original error? i.e if someone accidentally sends the string "Hello", but we convert it to a number before interpolating into the error message, it could throw off debugging.

danforbes: Yeah, I think I'm with Zeke on this one but I'm still not completely sure this is what Joe was suggesting so I'll just hold off.

PR #78, opened 2020-06-12
Rudimentary Error Handling

danforbes:


src/types/errors/SidecarUserError.ts
```diff
@@ -0,0 +1,13 @@
+import SidecarError from "./SidecarError";
+
+export default class SidecarUserError implements SidecarError {
+  name: 'SidecarUserError';
+  message: string;
+  stack: undefined;
```
amaurymartiny, line 6: I think there's a confusion between type and value. Here it means that `stack` is of type `undefined`, which means it can only take `undeinfed` as value.

Maybe you meant `stack: string | undefined = undefined`?

src/types/errors/SidecarUserError.ts
```diff
@@ -0,0 +1,13 @@
+import SidecarError from "./SidecarError";
+
+export default class SidecarUserError implements SidecarError {
```
amaurymartiny, line 3: why not extend?

danforbes:
This is not the right time for this PR...mea culpa in opening it. This should be considered as part of Sidecar v1.

PR #83, opened 2020-06-13
Fix bug when fetching genesis block

### LINKED ISSUE #82: Get block 0
Right now we are querying some fee params on `parentHash`, but of course it can't get them for block 0 because the parent is hardcoded `0x000...000`. But there are no txs in block 0, so we should only do fee calculation if `blockNumber > 0`.

danforbes:
Closes #82


danforbes:
Yep, @joepetrowski is definitely right. I've addressed this issue.

PR #90, opened 2020-06-16
Upgrade deps; Correct issues from upgrade; Setup jest unit tests for `sanitizeNumbers`;

### LINKED ISSUE #89: Use jest to unit test sanitizeNumbers
Configure Jest in the repo and setup some basic sanity checks for `sanitizeNumbers`.

emostov:
Closes #89 

### Motivation:
While writing testing for `santizeNumbers` I bumped @polkadot/api from 1.23.0.beta2 to 1.23.1 and the tests I wrote for `Compact<Balance>` started failing. I also realized the server would not start due to several `u32` types being created with a -1. I later found that it was failing on an enum type in `Vec<BalanceLock>` so I added enum handling to `sanitizeNumbers` as well. I figured the best move would be to just fix all the issues in one go since async coordination of dependent merges could be considered unnecessarily slow and tricky given the context.

### Overview:
- Upgrade all deps
- Add tests for sanitizeNumbers
- Correct sanitizeNumbers to correctly handle new Compact format
- Change all occurrences of `createType('u32', -1)` to `createType('u32', 0)` in order to not trigger the new validation error
- Update CI to incorporate testing

### Future PRs
- Roll out new `sanitizeNumbers` (#154)
- Add tests for every possible type that might go through sanitizeNumbers to ensure it acts appropriately (should also go into #154)


src/ApiHandler.ts
```diff
@@ -54,8 +54,8 @@ export default class ApiHandler {
 	constructor(api: ApiPromise) {
 		this.api = api;
 		// this.wsUrl = wsUrl;
-		this.specVersion = api.createType('u32', -1);
-		this.txVersion = api.createType('u32', -1);
+		this.specVersion = api.createType('u32', 0);
```
joepetrowski, line 57: The problem with this is that it gets initialized to `-1` because the spec number actually could be `0`, so whenever it makes its first call, we want to force it to call `ensureMeta` and get the right metadata.

src/ApiHandler.ts
```diff
@@ -701,8 +701,8 @@ export default class ApiHandler {
 				`Failed to get Metadata for block ${hash}, using latest.`
 			);
 			console.error(err);
-			this.specVersion = api.createType('u32', -1);
-			this.txVersion = api.createType('u32', -1);
+			this.specVersion = api.createType('u32', 0);
+			this.txVersion = api.createType('u32', 0);
```
joepetrowski, line 705: Same issue, these are valid spec and tx version numbers. If we're in an error condition we don't want to return results that might be wrong.

src/utils/test_util.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * A dummy return value to fetchStakingLedger which has not been run through `sanitizeNumbers`.
+ */
+export const PRE_SANITIZED_STAKING_RESPONSE = {
+	at: PRE_SANITIZED_AT,
+	staking: kusamaRegistry.createType('StakingLedger', {
+		stash: '5DRihWfVSmhbk25D4VRSjacZTtrnv8w8qnGttLmfro5MCPgm',
+		total: '0x0000000000000000ff49f24a6a9c00',
+		active: '0x0000000000000000ff49f24a6a9100',
+		unlocking: [],
```
joepetrowski, line 47: Is this a real example? Normally `total = active + unlocking.sum()`. It should work because that shouldn't affect `sanitizeNumbers`, but I don't think it's an actual state.

joepetrowski: Also why are these 15 bytes? Should be 16, no?

src/utils/test_util.ts
```diff
@@ -0,0 +1,82 @@
...
+		startingBlock: '299694200',
+	}
+);
+
+export const PRE_SANITIZED_RUNTIME_DISPATCH_INFO = kusamaRegistry.createType(
+	'RuntimeDispatchInfo',
+	{
+		weight: '0x7FFFFFFFFFFFFFFF',
+		class: 'operational',
+		partialFee: '0x7FFFFFFFFFFFFFFF',
```
joepetrowski, line 80: `weight` is correctly 8 bytes, but shouldn't `partialFee` be 16?

src/utils/utils.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+		expect(sanitizeNumbers(uIntTen)).toBe('10');
+
+		const uIntPaddedHex = new UInt(
+			kusamaRegistry,
+			'0x000000000000000004fe9f24a6a9c00'
+		);
+		expect(sanitizeNumbers(uIntPaddedHex)).toBe('22493750000000000');
+
+		const u32Max = kusamaRegistry.createType('u32', '0x7FFFFFFF');
+		expect(sanitizeNumbers(u32Max)).toBe('2147483647');
```
joepetrowski, line 43: Why not use the first bit? It's unsigned.
https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX

src/utils/test_util.ts
joepetrowski: Let's work on this in follow ups, but shouldn't block this.

PR #92, opened 2020-06-16
Use `unwrap()`-ed `StakingLedger` values converted to string

### LINKED ISSUE #91: `/staking/ADDRESS` fields `total` and `active` should always have decimal values.
Sometimes `/staking/ADDRESS` inappropriately returns non-decimal values for `total` and `active`. The values for these fields should be decimals.

emostov:
Closes #91 

Use `unwrap()`-ed  `StakingLedger` values converted to string to ensure proper decimal serialization.

### DIFF
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -400,7 +400,13 @@ export default class ApiHandler {
 
 		return {
 			at,
-			staking,
+			staking: {
+				stash: ledger.stash.toString(),
+				total: ledger.total.toString(),
+				active: ledger.active.toString(),
+				unlocking: ledger.unlocking.toString(),
+				claimedRewards: ledger.claimedRewards.toString(),
+			},
 			numSlashingSpans,
 		};
 	}


src/ApiHandler.ts
```diff
@@ -400,7 +400,13 @@ export default class ApiHandler {
 
 		return {
 			at,
-			staking,
+			staking: {
```
amaurymartiny, line 403: fyi, `Codec`s also have `.toJSON()` and `.toHuman()`, which might suit better than .toString (e.g. for fields that are numbers)

src/ApiHandler.ts
```diff
@@ -400,7 +400,13 @@ export default class ApiHandler {
...
+				stash: ledger.stash.toString(),
+				total: ledger.total.toString(),
+				active: ledger.active.toString(),
+				unlocking: ledger.unlocking.toString(),
+				claimedRewards: ledger.claimedRewards.toString(),
```
joepetrowski, line 408: We actually cannot assume that this is present. In some older Kusama runtimes it was `lastReward` and in older ones it didn't exist. Also `lastReward` was an integer but `claimedRewards` is an array of integers.

src/ApiHandler.ts
joepetrowski: +1 for using something on `staking` itself rather than its fields.

danforbes: Does this change the "shape" of the HTTP response?

emostov:
The issue these PR sought to address no longer seems to be a problem.

emostov:
The issue these PR sought to address no longer seems to be a problem.

PR #94, opened 2020-06-16
Reconfigure `/staking/ADDRESS` to accept stash and give `controller`, `rewardDestination`

emostov:
#### Overview

- Delete `payout/:address` and move that info to '/staking' in order to create a single endpoint for an accounts staking info.
- Start migration to a _more_ RESTful resource model by changing endpoint paths.
- Rudimentary error handling improvement in `get`

#### Endpoint change from `staking/:address/:number` to `accounts/:address/staking/:number` discussion

tl;dr - we should think about changing the path to `accounts/:address/staking/:number` because it more accurately models our resource taxonomy.

As we approach V1 and dialing in the api, the importance of communicating meaning through our endpoint increases and thus it behooves us to carefully consider how they model resources.

After offline discussion with @danforbes, consensus has begun to coalesce around modeling the resource `accounts`, representing the (collection of) `AccountId`s of the chain. Under `accounts` we would then have single resource, an `account`, identified by its `AccountId`, (commonly referred to in sidecar and its reference material as ADDRESS or `address`).

Following this we can identify the following resources and relationships to an `account`:

- `balance` belongs to an `account`; an `account` has one `balance`.
- `staking{information}` belongs to an `account`; an `account` has one `staking{information}`.
- `vesting{information}` belongs to an `account`; an `account` has one `vesting{information}`.

Using this model we can then define paths that semantically correspond:

- `accounts/:address/balance`
- `accounts/:address/staking`
- `accounts/:address/vesting`

If critical consensus can be reached on the aforementioned paths, it still leaves the question of where an optional `number` (block hash or height) parameter would semantically fit with our model. Some say REST principles lean towards using a query param because `number` is not necessarily a resource ID, but instead a search term. However, to another subset of people, simply putting `number` at the end of the path still communicates model while staying terse. These two options for `number` would look like:

1) `accounts/:address/staking?number=2643`
2) `accounts/:address/staking/:number`

To move the discussion forward, I have gone ahead and changed the staking paths to `accounts/:address/staking/` and `accounts/:address/staking/:number`

_What about the other two routes?_ 
In order to keep the diff for this PR minimal I decided to only change this route for now. However, if this change is implemented the other routes should be changed prior to the v1 release.

#### Follow up PRs:

- Creating a swagger to document API
- Modify other endpoints to stay in line with model

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,9 @@ a 32-byte hex string (`0x` followed by 64 hexadecimal digits) that denotes the b
 
 - `/balance/ADDRESS/NUMBER` fetch balances for `ADDRESS` at the block identified by 'NUMBER`.
 
-- `/payout/ADDRESS` fetch payout info for `ADDRESS` at latest finalized block.
+- `/staking/ADDRESS` fetch the staking info for `ADDRESS` at latest finalized block.
 
-- `/payout/ADDRESS/NUMBER` fetch payout info for `ADDRESS` at the block identified by 'NUMBER`.
-
-- `/staking/ADDRESS` fetch the staking ledger info for `ADDRESS` at latest finalized block.
-
-- `/staking/ADDRESS/NUMBER` fetch the staking ledger info for `ADDRESS` at the block identified by 'NUMBER`.
+- `/staking/ADDRESS/NUMBER` fetch the staking info for `ADDRESS` at the block identified by 'NUMBER`.
 
 - `/vesting/ADDRESS` fetch the vesting info for `ADDRESS` at latest finalized block.
 
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -364,7 +364,7 @@ export default class ApiHandler {
 	/**
 	 *
 	 * @param hash BlockHash to make call at.
-	 * @param stash: _Stash_ address.
+	 * @param stash _Stash_ address.
 	 */
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 	async fetchStakingInfo(hash: BlockHash, stash: string) {
@@ -380,14 +380,11 @@ export default class ApiHandler {
 			height: header.number.toNumber().toString(10),
 		};
 
-		if (
-			controllerOption.isNone === true ||
-			controllerOption.isSome !== true
-		) {
+		if (controllerOption.isNone) {
 			// if controllerOption.isNone, `stash` address param is not a stash.
 			throw {
 				error: `The address ${stash} is not a stash address.`,
-				statusCode: 422,
+				statusCode: 400,
 			};
 		}
 
@@ -403,20 +400,19 @@ export default class ApiHandler {
 			await api.query.staking.slashingSpans.at(hash, stash),
 		]);
 
-		// should always work because we know that we have a bonded pair
 		const stakingLedger = stakingLedgerOption.unwrapOr(null);
 
 		if (stakingLedger === null) {
+			// should never throw because by time we get here we know we have a bonded pair
 			throw {
 				error: `Staking ledger could not be found for address "${controller.toString()}"`,
 				statusCode: 404,
 			};
 		}
 
-		const numSlashingSpans =
-			slashingSpansOption.isSome === true
-				? slashingSpansOption.unwrap().prior.length + 1
-				: 0;
+		const numSlashingSpans = slashingSpansOption.isSome
+			? slashingSpansOption.unwrap().prior.length + 1
+			: 0;
 
 		return {
 			at,
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -380,23 +380,25 @@ export default class ApiHandler {
 			height: header.number.toNumber().toString(10),
 		};
 
-		const noneResponse = {
-			at,
-			controller: null,
-			rewardDestination: null,
-			numSlashingSpans: null,
-			staking: {},
-		};
-
 		if (bonded.isNone) {
 			// If bonded.isNone, address is not a stash.
-			return noneResponse;
+			return {
+				at,
+				controller: null,
+				rewardDestination: null,
+				numSlashingSpans: null,
+				staking: {},
+			};
 		}
 
-		// We can safely infer (optionBonded.isSome === true) and safely unwrap
+		// We can infer (optionBonded.isSome === true) and safely unwrap
 		const controller = bonded.unwrap();
 
-		const [stakingLedger, rewardDestination, slashingSpans] = await Promise.all([
+		const [
+			stakingLedger,
+			rewardDestination,
+			slashingSpans,
+		] = await Promise.all([
 			await api.query.staking.ledger.at(hash, controller),
 			await api.query.staking.payee.at(hash, address),
 			await api.query.staking.slashingSpans.at(hash, address),
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -380,7 +380,7 @@ export default class ApiHandler {
 			height: header.number.toNumber().toString(10),
 		};
 
-		if (optionController.isNone) {
+		if (optionController.isNone === true) {
 			// if bonded.isNone, `stash` address param is not a stash.
 			return {
 				at,
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -363,19 +363,18 @@ export default class ApiHandler {
 
 	/**
 	 *
-	 * @param hash: BlockHash
-	 * @param address: string - _Stash_ address
+	 * @param hash BlockHash to make call at.
+	 * @param address: _Stash_ address.
 	 */
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 	async fetchStakingLedger(hash: BlockHash, address: string) {
 		const api = await this.ensureMeta(hash);
 
-		// optionBonded: Option<AccountId>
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -40,13 +40,9 @@ a 32-byte hex string (`0x` followed by 64 hexadecimal digits) that denotes the b
 
 - `/balance/ADDRESS/NUMBER` fetch balances for `ADDRESS` at the block identified by 'NUMBER`.
 
-- `/payout/ADDRESS` fetch payout info for `ADDRESS` at latest finalized block.
+- `/staking/ADDRESS` fetch the staking info for `ADDRESS` at latest finalized block.
```
danforbes, line 43: I think we may want to model this as `/ADDRESS/staking`.

src/ApiHandler.ts
```diff
@@ -361,47 +361,63 @@ export default class ApiHandler {
...
+	 * @param hash BlockHash to make call at.
+	 * @param stash: _Stash_ address.
+	 */
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-	async fetchStakingLedger(hash: BlockHash, address: string) {
+	async fetchStakingInfo(hash: BlockHash, stash: string) {
 		const api = await this.ensureMeta(hash);
 
-		const [header, staking] = await Promise.all([
+		const [header, optionController] = await Promise.all([
```
danforbes, line 373: Think `controllerOption` would be more evocative here.

src/ApiHandler.ts
```diff
@@ -361,47 +361,63 @@ export default class ApiHandler {
...
 
 		const at = {
 			hash,
 			height: header.number.toNumber().toString(10),
 		};
 
-		if (staking.isNone) {
-			// Address is not a Controller, no need to look up slashing spans.
+		if (optionController.isNone) {
+			// if bonded.isNone, `stash` address param is not a stash.
```
danforbes, line 384: We may actually want to throw a 4xx in this case.

src/ApiHandler.ts
```diff
@@ -361,47 +361,63 @@ export default class ApiHandler {
...
+				staking: {},
 			};
 		}
 
-		const ledger = staking.unwrap(); // should always work if staking.isSome
-		const slashingSpans = await api.query.staking.slashingSpans.at(
-			hash,
-			ledger.stash
-		);
+		// we can infer (optionBonded.isSome === true) and safely unwrap
```
danforbes, line 394: Either justify this statement or add explicit error checking.

src/ApiHandler.ts
```diff
@@ -361,47 +361,63 @@ export default class ApiHandler {
...
 
-		let numSlashingSpans;
-		if (slashingSpans.isSome) {
-			const span = slashingSpans.unwrap();
-			numSlashingSpans = span.prior.length + 1;
-		} else {
-			numSlashingSpans = 0;
-		}
+		const [
+			optionStakingLedger,
```
danforbes, line 398: `stakingLedgerOption`...

src/ApiHandler.ts
```diff
@@ -361,47 +361,63 @@ export default class ApiHandler {
...
+			optionSlashingSpans,
+		] = await Promise.all([
+			await api.query.staking.ledger.at(hash, controller),
+			await api.query.staking.payee.at(hash, stash),
+			await api.query.staking.slashingSpans.at(hash, stash),
+		]);
+
+		// should always work because we know that we have a bonded pair but just
+		// in case we return a default value of null so we know to handle later
+		const stakingLedger = optionStakingLedger.unwrapOr(null);
```
danforbes, line 409: Would it be appropriate to check for `null` and throw an error if that's the case?

README.md
danforbes: Speaking in RESTful terms `/staking/ADDRESS` implies "get me the staking resource with this address" and that is not what we are modeling. It sounds like we are modeling something like `accounts/ADDRESS/staking` or something...

src/ApiHandler.ts
emostov: For both error comments, I agree. I setup up some rudimentary error handling which bubbles up the message and status code to get.

Do you think we should put in some slightly more robust error handling features in this PR?

src/ApiHandler.ts
emostov: Check out my modification to the above if statement and lmk if that makes sense.

README.md
emostov: I think that makes sense. I would just want to make it clear that we would _not_ support the route `accounts/` since that implies listing all the accounts or paginated listing of accounts. The reason being that I don't think there is a need for that and I am not sure if there is a clear path to support that.

joepetrowski: > The reason being that I don't think there is a need for that and I am not sure if there is a clear path to support that.

100%, there is no practical path to support this

src/ApiHandler.ts
```diff
@@ -361,47 +361,69 @@ export default class ApiHandler {
...
 
-		if (staking.isNone) {
-			// Address is not a Controller, no need to look up slashing spans.
-			return {
-				at,
-				staking: {},
-				numSlashingSpans: null,
+		if (
+			controllerOption.isNone === true ||
+			controllerOption.isSome !== true
```
joepetrowski, line 385: Why both? Doesn't one imply the other?

src/main.ts
```diff
@@ -245,19 +221,21 @@ async function main() {
 	//
 	// Substrate Reference:
 	// - Staking Pallet: https://crates.parity.io/pallet_staking/index.html
+	// - `RewardDestination`: https://crates.parity.io/pallet_staking/enum.RewardDestination.html
+	// - `Bonded`: https://crates.parity.io/pallet_staking/struct.Bonded.html
 	// - `StakingLedger`: https://crates.parity.io/pallet_staking/struct.StakingLedger.html
-	get('/staking/:address', async (params) => {
+	get('/accounts/:address/staking', async (params) => {
```
joepetrowski, line 227: I'm not really comfortable with this. Even if it's the direction we want to go, now this pattern is broken across all of sidecar. This should be a separate discussion around the sidecar API where we change everything.

README.md
joepetrowski: > Speaking in RESTful terms /staking/ADDRESS implies "get me the staking resource with this address" and that is not what we are modeling. It sounds like we are modeling something like accounts/ADDRESS/staking or something...

Thinking about this more in the context of the current/active era thing, I disagree with this model. I would not put era under artifacts, but it may be useful to have an endpoint like "info about staking" that's not related to an address. For example:

- `/staking`: Give me info about staking. Era, session, number of validators, next election window, block of next era/session change
- `staking/ADDRESS`: Give me staking information specific to this address

src/ApiHandler.ts
danforbes: Also, do not write `controllerOption.isNone === true`; it can just be `if (controllerOption.isNone)`.

src/ApiHandler.ts
```diff
@@ -361,47 +361,69 @@ export default class ApiHandler {
...
+		) {
+			// if controllerOption.isNone, `stash` address param is not a stash.
+			throw {
+				error: `The address ${stash} is not a stash address.`,
+				statusCode: 422,
```
danforbes, line 390: The simplest solution is usually the correct one.
```suggestion
				statusCode: 400,
```

src/ApiHandler.ts
```diff
@@ -361,47 +361,69 @@ export default class ApiHandler {
...
+			stakingLedgerOption,
+			rewardDestination,
+			slashingSpansOption,
+		] = await Promise.all([
+			await api.query.staking.ledger.at(hash, controller),
+			await api.query.staking.payee.at(hash, stash),
+			await api.query.staking.slashingSpans.at(hash, stash),
+		]);
+
+		// should always work because we know that we have a bonded pair
```
danforbes, line 406: This is a good comment but I would suggest moving inside your `if` statement and rephrasing it: `// should never throw because...`

README.md
emostov: Reverted to `/staking`! Hopefully will have a swagger PoC type thing up soon

src/ApiHandler.ts
```diff
@@ -361,47 +361,65 @@ export default class ApiHandler {
...
+			await api.query.staking.payee.at(hash, stash),
+			await api.query.staking.slashingSpans.at(hash, stash),
+		]);
+
+		const stakingLedger = stakingLedgerOption.unwrapOr(null);
+
+		if (stakingLedger === null) {
+			// should never throw because by time we get here we know we have a bonded pair
+			throw {
+				error: `Staking ledger could not be found for address "${controller.toString()}"`,
```
joepetrowski, line 408: Or just say stash, since that's what the user passed in.
```suggestion
				error: `Staking ledger could not be found for controller address "${controller.toString()}"`,
```

src/ApiHandler.ts
```diff
@@ -361,47 +361,65 @@ export default class ApiHandler {
...
 		};
 
-		if (staking.isNone) {
-			// Address is not a Controller, no need to look up slashing spans.
-			return {
-				at,
-				staking: {},
-				numSlashingSpans: null,
+		if (controllerOption.isNone) {
+			// if controllerOption.isNone, `stash` address param is not a stash.
```
danforbes, line 384: The code is now good enough that this comment is useless :rocket: 
```suggestion
			// if controllerOption.isNone, `stash` address param is not a stash.
```

Issue #81, opened 2020-06-12
Give useful Staking info


joepetrowski:
The `staking` endpoint should return a bunch of information that would inform a validator or nominator about whether or not (or when) they should make a transaction.

```
getStaking(blockId) -> {
  at: { blockNumber, hash },
  ValidatorCount: number, // of validators in the set
  ValidatorSet: Vec<AccountId>, // list of validators for this session
  ActiveEra: number, // ActiveEra.index
  ForceEra: enum, // status of era forcing
  NextEra: number, // block number or unix time of next era
  NextSession: number, // block number or unix time of next session
  UnappliedSlashes: Vec<UnnappliedSlash>, // array of upcoming slashes, keyed by era
  QueuedElected: Option<ElectionResult>, // contains an array of stashes elected
  ElectionStatus: { status, toggle }, // see below
}
```

The final item, `ElectionStatus` deserves particular discussion. Certain staking transactions that would affect the outcome of a validator are not allowed while the submission window for election results is open.

`staking.EraElectionStatus` returns either `Closed` or `Open(BlockNumber)` with the block number from which the window has been open.

Todo:
- Look up how to predict the end of an era (see Polkadot JS Apps).
- Look up when election opens and closes (IIRC, it's open for the last 1/16th of a session on Polkadot and 1/4 of a session on Kusama).

To return:
```
electionStatus:
  window: 'Open' | 'Closed',
  toggle: blockNumber
```
Where `blockNumber` is the future block when the window will change state. Note: as this is based on slots and not blocks, a toggle `UnixTime` might be more accurate than a block number.

PR #101, opened 2020-06-17
Add activeEra and currentEra to `fetchTxArtifacts`

### LINKED ISSUE #99: Give the exact era in numeric form under the path `tx/artifacts`
For discussion.

emostov:
Closes #99

### DIFF
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -460,12 +460,16 @@ export default class ApiHandler {
 			genesisHash,
 			name,
 			version,
+			currentEraOption,
+			activeEraOption,
 		] = await Promise.all([
 			api.rpc.chain.getHeader(hash),
 			api.rpc.state.getMetadata(hash),
 			api.rpc.chain.getBlockHash(0),
 			api.rpc.system.chain(),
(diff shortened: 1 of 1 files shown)


src/ApiHandler.ts
```diff
@@ -480,6 +484,15 @@ export default class ApiHandler {
 			specName: version.specName.toString(),
 			specVersion: version.specVersion,
 			txVersion: version.transactionVersion,
+			// TODO - when unwrap fails, do we want to throw or give an error just for this field?
```
danforbes, line 487: This sounds like a @shawntabrizi question to me. From my POV, this field can be `null`, but I'm not sure if that's the correct idiom to accurately model what's happening at the node level or whatever.

shawntabrizi: Active era and current era will only be null if the Staking system hasn't started yet. Throwing an error or assuming it is zero is probably fine here depending on what behavior you want.

joepetrowski:
Let's do this in a fresh branch for https://github.com/paritytech/substrate-api-sidecar/issues/81

PR #104, opened 2020-06-18
Support upcoming runtimes

athei:
This PR adds support for the [upcoming](https://github.com/paritytech/polkadot/pull/1263) polkadot, kusama and westend runtimes. Westend starting with runtime 10 is now also support but untested.

In addition to these changes I also added the following modifications:

- By building the calc-fee rust project with `FEE_DEBUG=1` in the environment you get some nice debugging output
- I moved the runtime switching logic to the calc-fee project.

### DIFF
--- a/calc-fee/src/lib.rs
+++ b/calc-fee/src/lib.rs
@@ -1,10 +1,11 @@
-mod panic;
+mod debug;
 
+use core::str::FromStr;
+use log::info;
 use serde_derive::Deserialize;
 use sp_arithmetic::{FixedI128, FixedPointNumber, Perbill};
 use sp_arithmetic_legacy::Fixed128 as Fixed128Legacy;
 use wasm_bindgen::prelude::*;
-use core::str::FromStr;
 
 type Balance = u128;
 type Weight = u64;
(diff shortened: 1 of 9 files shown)


calc-fee/src/lib.rs
```diff
@@ -18,48 +19,70 @@ pub struct JSCoefficient {
...
+			("kusama", v) if 1062 < v && v < 2011 => V1((new_i128(inner), false)),
+			("kusama", v) if 2011 <= v => V2(new_u128(inner)),
+
+			("westend", 10) => V0(new_legacy_128(inner)),
+			("westend", v) if 10 < v && v < 31 => V1((new_i128(inner), false)),
+			("westend", v) if 31 <= v => V2(new_u128(inner)),
+
+			_ => {
+				info!("Unsupported runtime: {}#{}", spec_name, spec_version);
+				return None;
```
emostov, line 55: This warning is a good idea, and I think it works well for now. It does make me think though that in the future we should maybe think about how to make _all_ of sidecar easily configurable for a wide array of substrate based chains. Polkadot-js already easily supports adding types, and I am wondering if we could do something of similar effect with fee stuff in this crate.

athei: Instead of bailing we could just use the current `Multiplier` as long as the chain supplied a valid `Polynomial`. That would instantly support all current chains that use the `ChargeTransactionPayment` (should be most of them).

So maybe we should rather explicitly ban unsupported versions of our runtimes than disallowing anything we do not know.

Issue #100, opened 2020-06-17
Fix sanitizeNumbers


emostov:
Currently `sanitizeNumbers` uses internal properties of Codec, which is not not reliable. Further it uses any for both param and return types.

Ideally `sanitizeNumbers` can gracefully handle options, and may possibly need logic specific to handling the types *Int, Option, Tuple, Vec, Struct, Result, *Map, *Set, and Enum.

A _very_ rough code snippet from @jacogr may give inspiration to (part of a) possible solution.
```
function sanitize (value: Codec): number | string | null {
  if (value instanceof AbstractInt) {
    return value.toString(10);
  } else if (value instanceof Option) {
    return value.isSome
      ?  sanitize(value.unwrap()) // make another pass, these can be nested
      : null;
  } else if (value instanceof Struct) {
   ...
  } else if (...) { ... }
}
```

A good fix to this could also encapsulate #89 by including unit tests.

PR #109, opened 2020-06-20
Give useful staking info 

### LINKED ISSUE #81: Give useful Staking info

emostov:
### Overview
Closes #81 

EDIT: `queuedElected` field removed. `queuedElected` is continuously updated while the election window is open, and is thus not desirable. Likely what would make sense is a `nextElected` field that shows the validator set for the `currentEra`.

N.B. when `forceEra.isForceNone` we assume the network is PoA (or some other mode where validators are not relevant) and thus we exclude all staking related fields except for slashes (in case of a weird scenario where it became ForceNone after validating for some period).

### Follow up PRs
- Compatibility with aura 
- Compatibility with older versions of babe that don't have history. For example use `currentEraStartSessionIndex` when `erasStartSessionIndex` doesn't exist. I think there are more areas in `deriveSessionAndEraProgress` that would need changes as well.
- Create a field `nextElected` that would be the `accountId`s from `staking.erasStakers.keys(currentEra)`, which would reflect upcoming validator set when `activeEra.index !== currentEra`
 - Unify error handling to use `errors-en.json` for error messages.
- Response types for all handler methods
- Constructors for response types
- Refactor sanitize numbers to handle rust types on case by case basis  and make sure plays friendly with response types definitions (see #100)
- create a configurable `epochDivisor` env for a more generic solution to `deriveElectionLookAhead`

---

#### Docs / comments
- [x] Docs in main above the route
- [x] Comments in ApiHandler function

#### Response values
- [x] at
- [x] validatorCount
- [x] validatorSet
- [x] activeEra(Index)
- [x] forceEra
- [x] nextEra
- [x] nextSession
- [x] unappliedSlashes
- electionStatus
    - [x] status
    - [x] toggle

#### Testing
- [x] equivocation behavior to test unappliedSlashes


joepetrowski:
> executing await api.query.staking.queuedElected.at(hash) fails an unreasonably high amount of the time with an error when trying to createType ElectionResult. I ended up breaking it into its own function and wrapping it in a try/catch block

This might only exist right after the election until the set change takes place. Check the Apps code for how Jaco determines "member of validator set in next session".

src/main.ts
```diff
@@ -212,6 +214,62 @@ async function main() {
...
+	//  	Each `UnappliedSlash` contains:
+	//		- `validator`: Stash accountId of the offending validator.
+	//		- `own`: The amount the validator will be slashed.
+	//		- `others`: Array of tuples of (accountId, amount) representing all the stashes of other
+	//		slashed stakers an and the amount they will be slashed.
+	//		- `reporters`: Array of accountIds of the reporters of the offence (bounty payout
+	//		recipients.)
+	//		- `payout`: Amount of bounty payout.
+	// - `queuedElected`: Array of accountIds for validators who have been elected. Only available
+	//		when the `ElectionResult` is available.
```
thiolliere, line 241: This is a bit missleading, `QueuedElected` hold the best phragmen solution yet submitted, but if a better solution is submitted next block then previous solution doesn't hold anymore.

src/ApiHandler.ts
```diff
@@ -644,4 +739,114 @@ export default class ApiHandler {
...
+			await api.query.babe.currentSlot.at(hash),
+			await api.query.babe.epochIndex.at(hash),
+			await api.query.babe.genesisSlot.at(hash),
+			await api.query.session.currentIndex.at(hash),
+			await api.query.staking.activeEra.at(hash),
+		]);
+
+		const { index: activeEra } = activeEraOption.unwrapOrDefault();
+
+		// TODO make sure this doesn't fail on querys older than history depth
```
thiolliere, line 775: it shouldn't: active era is (due to current pallet-session) at most one behind "current era" (considered as current era from the point of view of pallet-staking)
Thus active start will never be less than `current_era-history_depth`

kianenigma:
> > Executing await api.query.staking.queuedElected.at(hash) fails an unreasonably high amount of the time with an error when trying to createType ElectionResult. I ended up breaking it into its own function and wrapping it in a try/catch block.

Yes the storage should always be decodable, but it is most often `None`.

kianenigma:
> To compensate, we could instead create a field nextElected that would be the accountIds from staking.erasStakers.keys(currentEra), which I think would reflect upcoming validator set when activeEra.index != currentEra, however it is unclear to me if this would be appropriate.

If by compensate you mean compensate for the fact that `QueuedElected` is not decodable, then absolutely no. These two are not the same thing. QueuedElected stores the potentially next set and can change until the era is triggered. Once we have the new era, it is parsed and committed to `ErasStakers`. So they are not the same thing.

src/ApiHandler.ts
```diff
@@ -644,4 +739,114 @@ export default class ApiHandler {
...
+	 * function is useful in scenarios where it is necessary to gracefully handle
+	 * throws.
+	 *
+	 * @param api ApiPromise with ensured metadata.
+	 * @param hash block hash to get info at.
+	 */
+	private async tryFetchQueueElected(
+		api: ApiPromise,
+		hash: BlockHash
+	): Promise<ElectionResult | null> {
```
kianenigma, line 839: I think this is not needed and we need to sort out why queuedElected fails.

src/ApiHandler.ts
```diff
@@ -345,13 +352,101 @@ export default class ApiHandler {
...
+		const currentBlockNumber = number.toBn();
+
+		const nextSession = sessionLength
+			.sub(sessionProgress)
+			.add(currentBlockNumber);
+
+		// when ForceNone we are in PoA
+		const nextEra = forceEra.isForceNone
+			? null // If forceEra.isForceNone there is no next era (null)
+			: forceEra.isForceAlways
```
kianenigma, line 400: coding for a year or so mostly in rust, I've forgotten that this `? :` syntax even exists. I still think if else is more expressive though.

src/ApiHandler.ts
```diff
@@ -345,13 +352,101 @@ export default class ApiHandler {
...
+
+		return {
+			at: {
+				hash: hash.toJSON(),
+				height: currentBlockNumber.toString(10),
+			},
+			validatorCount: forceEra.isForceNone
+				? null
+				: validatorCount.toString(10),
+			activeEra: activeEra.toString(10) ?? null,
```
kianenigma, line 421: what does ?? do?

src/ApiHandler.ts
kianenigma: (you seem to use it a lot, nevermind then)

src/ApiHandler.ts
thiolliere: `(cond) ? true : false`

it reminds me C :-)

src/ApiHandler.ts
emostov: I prefer it in this case because it allows me to avoid using an undefined `let` over an if/else if/else block.

src/ApiHandler.ts
emostov: If the operator on the left is undefined or null, return the operator on the right. I use it often because I want to make sure our API is consistent in using null and we don't have an undefined sneak in.

src/ApiHandler.ts
emostov: Could active-era be an arbitrary amount behind current-era?

thiolliere: pallet-session only delay by one session so active-era is at least current-era - 1.

But in general if we change pallet session (I mean its code) and make it delaying by 3 sessions and new_era is trigger always then yes, active-era will be 3 index behind current-era.

PR #123, opened 2020-06-30
Setup Address Validation and Error Handling Middleware

### LINKED ISSUE #97: Validate address prior to calling polkadot.js api
When an address with improper length or non-base58 character is passed as a param an error occurs while polkadot.js tries to `createType(AccountId)`. This results in a 500 error. 

This can be avoided by validating the address is properly base58 encoded and has the correct length. If the request does not have a valid address a 4xx error can be thrown to indicate malformed user input and the error message can specify exactly what validation failed.

emostov:
Closes #97 

In order to do address validation as part of the middleware, it became apparent that it would be useful to have error handling middleware as well to refine the response. Additionally, I created a class that standardizes errors and ensures they are recognized by node as an `Error`.


tsconfig.json
```diff
@@ -29,6 +29,6 @@
     "./src/**/*.json",
     "./config/**/*.json",
     "./config/**/*.ts",
-    "./declarations/**/*.d.ts"
+    "./declarations/**/*.d.ts", "src/exceptions"
```
joepetrowski, line 32: new line?

src/exceptions/HttpException.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Base class for creating exceptions.
+ */
+export default class HttpException extends Error {
```
danforbes, line 4: Can we use something that exists as opposed to rolling our own? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/http-errors/index.d.ts

src/main.ts
```diff
@@ -63,8 +65,7 @@ async function main() {
 
 	function get(
 		path: string,
-		// eslint-disable-next-line @typescript-eslint/no-explicit-any
-		cb: (params: core.ParamsDictionary) => Promise<any>
+		cb: (params: core.ParamsDictionary) => Promise<any> // eslint-disable-line @typescript-eslint/no-explicit-any
```
danforbes, line 68: I prefer it the way it was before but this is purely stylistic.

PR #124, opened 2020-06-30
Validate block height

### LINKED ISSUE #67: Requesting a block that doesn't yet exist returns a 500
If you try an request a block by number that does not exist you get `{"error":"Interal Error"}`. I think it should instead give a more descriptive error, like `{"error":"requested block does not exist"}`. Maybe also have an appropriate 4xx status code.

An example of the issue would be to hook sidecar up to `wss://rpc.polkadot.io` and send GET `localhost:8080/block/{currentBlockNumber + 5}`.

emostov:
This pr builds on top of #123 and closes #67.

This should only be merged in after or instead of #123 since it updates some of the code introduced in that pr

On top of validating height, this PR also bumps us to Express v 5.0.0-alpha.8 which enables the app to automatically catch errors from async middleware and handlers using our error catching middleware.


danforbes:
Adding the diff on top of #123 https://github.com/paritytech/substrate-api-sidecar/compare/ss58-2...zeke-height-validation

src/App.ts
```diff
@@ -0,0 +1,53 @@
...
+) => void | Promise<void>;
+
+type AppConfiguration = {
+	controllers: BaseController[];
+	port: string;
+	preMiddleware: Middleware[];
+	postMiddleware: Middleware[];
+};
+
+export default class SidecarApp {
```
danforbes, line 20: Where is this used?

PR #131, opened 2020-06-30
Give correct error for too high block number

### LINKED ISSUE #67: Requesting a block that doesn't yet exist returns a 500

emostov:
Closes #67

Instead of handling this as its own validation in the middleware pipeline, I realized it makes more sense to do the validation in getHashForBlock. My reasoning is that proper validation requires a call to get the latest block. In getHashForBlock we do several less expensive checks prior and then only check for invalid height if the rpc query getBlockHash fails. Finally, some error catching is added in the get and post wrappers to check for the recently introduced HttpError.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -21,6 +21,7 @@ import { isHex } from '@polkadot/util';
 import * as bodyParser from 'body-parser';
 import * as express from 'express';
 import * as core from 'express-serve-static-core';
+import { BadRequest, HttpError } from 'http-errors';
 import * as morgan from 'morgan';
 
 import ApiHandler from './ApiHandler';
@@ -71,6 +72,16 @@ async function main() {
 				// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 				res.send(sanitizeNumbers(await cb(req.params)));
 			} catch (err) {
+				if (err instanceof HttpError) {
+					const code = err.status;
+					res.status(code).send({
+						code,
+						message: err?.message,
+						stack: err.stack,
+					});
+					return;
+				}
+
 				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 				if (err && typeof err.error === 'string') {
 					// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
@@ -535,9 +546,9 @@ async function getHashForBlock(
 	api: ApiPromise,
 	blockId: string
 ): Promise<BlockHash> {
-	try {
-		let blockNumber;
+	let blockNumber;
 
+	try {
 		const isHexStr = isHex(blockId);
 		if (isHexStr && blockId.length === 66) {
 			// This is a block hash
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -563,6 +574,15 @@ async function getHashForBlock(
 
 		return await api.rpc.chain.getBlockHash(blockNumber);
 	} catch (err) {
+		// Check if the block number is too high
+		const { number } = await api.rpc.chain.getHeader();
+		if (blockNumber && number.toNumber() < blockNumber) {
```
joepetrowski, line 579: Since you put `let blockNumber` before the `try`, will it always consider this truthy?

src/main.ts
```diff
@@ -563,6 +574,15 @@ async function getHashForBlock(
...
+			throw new BadRequest(
+				`Specified block number is larger than the current largest block. ` +
```
joepetrowski, line 581:
```suggestion
				`Specified block number is higher than the current finalized block height. ` +
```

src/main.ts
emostov: On line 549 `blockNumber` is equal to `undefined` - so it is a falsey value

src/main.ts
```diff
@@ -563,6 +574,15 @@ async function getHashForBlock(
 
 		return await api.rpc.chain.getBlockHash(blockNumber);
 	} catch (err) {
+		// Check if the block number is too high
+		const { number } = await api.rpc.chain.getHeader();
+		if (blockNumber && number.toNumber() < blockNumber) {
+			throw new BadRequest(
```
danforbes, line 580: This is a breaking change because we are now returning a different HTTP response code so I would suggest keeping this until v1.

emostov: I see the argument that it is a breaking change since we introduce a new error that potentially prevents an older error message from appearing. However I think there are two reasons this is ok right now. 1) we have a precedent for introducing these types of error handling changes pre v1 (see #123 as a recent example) and 2) it is replacing downstream error messages that I see as significantly less helpful.

danforbes: It's not the error _message_ that is really the problem, it's the error _code_. This could break integrations if people are expecting a `5xx` type error when the block number isn't right.

danforbes: I see what you mean about #123 though...

emostov: What do you think of changing it to a 500 but leaving the rest?

joepetrowski: I think our user group is small enough that we can ask them...

danforbes: If we _hadn't_ have already released #123 I would have gone with Zeke's suggestion but given the state of things, I think Joe's is more reasonable.

PR #134, opened 2020-07-02
Breakout main into controllers and app

### LINKED ISSUE #136: Refactor main out into an App class and controller classes

emostov:
Closes #136 

Overview:
In order to better handle code base growth and increased use of middleware, it has become apparent that having all of the routes, handlers, and middleware mounting in `main` was not sustainable. With that in consideration I propose breaking out our server into sub-components App, Controllers, and Services. This PR addresses the former 2. 

In this iteration, Controllers are both in charge of defining request handlers, and mounting those request handlers and necessary middleware to an express.Router. Keeping the routing and handlers together seems like the most straightforward option for now.

App coordinates the ordered mounting of application level middleware and the controllers routes.

ApiHandler has been left untouched, but in the future when functions like `fetchBlock` are broken up into smaller components to facilitate unit testing, it may make sense to break up ApiHandler into a few services for the sake of code organization as the file is getting large.

New Classes:
- App: Responsible for mounting middleware and controller routers.
- AbstractController: Base class for controllers. Include older helper functions to get hash from block and verify block number. Has new helper functions to sanitize and send responses, wrap async handlers, and wrap & mount async get handlers.
- Various controller classes - they where divided up somewhat arbitrarily, but I tried doing it on the basis of resource.

Future PRs:
- look into a way to incorporate sanitizedNumbers in the middleware pipeline
- changing APIHandler => {Base, Abstract}Service. Creating various sub services such as BlocksService
- switch everything to http-errors for uniformity

TODO
- [ ] Make sure no hanging responses due to poor async error handling
- [x] Test claims
- [x]  tx submit and tx estimate work on happy path
- [x] Function and variable naming patterns consistent throughout new code


src/App.ts
```diff
@@ -0,0 +1,90 @@
...
+		host,
+		port,
+	}: IAppConfiguration) {
+		this.app = express();
+		this.port = port;
+		this.host = host;
+
+		this.initMiddleware(preMiddleware);
+
+		// Setup a root route
```
joepetrowski, line 41:
```suggestion
		// Set up a root route
```

src/controllers/AbstractController.ts
```diff
@@ -0,0 +1,160 @@
...
+import {
+	IAddressNumberParams,
+	IAddressParam,
+	INumberParam,
+} from 'src/types/request_types';
+
+import sanitizeNumbers from './utils/sanitizeNumbers';
+
+type SidecarRequestHandler =
+	| RequestHandler<IAddressParam>
```
joepetrowski, line 16:
```suggestion
	RequestHandler<IAddressParam>
```

src/controllers/AbstractController.ts
```diff
@@ -0,0 +1,160 @@
...
+	get path(): string {
+		return this._path;
+	}
+
+	get router(): Router {
+		return this._router;
+	}
+
+	/**
+	 * Mount all controller handler methods on the classes private router.
```
joepetrowski, line 37:
```suggestion
	 * Mount all controller handler methods on the class's private router.
```

src/controllers/TxSubmitController.ts
```diff
@@ -0,0 +1,52 @@
...
+	 */
+	private txSubmit: RequestHandlerTx = async (req, res): Promise<void> => {
+		const { tx } = req.body;
+		if (!tx) {
+			throw {
+				error: 'Missing field `tx` on request body.',
+			};
+		}
+
+		TxSubmitController.sanitizedSend(res, await this.handler.submitTx(tx));
```
joepetrowski, line 50: Since this is _supposed_ to get a block hash back, the `sanitizedSend` call isn't going to change that, right?

src/main.ts
```diff
@@ -16,15 +16,21 @@
...
+import BalanceController from './controllers/BalanceController';
+import BlocksController from './controllers/BlocksController';
+import ClaimsController from './controllers/ClaimsController';
+import MetadataController from './controllers/MetadataController';
+import StakingController from './controllers/StakingController';
+import StakingInfoController from './controllers/StakingInfoController';
+import TxArtifactsController from './controllers/TxArtifactsController';
+import TxFeeEstimateController from './controllers/TxFeeEstimateController';
+import TxSubmitController from './controllers/TxSubmitController';
+import VestingController from './controllers/VestingController';
```
joepetrowski, line 33: Is there a better way to do this, like create a `./controllers/index.ts` and then `import * from ./controllers`?

src/types/request_types.ts
```diff
@@ -1,13 +1,31 @@
-import { Request } from 'express';
+import { RequestHandler } from 'express';
 import { ParamsDictionary, Query } from 'express-serve-static-core';
 
+/**
+ * Body for RequestHandlerTx. In other words, the body of a POST route that send a Tx.
```
joepetrowski, line 5:
```suggestion
 * Body for RequestHandlerTx. In other words, the body of a POST route that sends an encoded transaction.
```

src/types/response_types.ts
```diff
@@ -21,6 +21,12 @@ import {
 import U32 from '@polkadot/types/primitive/U32';
 import { AnyJson, Codec } from '@polkadot/types/types';
 
+export type AnyCodecAndJson =
+	| AnyJson
```
joepetrowski, line 25: Maybe this is a normal pattern in TS, feel free to accept or reject.
```suggestion
	AnyJson
```

src/controllers/AbstractController.ts
danforbes: Good eye!

src/controllers/TxSubmitController.ts
danforbes: I don't _believe_ so, but Zeke should confirm. It looks like `sanitizeSend` just sends the body through `sanitizeNumbers`, which doesn't seem to have any logic related to converting between block heights and hashes.

src/controllers/AbstractController.ts
emostov: This is actually an eslint/prettier thing

PR #143, opened 2020-07-02
Move error handling to middleware

### LINKED ISSUE #138: Use middleware to catch errors and sanitize response

emostov:
EDIT: Removed the sanitize numbers middleware. No monkey patching express functions for now.

Closes #138 

This PR is a combination of two related features that both affect the implementation of `get` and `post`. This is in preparation for merging #134 as that PR does not use a generic wrapper around express app.get and app.post.'

### Motivation 

Middleware functions are modular, can be put on routes/routers with fine grain control, and are individually tested-able.

- With our current setup of error handling we must have duplicate code across get and post. 
- The handling of different errors is intertwined and number sanitization is all done in one function, making good unit testing difficult. 
- When adding new types of error formats or phasing out old error formats there is a high cognitive over head with the current setup. Using middleware specific to an error formats will allow us to easily add or remove error formats from our codebase

- For number sanitization, having the functionality at the middleware level allows us to move it from the application level to the router level if we choose. The current layout I propose is at the application level as that mimics our current API

### What to look for while reviewing

- Make sure there are no hanging responses. Express does not natively catch errors thrown by async functions, so it is up to the try/catch wrappers in `get` and `post` to catch them.
- All error responses are the same as before. I tried to ensure that we handle all error formats the same way, but good to double check
- The modification of `res.send` is reasonable. Implementing this was tricking because 
    1) I needed to ensure the correct binding of `res` to the original `send` and 
     2) I had to do some generic typing that got a bit finicky since some of the types from @types/express had several `any` defaults and sanitizeNumbers is all `any`s


src/main.ts
```diff
@@ -112,19 +98,12 @@ async function main() {
...
-				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-				if (err && typeof err.error === 'string') {
-					res.status(500).send(sanitizeNumbers(err));
-					return;
-				}
-
-				console.error('Internal Error:', err);
-
-				res.status(500).send({ error: 'Internal Error' });
+				// Express way of passing an error to the error handling middleware
```
danforbes, line 105: I don't think this comment adds anything.

src/middleware/error_middleware.ts
```diff
@@ -1,16 +1,170 @@
...
+import { ErrorRequestHandler } from 'express';
+import { HttpError, InternalServerError } from 'http-errors';
+import * as HttpErrorConstructor from 'http-errors';
+
+import { IBasicError, ILegacyError, ITxError } from '../types/error_types';
+
+/**
+ * Handle HttpError instances.
+ *
+ * Should be put before middleware that handles Error, since HttpError
```
danforbes, line 10: This is a nice useful comment.

PR #146, opened 2020-07-03
Add explicit return types for public functions in ApiHandler

### LINKED ISSUE #145: Explicit return types for public ApiHandler functions

emostov:
Closes #145 

In an effort to clean up all the `eslint-disable` and just generally have better typescript code, I added types for all public functions in ApiHandler. 

New code is straightforward. The main thing I want to discuss is the naming conventions this PR establishes. 

In order to prevent naming collisions with polkadot-js types, I prefixed `I` in front of all the new interfaces I introduced. Just from my last month of writing TS I have seen this pattern fairly often while perusing the inter-webs, however not sure if this is the right use. I would say this establishes the precedent that all future interfaces should be prefixed with an `I`.

I realize that many of these don't have polkadot-js name space collisions, however for the sake of consistency I feel we should either go all or nothing.

### DIFF
--- a/src/ApiHandler.ts
+++ b/src/ApiHandler.ts
@@ -16,10 +16,15 @@
 
 import { ApiPromise } from '@polkadot/api';
 import { CalcFee } from '@polkadot/calc-fee';
-import { Struct } from '@polkadot/types';
+import { Metadata, Struct } from '@polkadot/types';
 import { getSpecTypes } from '@polkadot/types-known';
 import { GenericCall } from '@polkadot/types/generic';
-import { DispatchInfo, EraIndex } from '@polkadot/types/interfaces';
+import {
+	DispatchInfo,
+	EraIndex,
+	Hash,
+	RuntimeDispatchInfo,
+} from '@polkadot/types/interfaces';
 import { BlockHash } from '@polkadot/types/interfaces/chain';
 import { EventRecord } from '@polkadot/types/interfaces/system';
 import { u32 } from '@polkadot/types/primitive';
@@ -30,10 +35,14 @@ import * as BN from 'bn.js';
 
 import * as errors from '../config/errors-en.json';
 import {
-	IBlockResponse,
+	IAccountBalance,
+	IAccountStakingSummary,
+	IAccountVestingSummary,
+	IBlock,
 	ISanitizedCall,
 	ISanitizedEvent,
 	IStakingInfo,
+	ITxConstructionMaterial,
 } from './types/response_types';
 
 export default class ApiHandler {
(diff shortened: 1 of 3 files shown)


src/ApiHandler.ts
```diff
@@ -607,8 +610,7 @@ export default class ApiHandler {
 		}
 	}
 
-	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-	async submitTx(extrinsic: string) {
+	async submitTx(extrinsic: string): Promise<{ hash: Hash }> {
```
joepetrowski, line 613: But this does not always return this data type.

src/types/response_types.ts
```diff
@@ -19,3 +39,91 @@ export interface StakingInfo {
...
+	stateRoot: Hash;
+	extrinsicsRoot: Hash;
+	authorId: AccountId | undefined;
+	logs: ILog[];
+	onInitialize: IOnInitializeOrFinalize;
+	extrinsics: IExtrinsic[];
+	onFinalize: IOnInitializeOrFinalize;
+}
+
+export interface IAccountBalance {
```
joepetrowski, line 68:
```suggestion
export interface IAccountBalanceInfo {
```

src/types/response_types.ts
```diff
@@ -19,3 +39,91 @@ export interface StakingInfo {
...
+	[key: string]: unknown;
+}
+
+interface IExtrinsic {
+	method: string;
+	signature: ISignature | null;
+	nonce: Compact<Index>;
+	args: Codec[];
+	newArgs: ISanitizedArgs;
+}
```
joepetrowski, line 129: `tip, hash, info, events, success, paysFee`?

src/ApiHandler.ts
emostov: I think it always _returns_ this data type. Our errors are of a different format, but (from my understanding at least) since they are throws we do not include them in the data type

src/types/response_types.ts
emostov: Strange that `tsc` didn't complain here

src/types/response_types.ts
```diff
@@ -19,3 +39,95 @@ export interface StakingInfo {
...
+	method: string;
+	signature: ISignature | null;
+	nonce: Compact<Index>;
+	args: Codec[];
+	newArgs: ISanitizedArgs;
+	tip: Compact<Balance>;
+	hash: string;
+	events: ISanitizedEvent[];
+	paysFee: boolean | null;
+}
```
joepetrowski, line 133: Still need `success` and `info`
```
			return {
				method: `${method.sectionName}.${method.methodName}`,
				signature: isSigned ? { signature, signer } : null,
				nonce,
				args,
				newArgs: this.parseGenericCall(method).args,
				tip,
				hash,
				info: {},
				events: [] as SanitizedEvent[],
				success: defaultSuccess,
				// paysFee overrides to bool if `system.ExtrinsicSuccess|ExtrinsicFailed` event is present
				paysFee: null as null | boolean,
			};

PR #149, opened 2020-07-04
Add config validation

chevdor:
It is important to abort if the config is not valid.
This PR shows the compact version of the config if the config is ok, as info for the user.
However, if the config is NOT ok, it will show the full (= not compact) version to help the user figure out the issue then exit.

Here is a run where I am passing a bad W_URL:
```
$ NODE_ENV=web3 SAS_SUBSTRATE_WS_URL=BAD y start
yarn run v1.22.4
$ yarn run build && yarn run main
$ tsc
$ node ./build/src/main.js
SAS:
  📦 EXPRESS:
     ✅ LOG_MODE: errors
    Log level
 
     ✅ BIND_HOST: 127.0.0.1
    Network interface we bind to
 
     ✅ PORT: 8080
    Port of the web service
    regexp: ^\d{2,6}$
 
  📦 SUBSTRATE:
     ✅ NAME: Default Substrate Dev Node
    Name for our node endpoint
 
     ❌ WS_URL: BAD
    Websocket URL
    regexp: ^wss?:\/\/.*(:\d{4,5})?$
 
Your config is NOT valid, exiting
error Command failed with exit code 1.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@types/http-errors": "^1.6.3",
     "@types/morgan": "^1.9.1",
     "body-parser": "^1.19.0",
-    "confmgr": "^1.0.5",
+    "confmgr": "^1.0.6",
     "express": "^4.17.1",
     "http-errors": "^1.8.0",
     "morgan": "^1.10.0",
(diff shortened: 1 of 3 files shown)


chevdor:
The failure comes from the fact that confmgr now expect node v12 or above.
It this is an issue for you, I can check if I can do without the requirements. If you can leave with that, just bump up to 12+.

PR #154, opened 2020-07-07
Future proof `sanitizeNumbers`

### LINKED ISSUE #100: Fix sanitizeNumbers

emostov:
Closes #100 

In light of recent test coverage for `sanitizeNumbers`, I think this PR now makes sense to push through since known regressions are now documented in changes to the specs.

### Goal

The primary goal is to refactor sanitize numbers to not use any internals of polkadot-js (see `.raw` and `._raw`), not use `any`, and generally be more robust and maintainable.

Extending that, it should be able to handle any response Sidecar might send (errors, normal responses).

### Changes

- refactor of `sanitizeNumbers`
- create `polkadot-js` type folder with type guards and update effected imports.

### Design choices

- Right now I have it set up to be an `AnyJson` return type, which is ultimately achieved by type casting. I left in a check (that is rather expensive) that will `console.error` when a non-`AnyJson` value is about to be returned. I see a few options here 1) leave as is, 2) have it set to throw an error, 3) take out the check altogether, or 4) switch the return type to unknown and leave type checks to an external entity. Right now the two things documented that trip the error are BigInt and Symbol.
- There are some javascript types that it does not handle sensibly (see spec:`javascript types it cannot handle properly`) - most of them are a trivial fix that just involves another check in an `if` statement. Not sure if was overkill though to add them if they aren't expected and none of them completely break things.
- I broke up the sub function `sanitizeCodec` for readability and because those additional checks for non-codecs seemed unnecessary. However, I see a few places that could be slightly DRYed up by combining into one big function.

### What to look for when reviewing

- [ ] 1)  Is there anywhere that this function might break with Codec types or native JS types? For example, In the past we have had runtime failures while trying to map un-mappable types such as `Option`s, which looked mappable because the objects prototype chain had an Array.
- [ ] 2) Are there any important tests missing for native JS types and/or polkadot-js types?
- [ ] 3) Should we add support for types like `BigInt`, `Symbol`, `Number`, etc (see jest `javascript types it cannot handle properly`) 
- [ ] 4) Idiomatic code? Places that can be DRYed w/o performance hit? Sensible performance optimizations (almost every response goes through this function)? Readability/ maintainability?


src/controllers/AbstractController.ts
```diff
@@ -166,10 +167,7 @@ export default abstract class AbstractController {
 	 * @param res Response
 	 * @param body response body
 	 */
-	static sanitizedSend<T = unknown>(res: Response<T>, body: T): void {
-		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
danforbes, line 170: Yesssss this is the best kind of difffff

src/sanitize/sanitizeNumbers.ts
```diff
@@ -1,94 +1,140 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/no-unsafe-return */
```
danforbes, line 4: Woooooooo

src/sanitize/sanitizeNumbers.ts
```diff
@@ -1,94 +1,140 @@
...
 
-	if (data instanceof Enum) {
-		if (data.isBasic) {
-			return data.toJSON();
+	// // For the sake of API consistency this is staying out for now
+	// if (value instanceof Result) {
+	// 	return value.isOk
+	// 		? sanitizeNumbers(value.asOk)
+	// 		: sanitizeNumbers(value.asError);
+	// }
```
danforbes, line 48: I would avoid checking in comments like this. Consider leaving a `// TODO` instead.

src/sanitize/sanitizeNumbers.ts
```diff
@@ -99,58 +145,67 @@ export function sanitizeNumbers(data: any): any {
...
+ * is either a number or string.
+ *
+ * @param map Map | CodecMap
+ */
+function mapTypeSanitizeKeyValue(map: Map<unknown, unknown> | CodecMap) {
+	const jsonMap: AnyJson = {};
+
+	map.forEach((value: unknown, key: unknown) => {
+		const nonCodecKey = sanitizeNumbers(key);
+		if (
```
danforbes, line 196: I find this `if` statement very difficult to read. Edit: I just realized this is because you are using tabs instead of spaces...

PR #156, opened 2020-07-08
Add .env docker sample file

emostov:


.env-docker-sample
```diff
@@ -0,0 +1,5 @@
+SAS_SUBSTRATE_WS_URL=wss://rpc.polkadot.io
+
+SAS_SUBSTRATE_NAME="Parity Polkadot node"
```
joepetrowski, line 3:
```suggestion
SAS_SUBSTRATE_NAME="Parity Polkadot Public RPC Node"
```

joepetrowski:
I'm not sure if the name gets used anywhere, but it is the accurate name for that endpoint. One thing we could do is hardcode a list of known public nodes, and if you connect to one of them, print some message like, "You're connected to the {Parity, Web3 Foundation} public RPC node for {Polkadot, Kusama}. This is a great way to try out Sidecar, but please run your own nodes in production."

PR #158, opened 2020-07-08
Add chainName to server connecting message

emostov:
Screen Shot 2020-07-08 at 11 22 51 AM

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -55,9 +55,13 @@ async function main() {
 		types: config.CUSTOM_TYPES,
 	});
 
-	const chainName = await api.rpc.system.chain();
+	const [chainName, { implName }] = await Promise.all([
+		await api.rpc.system.chain(),
+		await api.rpc.state.getRuntimeVersion(),
+	]);
+
 	console.log(
-		`Connecting to ${chainName.toString()} via ${config.NAME} at ${
+		`Connected to chain ${chainName.toString()} with node implementation ${implName.toString()} at ${
 			config.WS_URL
 		}`
 	);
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -50,13 +50,18 @@ async function main() {
...
-	console.log(`Connecting to ${config.NAME} at ${config.WS_URL}`);
-
 	const api = await ApiPromise.create({
 		provider: new WsProvider(config.WS_URL),
 		types: config.CUSTOM_TYPES,
 	});
 
+	const chainName = await api.rpc.system.chain();
+	console.log(
+		`Connecting to ${chainName.toString()} via ${config.NAME} at ${
```
joepetrowski, line 60: At this point it's connected :)

Perhaps also change the `NAME` to be the client implementation name (e.g. [node-template](https://github.com/paritytech/substrate/blob/e824e8ab0fadec9949ebb8b9e14d98703d6b8d44/bin/node-template/runtime/src/lib.rs#L99) or [parity-polkadot](https://github.com/paritytech/polkadot/blob/edd8f3b22b36f850cece13a81aa8f6cb9f8532a3/runtime/polkadot/src/lib.rs#L92)). Then we get a message like, "Connected to Polkadot CC1 on the parity-polkadot client at ws://127.0.0.1:9944".
```suggestion
		`Connected to ${chainName.toString()} via ${config.NAME} at ${
```

emostov:
Screen Shot 2020-07-08 at 1 23 09 PM

We still have NAME as an env variable. In theory, NAME could be helpful if someone is managing multiple instances of sidecar.

PR #160, opened 2020-07-09
Add `sanitizeNumber` support for HashMap, js native Map, and BN.

emostov:
- Add support for 
    - Codec HashMap
    - BN
    - native js Map (alot of Codec is derivative of Map so I think this is good to support)
- DRY, reorganize test code with constants for commonly used numbers. No tests are removed, but some are moved for better organization.
- Add test coverage 
- Add future todos for areas of polkadot-js types that still need more coverage
- Document (with testing) native javascript types that are not supported

### DIFF
--- a/src/utils/test_util.ts
+++ b/src/utils/test_util.ts
@@ -24,6 +24,20 @@ function createKusamaRegistry(): TypeRegistry {
 
 export const kusamaRegistry = createKusamaRegistry();
 
+export const MAX_U128 = '340282366920938463463374607431768211455';
+
+export const MAX_I128 = '170141183460469231731687303715884105727';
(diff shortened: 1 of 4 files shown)


src/utils/test_util.ts
```diff
@@ -24,6 +24,20 @@ function createKusamaRegistry(): TypeRegistry {
...
+
+export const MAX_U64 = '18446744073709551615';
+
+export const MAX_I64 = '9223372036854775807';
+export const MIN_I64 = '-9223372036854775808';
+
+export const MAX_U32 = '4294967295';
+
+export const MIN_I32 = '-2147483648';
+export const MAX_I32 = '2147483647';
```
joepetrowski, line 40: swap?

PR #164, opened 2020-07-11
Add sanitizeNumber tests; Move sanitizeNumbers to sanitize folder

emostov:
- Clear out all the testing todos in `sanitizeNumbers`
- Move `sanitizeNumbers` to a file of its own name and into a new directory, `sanitize`. This is in anticipation that it will be used for more than just controller responses (.i.e error messages). 
- Other
    - deleted what should be the last remaining `NAME` reference
    - yarn seems to have automatically rearranged package.json in a different order than before. Tried changing back but yarn just changed it again


src/utils/test_util.ts
```diff
@@ -39,6 +39,16 @@ export const MAX_U32 = '4294967295';
 export const MAX_I32 = '2147483647';
 export const MIN_I32 = '-2147483648';
 
+export const MAX_U16 = '65535';
+
+export const MAX_I16 = '65535';
```
joepetrowski, line 44:
```suggestion
export const MAX_I16 = '32767';
```

src/sanitize/sanitizeNumbers.spec.ts
```diff
@@ -279,6 +334,43 @@ describe('sanitizeNumbers', () => {
...
+
+			it('converts i16', () => {
+				const z = new I16(kusamaRegistry, 0);
+				expect(sanitizeNumbers(z)).toBe('0');
+
+				const min = new I16(kusamaRegistry, MIN_I16);
+				expect(sanitizeNumbers(min)).toBe(MIN_I16);
+
+				const max = new I16(kusamaRegistry, MAX_I16);
+				expect(sanitizeNumbers(max)).toBe(MAX_I16);
```
joepetrowski, line 372: It's weird that `new I16()` worked when your `MAX_I16` was actually `MAX_U16`. @jacogr, is this something the `I16` constructor should check for?

jacogr: It should yes. So there is def. an unsigned check missing (as I recall we did do have a total bits check, but this assumes only unsigned, so in this case it would pass since the number of bits are correct, even if the value overflows. Which is not quite helpful when your code starts misbehaving)

PR #167, opened 2020-07-12
Testing and corrections for `getHashForblock`

### LINKED ISSUE #166: Unit tests for `getHashForBlock`

emostov:
Closes #166

The primary purpose of this PR is to unit test `getHashForBlock` in order to document its expected behavior. 

- In the process I identified a few edge cases where it did not work as expected (i.e. explicitly throws an error in the try block that the catch should propagate but instead overwrites with a different error). 
- As part of an ongoing effort to phase out older error formats, I replaced the older errors with `HttpError` derivatives. In some cases 500 statuses where changed to 400 (`BadRequest` is 400, `InternalServerError` is 500)
- In the case that someone passes a string that starts with `0x` but may have an odd length or invalid characters, we now throw a 400 error indicating that the hex string is invalid instead of the previous 500 error that simply said `Cannot get block hash for ${blockId}.`
- A `.catch` was added to `getHeader` in order to give a richer error message (otherwise it would just be 'Internal Error') in the event that the rpc call fails
- A side effect of my work was breaking out error types into into individual files grouped with `is` predicate type guards.


src/is/basicLegacyError.ts
```diff
@@ -0,0 +1,10 @@
+import { IBasicLegacyError } from 'src/types/error_types';
+
+/**
+ * Type guard to check if something is a subset of the interface BasicError.
+ *
+ * @param thing to check type of
+ */
+export function isBasicError(thing: unknown): thing is IBasicLegacyError {
```
danforbes, line 8: I would refactor this. I would put each error interface in its own file in `/src/types/errors` and then each one of these helper functions would be defined in the same file where the interface is defined. Is that possible?

src/is/legacyError.ts
```diff
@@ -0,0 +1,13 @@
+import { ILegacyError } from '../types/error_types';
```
danforbes, line 1: There's a lot of mixing of file naming conventions going on here. I see some files `NamedLikeThis` others `namedLikeThis` and some others `named_liked_this`...at some point we should refactor for consistency. I like for files to be `named-liked-this` :sweat_smile:

emostov: The rough naming convention I have been trying to follow is:
`NamedLikeThis` => exports a class of the same name
`namedLikeThis` => exports a function of the same name
`named_like_this` => exports several things

Definitely up to change, just not sure what the strategy should be.

danforbes: This looks like a very good standard to follow, despite the fact that it contradicts my preferences :sweat_smile: https://basarat.gitbook.io/typescript/styleguide#filename

emostov: This too: https://github.com/airbnb/javascript#naming--filename-matches-export

emostov: I will save this for a later PR, probably combined with a broader reorganizing of files to separate things out more.

PR #184, opened 2020-07-22
Test `ensureMeta`; De-pluralize transaction named classes

### LINKED ISSUE #177: Unit test `ensureMeta`
Unit test `ApiHandler`'s `ensureMeta`.

emostov:
Closes #177 

- Add unit tests for `AbstractService.ensureMeta`
- Change class names prefixed with `Transactions` to `Transaction`


src/services/AbstractService.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+			expect(api.registry.register).toHaveBeenCalledTimes(1);
+			expect(api.registry.register).toBeCalledWith({
+				mock: 'specTypes!',
+			});
+
+			api.rpc.state.getRuntimeVersion = getRuntimeVersion;
+			api.registry.register = register;
+		});
+
+		it('resets metadata and just changes `txVersion` when just the `txVersion` has changed', async () => {
```
joepetrowski, line 216: This _should_ never happen, and there are guards within frame-system to prevent it, so it would basically be negligence on the part of governance, but it's OK as a test.

PR #191, opened 2020-07-25
Update docker setup

emostov:
The previous docker flow we provided was out of date and incomplete. This fixes those issues and also provides a suggested way (`.env.docker`) to configure Sidecar within the image, although I am not sure if it is necessary to provide such an affordance.


README.md
```diff
@@ -161,7 +161,7 @@ yarn build:docker
 
 ```bash
 export VERSION=`cat package.json | jq -r .version`
-docker run --rm -it -p 8080:8080 substrate-api-sidecar:$VERSION
+docker run --rm -it -p 127.0.0.1:8080:80 substrate-api-sidecar:$VERSION
```
chevdor, line 164: - `:80` is usually used for UIs btw. You may pick 8000, 8080, or 3000 by default.
- While `127.0.0.1:8080:80` is safer, I would recommend simply `8080:80` which will work 'easier' for most people. You may point to an article about making docker infrastructure safer so we don't need to do it here.

I suggest to add a docker version for the non dev that shows up first in the doc:
```
docker run --rm -it -p 8080:80 w3f/substrate-api-sidecar
```

README.md
```diff
@@ -170,6 +170,10 @@ then you can test with:
 curl -s http://127.0.0.1:8080/block | jq
 ```
 
+---
+
+Consult comments in `.env.docker` for more info on setting environment variables.
```
chevdor, line 175: I would point to the specs.yml so you don't need to repeat what is in there.

.dockerignore
```diff
@@ -1,6 +1,8 @@
 .env*
-!.env-sample
+!.env.docker
```
chevdor, line 2: We do not need `.env.docker` inside the docker image so it is find ignoring it.
So you may remove the line requesting not to ignore it :)

The `.env.docker` content will/may be passed to the docker container using:
```
docker run --rm -it --env-file .env.docker <...>
```
=> it comes from outside the container and prevents having to pass many `-e ...` chunks in the docker command.

README.md
emostov: Where is the name `w3f/substrate-api-sidecar` specified?

.gitignore
```diff
@@ -5,5 +5,6 @@
 yarn-error.log
 **/*.rs.bk
 .env.*
-!.env-sample
+!.env.local
```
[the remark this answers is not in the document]
emostov: Got it. That will be good for me to keep in mind

.env.docker
```diff
@@ -0,0 +1,25 @@
+# This file can be used to declare environment variables.
+# N.B. the image will use the contents of this file from when it was built.
+#
+# After building the docker image, one way to use this file is by running:
+#   docker run -e NODE_ENV=docker --rm -it -p 127.0.0.1:8080:80 substrate-api-sidecar:$VERSION
+#
+# If you would like to build an image with an .env file of a different name
+# simply add `!.env.myfile` to the .dockerignore.
```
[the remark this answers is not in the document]
emostov: Sounds like I did not understand correctly I we do not actually need `.env` files in the image so I am just going to take out those comments.

README.md
emostov: For the point about not setting it `127.0.0.1`, what is the downside? I was thinking it would make other, non-docker, documentation easier to reason through since the it would always be the same/

For your last point about the docker version for non dev, could we just default to having the build command name it `substrate-api-sidecar`? I think it might be confusing to have two different naming schemes since we would then need to have different commands just to build images with different names.

.dockerignore
emostov: Good to know, I have not seen the `--env-file` option before.

.env.docker
```diff
@@ -0,0 +1,25 @@
...
+# Prior to building the Docker image you can set the ws url here. If it is not
+# set it will fallback to the default value specified in `specs.yaml` or a
+# value specified when the image is run.
+#
+# SAS_SUBSTRATE_WS_URL=wss://{my-node-ws-url}
+
+# The below values are the same as in the Dockerfile and are simply here for reference.
+#
+# Host within the containers virtual network
+# SAS_EXPRESS_BIND_HOST=0.0.0.0
```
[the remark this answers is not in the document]
emostov: I uncommented to make it clear that we need that value, but it is also the same as in the Dockerfile.

.env.docker
```diff
@@ -0,0 +1,25 @@
...
+#   docker run -e NODE_ENV=docker --rm -it -p 127.0.0.1:8080:80 substrate-api-sidecar:$VERSION
+#
+# If you would like to build an image with an .env file of a different name
+# simply add `!.env.myfile` to the .dockerignore.
+
+# Prior to building the Docker image you can set the ws url here. If it is not
+# set it will fallback to the default value specified in `specs.yaml` or a
+# value specified when the image is run.
+#
+# SAS_SUBSTRATE_WS_URL=wss://{my-node-ws-url}
```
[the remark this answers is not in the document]
emostov: I am leaving this for now. I prefer to keep default values in the `specs.yaml` so people can have the first start up of Sidecar be easier, although I am open to change if there is a compelling UX reason.

README.md
emostov: I guess maybe a better question to ask is what is the utility of having the image named with the version? Is this a common idiom?

.env.docker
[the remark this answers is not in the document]
chevdor: I think the here is that you assume that people will have a node running on localhost:9944 which will likely not be the case. So it will fail, and people will not know where it comes from.

README.md
chevdor: > Where is the name w3f/substrate-api-sidecar specified?

I found it here: https://hub.docker.com/r/web3f/substrate-api-sidecar

> For the point about not setting it 127.0.0.1, what is the downside? 

You should not assume that people use Docker on their local machine. In production, this docker container will run on a server. This 127.0.0.1 means that only services running on the server will be able to access sidecar.

> For your last point about the docker version for non dev, could we just default to having the build command name it substrate-api-sidecar?

Yes, you can! This is my second suggestion. The naming scheme is not up to us, this is docker standard: `<user>/<image>:<tag>`.

.env.docker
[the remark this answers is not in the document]
emostov: @joepetrowski and @danforbes - any opinion on if the default node ws url should be kept as local node or should be switched. to be empty?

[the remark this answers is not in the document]
joepetrowski: I do think most people will run sidecar on the machine where they host a node. I think leaving this as local and an error message saying that it can't find a node running, go to ... and do ... to fix it.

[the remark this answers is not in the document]
danforbes: I don't think it's safe to assume people will run Sidecar on the same machine as the node. On the contrary, I would assume that most operators would want to make sure that the load on Sidecar did not impact the performance of the node.

README.md
emostov: Great - removed the localhost ip and and updated the build and run command to not use version for simplicity.

PR #217, opened 2020-08-17
Implement `accounts/{accountId}/staking-payouts`

### LINKED ISSUE #216: Implement `account/{accountId}/staking-payouts`
Implement `account/{accountId}/staking-payouts` as defined in `openapi-proposals.yaml`.

emostov:
Closes #216 

In this PR we add `accounts/{accountId}/staking-payouts`, and Rust code to do payout calculations. The `@polkadot/calc-fee` wasm-pack built package becomes `@substrate/calc` as it now incorporates both fee calculation and payout calculation.

The main things to consider for this PR:
- Payout amounts match on-chain so this endpoint can be used for accounting
- No type issues going between JS - wasm. (I assume Perbill's like validator_commission fit into a u32, however if they don't we will run into overflow)
- API is designed in such a way that is maintainable and we will not need to make breaking changes in the future.

TODO:
- [x] Correct arithmetic methods to simulate substrate's Perbill
- [ ] Verify why 0% commissions still show up as one Perbill
- [x] General testing to verify accuracy of results
- [x] Update `openapi-proposal.yaml` to match

Follow up PRs
- [ ] Unit tests


calc-payout/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct CalcPayout {
+    total_reward_points: RewardPoint,
+}
+
+#[wasm_bindgen]
+impl CalcPayout {
+    pub fn from_params(total_reward_points: &str) -> CalcPayout {
```
athei, line 19:
```suggestion
    pub fn from_params(total_reward_points: &str) -> Self {
```

calc-payout/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        debug::setup();
+
+        info!("from_params({}) -> ", total_reward_points);
+
+        let calc_payout = Self {
+            total_reward_points: RewardPoint::from_str(total_reward_points).unwrap(),
```
athei, line 25: You can pass a `u32` (any integer below `u128`)  natively without going through a `&str` or any serialisation using `JsValue`. You already have done that in `calc_payout`.

calc-payout/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        calc_payout
+    }
+
+    pub fn calc_payout(
+        &self,
+        validator_reward_points: RewardPoint,
+        era_payout: &str,
+        validator_commission: &str,
+        nominator_exposure: &str,
+        total_exposure: &str,
```
athei, line 39: Please consider if some of these values can be passed natively.

athei (commented):
Only reviewed the Rust part. Looks generally good to me. However, I thought maybe it would be better to put all Rust stuff in one package and give it a more general name rather than having one per function. Otherwise this creates a lot of code duplication and friction when adding new stuff.

You could just put those functionalities into different submodules.

README.md
```diff
@@ -63,14 +63,14 @@ If you are connecting to [Substrate Node Template](https://github.com/substrate-
 }
 ```
 
-## Fee Calculation Debugging
+## Fee Calculation & Payout Calculation Debugging
```
danforbes, line 66:
```suggestion
## Debugging Fee & Payout Calculations
```

calc/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 version = "0.1.0"
-authors = ["Alexander Theißen <alexander.theissen@parity.io>"]
+authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 license = "Apache-2.0"
 publish = false
-repository = "https://github.com/polkadot-js/wasm/"
-description = "Calculate extrinsic fees off-chain"
+repository = "https://github.com/paritytech/substrate-api-sidecar"
+description = "Off-chain calculations for substrate"
```
danforbes, line 9:
```suggestion
description = "Off-chain calculations for Substrate Sidecar."
```

calc/README.md
```diff
@@ -0,0 +1 @@
+# calc
```
danforbes, line 1: I would suggest that you either delete this file or replace it with something more useful.

calc/src/calc_fee.rs
```diff
@@ -0,0 +1,203 @@
+use crate::debug;
```
kianenigma, line 1: I think you should have used git mv somewhere to avoid this. These are not new, right?

calc/src/calc_payout.rs
```diff
@@ -0,0 +1,90 @@
...
+            total_reward_points: total_reward_points as RewardPoint,
+            era_payout: Balance::from_str(era_payout).unwrap(),
+        };
+
+        info!("calc_payout: {:#?}", calc_payout);
+
+        calc_payout
+    }
+
+    pub fn calc_payout(
```
kianenigma, line 34: You could look into making a PR to substrate and so that pallet-staking exposes a function that only does the calculation, and then you simply call into that here, rather than duplicating code. 

Given that it can be done with sensible changes to staking, I will totally support this.

calc/src/calc_payout.rs
```diff
@@ -0,0 +1,90 @@
...
+        &self,
+        validator_reward_points: RewardPoint,
+        validator_commission: u32,
+        nominator_exposure: &str,
+        total_exposure: &str,
+        is_validator: bool,
+    ) -> String {
```
kianenigma, line 41:
```suggestion
    ) -> String {Why do you return a string 🤔 
```

calc/src/debug.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "debug")]
+pub fn setup() {
```
kianenigma, line 2: nice 👍

openapi/openapi-proposal.yaml
```diff
@@ -116,33 +116,38 @@ paths:
       description: Returns payout information for the last specified eras. If
         specifying both the depth and era query params, this endpoint will
         return information for (era - depth) through era. (i.e. if depth=5 and
-        era=20 information will be returned for eras 16 through 20). N.B. non-archive
-        nodes cannot query eras less then `current_era - HISTORY_DEPTH`.
+        era=20 information will be returned for eras 16 through 20). N.B. You
+        cannot query eras less then `current_era - HISTORY_DEPTH`. N.B. The 
+        `nominator*` fields correspond to the address being queried, even if it
+        is a validator's _stash_ address. This is because a validator is technically
+        nominating themself.
```
joepetrowski, line 123:
```suggestion
        nominating itself.
```

src/controllers/accounts/AccountsStakingPayoutsController.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Description:
+ * Returns payout information for the last specified eras. If specifying both
+ * the depth and era query params, this endpoint will return information for
+ * (era - depth) through era. (i.e. if depth=5 and era=20 information will be
+ * returned for eras 16 through 20). N.B. You cannot query eras less then
+ * `current_era - HISTORY_DEPTH`.
+ *
+ * N.B. The `nominator*` fields correspond to the address being queried, even if it
+ * is a validator's _stash_ address. This is because a validator is technically
+ * nominating themself.
```
joepetrowski, line 52:
```suggestion
 * nominating itself.
```

joepetrowski (approved):

PR #222, opened 2020-08-18
Add `noMeta` query param to `tx/artifacts`

### LINKED ISSUE #221: Add query param for metadata inclusion in the /tx/artifacts endpoint
The inclusion of the metadata can slow down a call to the `/tx/artifacts` endpoint by a few seconds. For the ux experience of wallets, it would nice to have the ability to not include this when specifying an optional query parameter, such as `noMetadata`.

emostov:
Closes #221

### DIFF
--- a/src/controllers/transaction/TransactionMaterialController.ts
+++ b/src/controllers/transaction/TransactionMaterialController.ts
@@ -13,6 +13,10 @@ import AbstractController from '../AbstractController';
  * - (Optional) `number`: Block hash or number at which to query. If not provided, queries
  *   finalized head.
  *
+ * Query
+ * - (Optional) `noMeta`: If true, does not return metadata hex. This is useful when metadata is not
+ * needed and response time is concern. Defaults to false.
+ *
  * Returns:
  * - `at`: Block number and hash at which the call was made.
  * - `genesisHash`: The hash of the chain's genesis block.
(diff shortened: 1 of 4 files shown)


src/controllers/transaction/TransactionMaterialController.ts
```diff
@@ -56,14 +60,16 @@ export default class TransactionMaterialController extends AbstractController<
 	 * @param res Express Response
 	 */
 	private getTxArtifacts: RequestHandler = async (
-		_req,
+		{ query: { noMeta } },
 		res
 	): Promise<void> => {
 		const hash = await this.api.rpc.chain.getFinalizedHead();
 
+		const noMetaArg = noMeta === 'true' ? noMeta : undefined;
```
danforbes, line 68: Is there not some way to convert this to a proper boolean using the Express middleware or something?

src/services/transaction/TransactionMaterialService.ts
```diff
@@ -11,10 +11,34 @@ export class TransactionMaterialService extends AbstractService {
 	 * @param hash `BlockHash` to make call at
 	 */
 	async fetchTransactionMaterial(
-		hash: BlockHash
+		hash: BlockHash,
+		noMeta?: string
```
danforbes, line 15: Would suggest making this a boolean, so then the `if` statement can be rewritten as `if (noMeta)`, which I think is cleaner.

src/controllers/transaction/TransactionMaterialController.ts
emostov: Converted to a bool, right here. I think changing the query param to be a boolean in the middleware might be to much given we just need it in this service and its a one liner

joepetrowski (approved):

Issue #227, opened 2020-08-21
Implement V1 endpoints


emostov:
## Initial v1.0

- [x] /accounts/{accountId}/balance-info (conversion of existing endpoint)  (#248)
- [x] /accounts/{accountId}/staking-info (conversion of existing endpoint)  (#248)
- [x] /accounts/{accountId}/staking-payouts (#217)
- [x] /accounts/{accountId}/vesting-info (conversion of existing endpoint)  (#248)
- [x] /blocks/{blockId} (conversion of existing endpoint)
- [x] /blocks/head (conversion of existing endpoint)
- [x] /node/network (#226)
- [x] /node/transaction-pool (#226)
- [x] /node/version (#226)
- [x] /transaction (conversion of existing endpoint)  (#246)
- [x] /transaction/dry-run (#236)
- [x] /transaction/fee-estimate (conversion of existing endpoint)  (#246)
- [x] /transaction/material (conversion of existing endpoint) (#246)
- [x] /runtime/code (https://github.com/paritytech/substrate-api-sidecar/pull/228)
- [x] /runtime/spec (https://github.com/paritytech/substrate-api-sidecar/pull/228)
- [x] /runtime/metadata (conversion of existing endpoint) (#242)
- [x] /pallets/staking/progress (conversion of existing endpoint) (#248)
- [x] /pallets/{palletId}/storage/{storageItemId}

### Other
- [x] hosted swagger ui (#239)

PR #228, opened 2020-08-22
Implement `runtime/code` & `runtime/spec`

emostov:
#227 

Implements `runtime/code` & `runtime/spec` as specified in `openapi-proposal.yaml`.


src/services/runtime/RuntimeCodeService.ts
```diff
@@ -0,0 +1,32 @@
+import { Option, Raw } from '@polkadot/types';
+import { BlockHash } from '@polkadot/types/interfaces';
+import { IMetadataCode } from 'src/types/responses';
+
+import { AbstractService } from '../AbstractService';
+
+// https://github.com/shawntabrizi/substrate-graph-benchmarks/blob/master/js/extensions/known-keys.js#L21
```
joepetrowski, line 7:
```suggestion
// https://github.com/shawntabrizi/substrate-graph-benchmarks/blob/ae9b82f/js/extensions/known-keys.js#L21
```

joepetrowski (approved):

PR #232, opened 2020-08-24
Add configuration options for using metadata of parent block [PARENT_VERSION]

### LINKED ISSUE #231: `ensureMeta` should use runtime metadata from parent block
Currently `ensureMeta` re-sets the metadata if the current blocks spec version is different then the registries spec version. Recently though, [through this comment,](https://github.com/polkadot-js/api/issues/2507#issuecomment-678823180) it came up that we should be using the spec version of the parent block because the extrinsics inside the block are executed with the previous spec version.

A potential solution would likely be some variation of just passing the header of the block the call is made `at` to the service, which could then pass the header to ensureMeta so the specVersion could be fetched with `api.rpc.state.getRuntimeVersion(header.parentHash)`.  

It is worth noting here that `scheduler` normally calls `set_code` in `on_initialize` and due to the Wasm blob's size no  other extrinsics are able to make their way into a block. So there should be a fairly limited subsets of blocks where this edge case applies.

emostov:
Fixes #231

## Issue

Runtime upgrades are done by passing a massive Wasm blob of the runtime code to the dispatchable `set_code`. The block that processes `set_code` will have the runtime version information of the passed in Wasm blob, meaning the version of the block will be different than the version of its parent. Despite this new versioning, any extrinsics within the block are executed using the old runtime version.

In most circumstances no other extrinsics will make there way into a block with `set_code` for a combination of reasons:

1) The `set_code` call with the Wasm blob is (under normal circumstances) always well over the weight limit of the block, meaning after it is executed there will not be any more room for other extrinsics in the block.

2) There is an edge case where some other extrinsic(s) are put into the block prior to `set_code` being called. With chains that use the `scheduler` pallet for runtime upgrades this should not be an issue as the call to `set_code` is done in `on_initialize`, thus it should be executed before other extrinsics and leave no extra room. (N.B. this is not totally guaranteed, so use this info at your discretion) 

* Worth noting here that `scheduler` pallet is used by Polkadot, newer versions of Kusama, and others.

## Presented solution

At the beginning of almost every public service function `ensureMeta` is called to ensure that the polkadot-js registry has the proper metadata to decode the scale codec. In order to fix this issue, `ensureMeta` needs to know the version of the parent block. To get this information reliably `ensureMeta` needs 2 additional, synchronously executed, RPC calls. On my machine they added ~200ms, which is ~200ms for almost ever single Sidecar route.

Due to this performance degradation, this solution proposes an approach that allows a user to either
1) Opt-in to fetching the parent blocks version info for every block (see `SAS_SUBSTRATE_PARENT_VERSION=on`)
2) Fetch the parent blocks version info for blocks with hashes the are listed in a dedicated config file (see `config/setCodeBlocks.ts`). Maintainers and developers of various substrate chains may make PRs to add block hashes to the config in the standard code shipment. The user can add block hashes themselves as well to their own `config/setCodeBlocks.ts`.

It will be up to users to determine what option fits there use case, system setup, and priorities best. e.g. users may want to test how much time penalty there is for `SAS_SUBSTRATE_PARENT_VERSION=on` on there systems and make a judgement with that in mind.

## Downsides

The major downfall of this current implementation, is that for the aforementioned option 2 one needs a block hash, which can only be known once the block is made (and gossiped). This means that users will have to manually add the block hash to the config after seeing it and then restart Sidecar in order to read some queries made at a runtime upgrade block.

I can think of a few potential solutions to get around this, but they would entail significant RPC traffic to determine if `set_code` happened at a given block height, and therefore are not desirable.

## TODO:
- [ ] Add block hashes to `config/setCodeBlocks.ts`
- [ ] List versions of Kusama/Westend that are known not to use the scheduler pallet so we users can have a warning

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,15 @@ Following ENV variables can be set:
     setting it to anything else. LOG_MODE defaults to only "errors".
 -   `SAS_SUBSTRATE_WS_URL`: WebSocket URL to which the RPC proxy will attempt to connect to, defaults to
     `ws://127.0.0.1:9944`.
+-   `SAS_SUBSTRATE_PARENT_VERSION`: wether or not to always use the version of the parent block to set
+    metadata for scale decoding. This will normally have a slight time penalty, but will guarantee that
+    extrinsics within a block that does a runtime upgrade with `set_code` are always decoded correctly.
+    If set to `off`, one can selectively hard code the hash of blocks to use the parents version of to
+    set the metadata for scale decoding. The hash for block's with `set_code` can be hardcoded in
+    `config/setCodeBlocks.ts`. Defaults to false. Note: for chains that use the `scheduler` pallet to
+    do runtime upgrades, it is very unlikely this will be an issue as `scheduler` ensures that `set_code`
+    is executed in `on_initialize` and it is highly unlikely there would be any additional room in a
+    block for other extrinsics.
 
 If you are connecting to [Substrate Node Template](https://github.com/substrate-developer-hub/substrate-node-template), please add the following custom types in `config/types.json`.
 
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ Following ENV variables can be set:
 -   `SAS_SUBSTRATE_WS_URL`: WebSocket URL to which the RPC proxy will attempt to connect to, defaults to
     `ws://127.0.0.1:9944`.
 -   `SAS_SUBSTRATE_PARENT_VERSION`: always use the version of the parent block to set
-    metadata for scale decoding. `on` means always fetch and use the parent blocks metadata, while `off`
+    metadata for scale decoding. `on` means always fetch and use the parent block's metadata, while `off`
     means _only_ fetch and use the parent blocks metadata for the blocks with hashes specified in
-    `config/setCodeBlocks.ts`. Fetching and using the parent blocks metadata will have a slight time
+    `config/setCodeBlocks.ts`. Fetching and using the parent block's metadata will have a slight time
     penalty, but will guarantee that extrinsics within a block that executed a runtime upgrade with
     `set_code` are always decoded correctly. Defaults to `off`. Note: for chains that use the
     `scheduler` pallet to do runtime upgrades, it is very unlikely this will be an issue as
--- a/README.md
+++ b/README.md
@@ -51,15 +51,16 @@ Following ENV variables can be set:
     setting it to anything else. LOG_MODE defaults to only "errors".
 -   `SAS_SUBSTRATE_WS_URL`: WebSocket URL to which the RPC proxy will attempt to connect to, defaults to
     `ws://127.0.0.1:9944`.
--   `SAS_SUBSTRATE_PARENT_VERSION`: wether or not to always use the version of the parent block to set
-    metadata for scale decoding. This will normally have a slight time penalty, but will guarantee that
-    extrinsics within a block that does a runtime upgrade with `set_code` are always decoded correctly.
-    If set to `off`, one can selectively hard code the hash of blocks to use the parents version of to
-    set the metadata for scale decoding. The hash for block's with `set_code` can be hardcoded in
-    `config/setCodeBlocks.ts`. Defaults to false. Note: for chains that use the `scheduler` pallet to
-    do runtime upgrades, it is very unlikely this will be an issue as `scheduler` ensures that `set_code`
-    is executed in `on_initialize` and it is highly unlikely there would be any additional room in a
-    block for other extrinsics.
+-   `SAS_SUBSTRATE_PARENT_VERSION`: always use the version of the parent block to set
+    metadata for scale decoding. `on` means always fetch and use the parent blocks metadata, while `off`
+    means _only_ fetch and use the parent blocks metadata for the blocks with hashes specified in
+    `config/setCodeBlocks.ts`. Fetching and using the parent blocks metadata will have a slight time
+    penalty, but will guarantee that extrinsics within a block that executed a runtime upgrade with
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -51,6 +51,16 @@ Following ENV variables can be set:
     setting it to anything else. LOG_MODE defaults to only "errors".
 -   `SAS_SUBSTRATE_WS_URL`: WebSocket URL to which the RPC proxy will attempt to connect to, defaults to
     `ws://127.0.0.1:9944`.
+-   `SAS_SUBSTRATE_PARENT_VERSION`: always use the version of the parent block to set
+    metadata for scale decoding. `on` means always fetch and use the parent blocks metadata, while `off`
+    means _only_ fetch and use the parent blocks metadata for the blocks with hashes specified in
+    `config/setCodeBlocks.ts`. Fetching and using the parent blocks metadata will have a slight time
```
joepetrowski, line 57:
```suggestion
    metadata for scale decoding. `on` means always fetch and use the parent block's metadata, while `off`
    means _only_ fetch and use the parent blocks metadata for the blocks with hashes specified in
    `config/setCodeBlocks.ts`. Fetching and using the parent block's metadata will have a slight time
```

dvdplm:
Maybe you could combine the two approaches, such that up until the hardcoded blocks skip the parent block fetch and once the chain goes beyond the list of known transition points, switch to fetching version info from the parent block?

A question: is there a reason for allowing xts in the block that use `set_code`? Or would it be reasonable to ask for this tricky behaviour to change?

joepetrowski:
> Maybe you could combine the two approaches, such that up until the hardcoded blocks skip the parent block fetch and once the chain goes beyond the list of known transition points, switch to fetching version info from the parent block?

The problem with this is that Sidecar is stateless, so it doesn't really have the concept of knowing if it's past a certain block hash. I think matching on hash is a good approach. I also wonder if we can make the list accept type `int` and fetch the block number in the check (since it comes with `header`) and then back up and get the parent if that matches. Then we could add upcoming scheduler blocks for which we don't yet have the hash.

> A question: is there a reason for allowing xts in the block that use `set_code`? Or would it be reasonable to ask for this tricky behaviour to change?

If we want to make this generic for all Substrate based chains, I don't think we can make this assumption.

emostov:
> Maybe you could combine the two approaches, such that up until the hardcoded blocks skip the parent block fetch and once the chain goes beyond the list of known transition points, switch to fetching version info from the parent block?

Like @joepetrowski, I believe we would need block numbers. Currently we only have a block number when the user supplies it. Otherwise we either use `getFinalizedHead()` which returns a hash or we use the user supplied a hash. So in most logic branches we would need to query to get the block number and then cross check it against a list of ranges. Something that could help us optimize would be if `getFinalizedHead()` returned a number in addition to hash. I think some users continuously poll the `/block` endpoint which uses `getFinalizedHead()` - so potentially a worthwhile optimization. 

A further optimization would be if we had lists of ranges with hardcoded version info, saving us a query to `getRuntimeVersion` for known ranges. However I think this may be premature and add more complexity than its worth for the maintainers at this point.

> A question: is there a reason for allowing xts in the block that use `set_code`? Or would it be reasonable to ask for this tricky behaviour to change?

I am not sure what is "reasonable", but I imagine it might make things a little easier for ecosystem/tooling to have guarantees for all of substrate/frame based chains that no other extrinsics exists in `set_code` blocks.

README.md
```diff
@@ -51,6 +51,16 @@ Following ENV variables can be set:
...
+-   `SAS_SUBSTRATE_PARENT_VERSION`: always use the version of the parent block to set
+    metadata for scale decoding. `on` means always fetch and use the parent block's metadata, while `off`
+    means _only_ fetch and use the parent blocks metadata for the blocks with hashes specified in
+    `config/setCodeBlocks.ts`. Fetching and using the parent block's metadata will have a slight time
+    penalty, but will guarantee that extrinsics within a block that executed a runtime upgrade with
+    `set_code` are always decoded correctly. Defaults to `off`. Note: for chains that use the
+    `scheduler` pallet to do runtime upgrades, it is very unlikely this will be an issue as
+    `scheduler` ensures that `set_code` is executed in `on_initialize` and it is highly unlikely
+    there would be any additional room in a block for other extrinsics, but please use this info with
+    discretion.
```
danforbes, line 63: Might try to make this a bit more helpful/clear by updating to something like...

```suggestion
    `set_code` are always decoded correctly. Defaults to `off`. We recommend that chains that
    use scheduled upgrades use the default behavior, however, please consider your use case
    when making a final decision.
```

emostov:
I just deprecated `ensureMeta` due to https://github.com/polkadot-js/api/pull/2545 - so this PR is mute assuming we do not resurrect `ensureMeta`.

dvdplm:
@insipx do you mind taking a look at this too?

insipx:
> I just deprecated ensureMeta due to polkadot-js/api#2545 - so this PR is mute assuming we do not resurrect ensureMeta.
> @insipx do you mind taking a look at this too?

If i understand correctly, this PR is no longer needed because PolkadotJS does the metadata update itself if it detects a change in spec version and registers this change in it's own type registry?

I think that this method of hardcoding the blocks at which the transition occurs is the way to go for a stateless approach. The `getRuntimeVersion` call is slow by nature since it needs to fetch the WASM blob for that block and call into it in order to get the runtime version, and calling into WASM is expensive. There's not really a way around that unless there is a change in how substrate stores/gets the spec for each block. So then this translates into a performance penalty when calling runtime version for each block via the RPC.

Substrate-archive had the same problem, but it has the benefit of knowing each historical block, so it can do a binary search for runtime versions for batches of blocks thereby reducing calls into the runtime.

PR #239, opened 2020-08-27
Add OpenApi UI structure in docs folder for future use

emostov:
This PR adds an OpenApi UI that will be hosted by github pages. The plans for v1.0.0 release are reflected in `openapi-v1.yaml`, while the plans for routes to add in v1.1.0 are reflected in `openapi-v1-todo.yaml`. The UI is currently set to show the v1.0.0 paths, which are ~not yet~ fully implemented.

Relates to #227


emostov:
@danforbes You should be able to pull the branch down and test it that way. Its in `docs/dist/index.html`. You will need to run `yarn build:docs` if you make any changes. For now I am more looking for feedback on the setup of the UI components and not the exact api specs since some still might change given the open PRs.

Also you can access the version from this branch here: https://paritytech.github.io/substrate-api-sidecar/dist/

docs/src/openapi-v1-0.yaml
```diff
@@ -0,0 +1,1317 @@
...
+          format: unsignedInteger
+    ElectionStatus:
+      type: object
+      properties:
+        status:
+          type: object
+          description: >-
+            Era election status: either `Close: null` or `Open: <BlockNumber>`.
+            A status of `Close` indicates that the submission window for solutions
+            from off-chain Phragmen is not open. A status of `Open` indicates the
```
dvdplm, line 1060:
```suggestion
            from off-chain Phragmen is not open. A status of `Open` indicates that the
```
maybe?

docs/src/openapi-v1-0.yaml
```diff
@@ -0,0 +1,1317 @@
...
+          description: Node's chain name.
+      description: Version information of the node.
+    NodeRole:
+      type: string
+      description: Role the node is running.
+      enum:
+      - Full
+      - LightClient
+      - Authority
+      - Sentry
```
dvdplm, line 931: Sentry nodes are being deprecated – worth mentioning here?

dvdplm (commented):
Not all calls have error descriptions – does that mean they can't fail? In general this lgtm a part from a few nits.

docs/src/openapi-v1-0.yaml
joepetrowski: Yes!

docs/dist/oauth2-redirect.html
```diff
@@ -0,0 +1,68 @@
...
+        if ((
+          oauth2.auth.schema.get("flow") === "accessCode"||
+          oauth2.auth.schema.get("flow") === "authorizationCode"
+        ) && !oauth2.auth.code) {
+            if (!isValid) {
+                oauth2.errCb({
+                    authId: oauth2.auth.name,
+                    source: "auth",
+                    level: "warning",
+                    message: "Authorization may be unsafe, passed state was changed in server Passed state wasn't returned from auth server"
```
joepetrowski, line 40: Confusing

joepetrowski (approved):

dvdplm:
> In theory we only have to detail all _known_ errors. So essentially all 4xx errors should be detailed and some 5xx that are hard coded. I actually have not gone through the code and the specs though to make sure they align. I mostly just have errors from middleware that apply broadly. If we want to be thorough this definitely needs to be looked at carefully.

It's ok to defer this to a separate PR. Can you file a ticket?

docs/src/openapi-v1.yaml
```diff
@@ -1566,7 +942,7 @@ components:
       description: Version information of the node.
     NodeRole:
       type: string
-      description: Role the node is running.
+      description: Role the node is running. (Sentry nodes are being deprecated.)
```
dvdplm, line 945:
```suggestion
      description: Role of this node. (N.B. Sentry nodes are being deprecated.)
```

dvdplm (approved):
lgtm

PR #242, opened 2020-08-28
Implement `/runtime/metadata`

emostov:
This PR implements `/runtime/metadata` as per `openapi-proposal.yaml`. Additionally it updates the README.md to indicate which paths will be deprecated.

Relates to #227


joepetrowski (approved):
Somehow `/block` got into the README twice, just delete that and it's good to go.

Issue #257, opened 2020-09-03
Label event `data` parameters in `/blocks` endpoint


emostov:
Currently event data just comes across as an array of values. It would be more human readable if those values where labelled. This should be possible because the info is available in the runtime metadata as:
```                    
"events": [
                        {
                            "name": "ExtrinsicSuccess",
                            "args": [
                                "DispatchInfo"
                            ],
                            "documentation": [
                                " An extrinsic completed successfully. [info]"
                            ]
                        },
..]
```
It should be investigated if polkadot-js already has this info built in.

This should get in before v1.0.0 is released

PR #260, opened 2020-09-06
Add eventDocs & extrinsicDocs query params to /blocks

emostov:
Closes: #257 

This PR adds 2 optional query params for the `/blocks` endpoint: `eventDocs` & `extrinsicDocs`. They are both boolean parameters that allow the user to opt into having `docs` in the response. `eventDocs=true` will include `docs` for every event while `extrinsicDocs` will include `docs` for top level extrinsics int the `extrinsics` array/

See sample response from `/blocks/1470716?eventDocs=true&extrinsicDocs=true` here: https://gist.github.com/emostov/dafb0b017df13b64fb411f6e6d769469


src/controllers/blocks/BlocksController.ts
```diff
@@ -13,6 +13,15 @@ import AbstractController from '../AbstractController';
  * - (Optional) `number`: Block hash or height at which to query. If not provided, queries
  *   finalized head.
  *
+ * Query:
+ * - (Optional) `eventDocs`: When set to `true`, every event will have an extra
+ * 	`docs` property containing an array of strings that represent each line of
+ * 	events documentation
```
joepetrowski, line 19:
```suggestion
 * 	the event's documentation.
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -13,6 +13,15 @@ import AbstractController from '../AbstractController';
...
+ * - (Optional) `extrinsicDocs`: When set to `true`, every extrinsic will have an extra
+ * 	`docs` property containing an array of strings that represent each line of
+ * 	events documentation
```
joepetrowski, line 22:
```suggestion
 * 	the extrinsic's documentation.
```

joepetrowski (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -212,6 +229,7 @@ export class BlocksService extends AbstractService {
 				success: defaultSuccess,
 				// paysFee overrides to bool if `system.ExtrinsicSuccess|ExtrinsicFailed` event is present
 				paysFee: null as null | boolean,
+				docs: extrinsicDocs ? extrinsic.meta.documentation : undefined,
```
dvdplm, line 232: Elsewhere you have `const eventDocsArg = eventDocs === 'true';` which implies that this should be a `bool`, so I'd expect this to be `docs: extrinsicDocs ? extrinsic.meta.documentation : false,`. Or why do you prefer `undefined` here?

src/controllers/blocks/BlocksController.ts
```diff
@@ -13,6 +13,15 @@ import AbstractController from '../AbstractController';
  * - (Optional) `number`: Block hash or height at which to query. If not provided, queries
  *   finalized head.
  *
+ * Query:
+ * - (Optional) `eventDocs`: When set to `true`, every event will have an extra
+ * 	`docs` property containing an array of strings that represent each line of
+ * 	events documentation
+ * - (Optional) `extrinsicDocs`: When set to `true`, every extrinsic will have an extra
+ * 	`docs` property containing an array of strings that represent each line of
```
dvdplm, line 21: I think it's unfortunate that the docs come back as an array of strings rather than a single string (with newlines). Is that something that is easy to change?

src/controllers/blocks/BlocksController.ts
```diff
@@ -73,12 +82,18 @@ export default class BlocksController extends AbstractController<
...
 	 */
-	private getLatestBlock: RequestHandler = async (_req, res) => {
+	private getLatestBlock: RequestHandler = async (
+		{ query: { eventDocs, extrinsicDocs } },
+		res
+	) => {
 		const hash = await this.api.rpc.chain.getFinalizedHead();
 
+		const eventDocsArg = eventDocs === 'true';
+		const extrsinsicDocsArg = extrinsicDocs === 'true';
```
mattrutherford, line 92: small typo here maybe: `extrsinsicDocsArg`

src/controllers/blocks/BlocksController.ts
kianenigma: Yeah I guess we could do some parsing.. The array of string is the side effect of how the metadata is being generated.

src/services/blocks/BlocksService.ts
kianenigma: AFAIU This is setting the _final value_ itself inside a `{}`, thus the field `docs` will be set either to `extrinsic.meta.documentation` or `undefined`.

src/services/test-helpers/responses/blocks/blocks789629.json
```diff
@@ -52,11 +52,34 @@
...
             }
+          ],
+          "docs": [
+            " An extrinsic completed successfully. [info]"
           ]
         }
       ],
       "success": true,
-      "paysFee": true
+      "paysFee": true,
```
kianenigma, line 62: what's up with this file?

src/types/responses/FrameMethod.ts
```diff
@@ -0,0 +1,11 @@
...
+	pallet: string;
+	method: string;
+}
+
+export function isFrameMethod(thing: unknown): thing is IFrameMethod {
+	return (
+		typeof (thing as IFrameMethod).pallet === 'string' &&
+		typeof (thing as IFrameMethod).method === 'string'
+	);
+}
```
kianenigma, line 11: fascinating 🌈 :D

kianenigma (approved):
LGTM

src/services/blocks/BlocksService.ts
emostov: The reason I chose `undefined` here is that when a property (in this case `docs`) is set to undefined in a response body (in this case a plain-old-javascript-object), the response body will omit that field. If we wanted to keep the field in even if they did not want docs I think `null` would make the most sense.

I think have `docs` set to `false` when the user does not opt in to them might be confusing because the inverse would be having an array of strings for `docs`, not `true`.

src/controllers/blocks/BlocksController.ts
emostov: We certainly can do parsing, appending a new line char and then joining them with a string. I will make that change and update the sample response so we can look at it.

src/services/test-helpers/responses/blocks/blocks789629.json
emostov: Is used to test the response from the service. This way anytime there is a user contract change in the service it forces one to change this file and document the change. Additionally it helps prevent accidental regressions in the user contract.

emostov:
I have refactored and changed the response so that instead of an array of strings, the docs is not a single string with new line characters at the end of each line. Take a look at the updated test respons json and the gist to see what the response looks like.

Issue #253, opened 2020-09-01
Errors getting blocks on Kusama


joepetrowski:
Node: v0.8.23
Sidecar: cb0114c3ce8d723e76b10df50cff453d0f4ba85a

Lots of issues with some recent blocks:
```
GET /block/3860498 200 89.430 ms - 2542
GET /block/3860499 200 105.013 ms - 1403
GET /block/3860500 200 140.271 ms - 1267
GET /block/3860501 200 123.557 ms - 1403
Unable to decode storage system.events: createType(Vec<EventRecord>):: Vec length 640123136 exceeds 32768
2020-09-01 20:33:34        RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: createType(Vec<EventRecord>):: Vec length 640123136 exceeds 32768
GET /block/3860502 200 94.861 ms - 2023
GET /block/3860503 200 83.428 ms - 2563
GET /block/3860504 200 90.262 ms - 1403
GET /block/3860505 200 84.059 ms - 1075

GET /block/3860965 200 120.164 ms - 1075
GET /block/3860966 200 107.344 ms - 1403
GET /block/3860967 200 119.523 ms - 1075
GET /block/3860968 200 102.427 ms - 1403
Unable to decode storage system.events: createType(Vec<EventRecord>):: Vec length 1020887296 exceeds 32768
2020-09-01 20:34:21        RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: createType(Vec<EventRecord>):: Vec length 1020887296 exceeds 32768
GET /block/3860969 200 82.530 ms - 1889
GET /block/3860970 200 81.501 ms - 1075
GET /block/3860971 200 84.792 ms - 1595
GET /block/3860972 200 108.438 ms - 1267
GET /block/3860973 200 112.822 ms - 1403

GET /block/3861060 200 108.970 ms - 1075
GET /block/3861061 200 108.838 ms - 1595
GET /block/3861062 200 96.271 ms - 1595
GET /block/3861063 200 37.912 ms - 1075
Unable to decode storage system.events: createType(Vec<EventRecord>):: Vec length 842006784 exceeds 32768
2020-09-01 20:34:30        RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: createType(Vec<EventRecord>):: Vec length 842006784 exceeds 32768
GET /block/3861064 200 75.661 ms - 2981
[...]

PR #270, opened 2020-09-09
Configurable write logs to file

emostov:
# issue

The Sidecar team has been getting occasional reports of [`createType` issues](https://github.com/paritytech/substrate-api-sidecar/issues/253) from users. However, these errors are hard to reproduce and insufficient logging has been produced for them. An initial effort was made to create a simple CLI tool (see [here](https://github.com/danforbes/polkadot-js-scripts) & [here](https://github.com/emostov/polkadot-js-scripts) that would be used to capture the bare bones polkadot-js RPC calls that have been failing. The idea with logging RPC calls and responses is that we can look at the scale encoded values in response that failed on `createType` and check to see if the data is corrupt. Despite using the tools ourselves and some of Sidecars's users trying the tools on their setups we have not been able to log a `createType` error. With this lack of being able to reproduce the issue on demand, the importance of more robust, in production, logging functionality rises.

# presented solution

In this PR I introduce winston logging, which gives us rich configurability and a [wide host of transport options](https://github.com/winstonjs/winston/blob/master/docs/transports.md#winston-transports). Materially for the user, this modifies the console output to include time stamps and log levels, while also allowing logging to file. I introduce a host of configuration options for log levels and various specifics of file logging (see `README.md` additions for details).

All file logging is in a structured, JSON format that at minimum includes a timestamp, message, and level field.

In the process I have removed the morgan logger and replaced it with express-winston logger. The format for how requests are logged is now slightly changed in the console and and a large amount of info is now dumped to file for each request. I could not get the status code numbers color coded like before (not _important_ but a nice touch), however the log level is still color coded.

Our configuration manager only allows one to set their `.env` file with `NODE_ENV`, which is a name space collision with what we need for @polkadot-js/api RPC logging (`NODE_ENV=test`). The solution chevdor presented is to symlink your desired `.env` file with `.env.test`, I detail this process in the README

## logging RPC

The logging we care most about comes from @polkadot-js/api. In order to enable the logging one must set `NODE_ENV=test` and package will `console.log` too `stdout`.

 As far as I can tell we cannot put our own logging utilities into a dependency like @polkadot-js/api. So the only option I could think of was monkey patching `console.{log, error, warn}` with a function that wraps the equivalent winston logger level (see: src/logging/consoleOverride.ts). In the monkey path I had to change how winston formats inputs so it would behave like node.js's console.

The script `yarn start:log-rpc` was added to the `package.json` and it sets `NODE_ENV=test` when running `main.js`, so RPC logs will be output from @polkadot-js/api.

---

### other logging/config options we could consider:
- Look through options for file transport: https://github.com/winstonjs/winston/blob/master/docs/transports.md#file-transport
- Option to always write to same file (maybe helpful for those trying to redirect output?
- Rotating files option? David mentioned sys admins might want to do this themselves, but could perhaps could still be helpful for some?

### points for feedback
- Are logging options clear?
- Does the file logging setup allow afford our users the correct amount and type of control?
- Should we try to log any more or less info (I see the most place for change in what we log to file for HTTP requests)?


danforbes (approved):

chevdor:
I think that writing the logs yourself is re-inventing the wheel heavily. See below.

Regarding the `.env` vs `.env.test` issue, my recommendation is to set `NODE_ENV` to what you need: `NODE_ENV=test` and if you want to use `.env` instead of `.env.test, you can create a symlink: `ln -s .env .env.test`.

To solve your logging problem in a clean way, I would suggest using a logger such as winston where you can log things without having to care about WHERE it lands and then tell the logger to write files, or write to a DB, to a log manager, etc....

If you need a sample, you can check out:
- definition: https://gitlab.com/Polkabot/polkabot/-/blob/master/packages/polkabot-api/src/LoggerFactory.ts
- instantiation: https://gitlab.com/Polkabot/polkabot/-/blob/master/packages/polkabot/src/index.ts#L17
- usage: https://gitlab.com/Polkabot/polkabot/-/blob/master/packages/polkabot/src/index.ts#L81

dvdplm:
I second @chevdor here: I'd prefer an off-the-shelf logger. Not familiar with the node ecosystem though so I have no opinion of `winston`. Good loggers take care of nasty stuff like working with log rotation, crashes and platform dependent details and should be pretty easy to use.

emostov:
@chevdor, @dvdplm - all good points that I agree with. The thing that stopped me from that originally using winston is that we need logging from a dependency and I can't put a third party logging statement directly in the dependencies code. But thinking about it I could wrap the logger in a function with the same signature as console.log and feed all the console statements through that - so I'll give that a try unless there is a better way.

specs.yml
```diff
@@ -1,10 +1,39 @@
...
+  LOG:
+    CONSOLE_LEVEL:
+      description: Log level for console transport
+      default: info
+      regexp: ^error|warn|info|http|verbose|debug|silly$,
+
+    FILE_USE:
+      description: Wether or not to use file transport
+      default: yes
+      type: string
```
chevdor, line 11: This is probably where you should use a `boolean` instead of a string.
Cherry on top, `confmgr` will handle true/false as well as True/False, 0/1 and also yes/no.
Reference: https://gitlab.com/chevdor/confmgr/-/blob/master/src/ConfigManager.ts#L117

chevdor (changes requested):
I suggest switching from `string` to `boolean`. See details below.

src/logging/consoleOverride.ts
```diff
@@ -0,0 +1,51 @@
...
+		//
+		// `(format.apply(format, args as [string]) as unknown) as object: TS incorrectly says the we
+		// need a object as an argument to `info.call`. However, it will except a string perfectly fine,
+		// which is what `format.apply` returns.
+		logger.info.call(
+			logger,
+			(format.apply(format, args as [string]) as unknown) as object
+		);
+	};
+	console.info = function (...args: unknown[]) {
```
chevdor, line 27: This function has a lot of duplication.
Basically, what you are doing is to say: "for function A, B and C, please use this implementation".

 You can make it shorter and more concise using:
```
['info', 'error', 'warn', 'log'].map (level => {
     console[level] = function (...);
}
```

emostov:
See updated top comment that reflects recent changes.

Here is a snippet of what logging looks like in the console: 
Screen Shot 2020-09-15 at 12 28 42 AM

For now I chose to put a stack trace dump in the console whenever available since it can be helpful for tracking down the source of an issue, especially when users are asking for help debugging. It is less effort than asking someone to look in a file. However, I also think its reasonable to not include the stack trace by default in the console.

src/logging/transformers/stripAnsi.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Strip ANSI characters from strings in simple, arbitrary data. N.B. this is not
+ * hardened to work with any possible javascript type and is only meant for arrays,
+ * basic objects and strings.
+ *
+ * @param data arbitrary data
+ */
+function sanitizeAnsi(data: unknown): unknown {
```
dvdplm, line 33: I think `sanitize` is the wrong verb here: it's usually used to guard against security hazards. We're really stripping out data we don't want here, but not for security reasons. Can it be named `stripAnsiShellCodes` or some such?

It is called recursively (unbounded). I don't think that is a big concern, but maybe worth pondering for a second if users have any way to submit a malicious payload that would crash the server?

src/logging/transformers/stripAnsi.ts
```diff
@@ -0,0 +1,63 @@
...
+const ansiRegex = new RegExp(pattern, 'g');
+
+/**
+ * Remove ANSI characters from a string by replacing them with empty an empty string.
+ *
+ * @param text
+ */
+function ansiReplace(text: unknown): typeof text {
+	return typeof text === 'string' ? text.replace(ansiRegex, '') : text;
+}
```
dvdplm, line 24: Can't we assume `text` is a string here? The docs seem to do so already and the call site below also checks the type before calling. 

My TS is weak, but is the return value right here? Should it really be `typeof text`? Does that mean "the return type is whatever `text` was"?

src/logging/transformers/nodeUtilFormat.ts
```diff
@@ -0,0 +1,19 @@
+import { TransformableInfo } from 'logform';
+import { SPLAT } from 'triple-beam';
+import util from 'util';
+import { format } from 'winston';
+
+/**
+ * Console.log style formatting using node's `util.format`. We need this so we
+ * can override console.{log, error, etc.} without issue.
+ */
+export const nodeUtilFormat = format(
```
dvdplm, line 10: I don't understand what this does, can you ELI5? What does `SPLAT` do?

src/logging/consoleOverride.ts
```diff
@@ -0,0 +1,36 @@
...
+	[
+		['log', 'info'],
+		['info', 'info'],
+		['warn', 'warn'],
+		['error', 'error'],
+		['debug', 'debug'],
+	].forEach(([consoleLevel, winstonLevel]) => {
+		// Sacrereligious typecasting explained:
+		//
+		// `args as [string]`: fomat @types dictate that it needs an array of at least length 1. However,
```
dvdplm, line 22:
```suggestion
		// `args as [string]`: format @types dictate that it needs an array of at least length 1. However,
```

src/logging/consoleOverride.ts
```diff
@@ -0,0 +1,36 @@
...
+		// from testing this is not neccesary, so we override the type as a string tuple.
+		//
+		// `(format.apply(format, args as [string]) as unknown) as object: TS incorrectly says the we
```
dvdplm, line 25:
```suggestion
		// `(format.apply(format, args as [string]) as unknown) as object`: TS incorrectly says the we
```

src/logging/consoleOverride.ts
```diff
@@ -0,0 +1,36 @@
...
+		// need a object as an argument to `info.call`. However, it will except a string perfectly fine,
```
dvdplm, line 26:
```suggestion
		// need a object as an argument to `info.call`. However, it will accept a string perfectly fine,
```

specs.yml
```diff
@@ -1,10 +1,45 @@
...
+      type: number
+      regexp: "[0-9]+"
+
+    FILE_PATH:
+      description: Path to log file
+      default: ./logs/file-transport.log
+      type: string
+
+    FILE_STRIP_ANSI:
+      description: Wether or not to strip ANSI characters when writing to file transport
```
dvdplm, line 38:
```suggestion
      description: Whether or not to strip ANSI characters when writing to file transport
```

specs.yml
```diff
@@ -1,10 +1,45 @@
...
-      regexp: ^error|all|none$
+  LOG:
+
+    CONSOLE_LEVEL:
+      description: Log level for console transport
+      default: info
+      regexp: ^error|warn|info|http|verbose|debug|silly$,
+
+    FILE_USE:
+      description: Wether or not to use file transport
```
dvdplm, line 10:
```suggestion
      description: Whether or not to use file transport
```

src/middleware/logger/httpLogger.ts
```diff
@@ -0,0 +1,22 @@
...
+export const httpLoggerCreate = (winstonInstance: winston.Logger): e.Handler =>
+	expressWinston.logger({
+		// transports,
+		winstonInstance,
+		msg: `{{req.method}} {{req.url}} {{req.res.statusCode}} {{res.responseTime}}ms`,
+		level: (_req, res) => {
+			const { statusCode } = res;
+			if (statusCode < 400) {
+				// `http` is above `info`, so one needs to opt into `http` or above log level to view sub 400
+				return 'http';
```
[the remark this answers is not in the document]
emostov: See npm log levels: https://www.npmjs.com/package/winston#logging-levels

I don't thinks its particularly meaningful but it is the default

src/logging/transports/fileTransport.ts
```diff
@@ -0,0 +1,37 @@
...
+export function fileTransport() {
+	const transformers = [
+		stripTimestamp(),
+		nodeUtilFormat(),
+		format.timestamp({ format: 'YYYY-MM-DD HH:mm:ss.SSS' }),
+		format.prettyPrint(),
+	];
+
+	if (config.FILE_STRIP_ANSI) {
+		console.info('STRIP');
```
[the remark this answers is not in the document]
emostov: forgot to push my last clean up commit 🤦

src/logging/transports/fileTransport.ts
```diff
@@ -0,0 +1,37 @@
...
+		transformers.unshift(stripAnsi());
+	} else {
+		console.debug('NO STRIP');
+	}
+
+	return new transports.File({
+		level: config.FILE_LEVEL || 'http',
+		filename: config.FILE_PATH || `./logs/file-transport.log`,
+		maxsize: config.FILE_SIZE || 524_288_000, // 500MB
+		maxFiles: config.FILE_COUNT || 2, // Max 1gb
```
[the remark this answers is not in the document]
emostov: It does not stop logging, it will just delete the oldest file. Should we try to do something else?

src/logging/transports/consoleTransport.ts
```diff
@@ -0,0 +1,38 @@
...
+export function consoleTransport() {
+	/**
+	 * A simple printing format for how `TransformableInfo` shows up.
+	 */
+	const simplePrint = format.printf((info: TransformableInfo) =>
+		info?.stack // If there is a stack dump (e.g. error middleware), show that in console
+			? `${info?.timestamp as string} ${info?.level}: ${
+					info?.message
+			  } \n ${info?.stack as string}`
+			: `${info?.timestamp as string} ${info?.level}: ${info?.message}`
```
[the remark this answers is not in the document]
emostov: Changed ... just wish we had Rust if else statements with implicit returns in TS

src/logging/transformers/stripAnsi.ts
emostov: You are correct in the use of `typeof`. 

I just removed this function since we where not reusing the logic anywhere else, and like you say we do a type check right before calling it.



I think when I initially added that function I thought I would use it in another way where it could be helpful to do runtime checks of the type to make sure it doesn't throw.

emostov:
@dvdplm - I updated the config options for console to take into consideration those who use `stdout` to collect logging info (console transport goes to `stdout`). Changes reflected in the README.

Let me know when you think its ready to merge.

cc @joepetrowski

src/middleware/logger/httpLogger.ts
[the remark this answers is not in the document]
dvdplm: TIL.

src/logging/transports/fileTransport.ts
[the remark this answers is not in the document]
dvdplm: It deletes the oldest and creates a new file? I'd prefer we don't try to manage log file size at all and just let `logrotate` (or whatever the node operator prefers) to do its thing. Having it here sure is convenient but I'm afraid it'll lead to issues down the line when `sidecar` tries to delete a file while some other process is busy reading/rotating it.

I think it's fine for now and no need to block this PR, but definitely something worth pointing out in the docs.

dvdplm (approved):

src/Config.ts
```diff
@@ -8,9 +8,17 @@ import * as configTypes from '../config/types.json';
 export interface ISidecarConfig {
 	HOST: string;
 	PORT: number;
-	LOG_MODE: string;
 	WS_URL: string;
 	CUSTOM_TYPES: Record<string, string> | undefined;
+	CONSOLE_LEVEL: string;
```
chevdor, line 13: Regarding `confmgr`: your App is split into a few modules.
I think it could be helpful to keep this structure here instead of flattening it.
That also allows passing a whole 'LoggerConfig' Object to the logger related code.

So in short (pseudo code):
```
export SideCarConfigExpress {
  HOST: string;
  PORT: number;
}

export SideCarConfigLogs {
  ...
}


export interface ISidecarConfig {
    EXPRESS: SideCarConfigExpress;
    LOGS: SideCarConfigLogs;
    ....
}
```

src/logging/transports/fileTransport.ts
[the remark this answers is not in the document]
emostov: If we take out `maxSize` and `maxFile` then it will just keep writing to the same file. That would be a super quick change. We could also have yet another config option that just says "Only write to one file". Either way is ok for me. 
Honestly the main reason I had maxFile and maxSize is so some user who isn't really aware of the logs has random a random issue where Sidecar is sucking up disk space.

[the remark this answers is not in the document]
dvdplm: > Honestly the main reason I had maxFile and maxSize is so some user who isn't really aware of the logs has random a random issue where Sidecar is sucking up disk space.

The default is to send logs to stdout only yes? If they opt in to log to file but do not read the docs and do not set up log rotation properly... well, that's on them.

PR #274, opened 2020-09-09
Remove irrelevant fields for unsigned extrinsics

### LINKED ISSUE #215: Set `tip` and `nonce` to null for unsigned transactions and inherents in `/block` response
Currently, inherents and unsigned transactions have a `tip` and `nonce` field despite the information not being relevant to there validation or execution. These fields should thus be set to null.

Additionally, it is worth investigating if `paysFee` should always be false for inherents. Currently inherents always have pays fee as true because the `system.ExtrinsicSuccess` seems to default to having it as a true value despite inherents not having any users and not paying any fees.

emostov:
Closes #215 

This PR affects anything in the extrinsics array of the `/blocks` endpoint that does not have a signature.

I make the assumption that anything that does not have a signature:
- does not pay fees (`paysFee` is always `false`)
- does not have a tip (`tip` field is emitted)
- does not have a nonce (`nonce` field is emitted)

As far as I know though there is not a concrete way to distinguish between an unsigned transaction and an inherent - their external interface is the same and the differences come from how they are processed.

### DIFF
--- a/src/services/blocks/BlocksService.spec.ts
+++ b/src/services/blocks/BlocksService.spec.ts
@@ -39,7 +39,7 @@ describe('BlocksService', () => {
 				sanitizeNumbers(
 					await blocksService.fetchBlock(blockHash789629, true, true)
 				)
-			).toStrictEqual(blocks789629Response);
+			).toMatchObject(blocks789629Response);
 		});
 
 		it('throws when an extrinsic is undefined', async () => {
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -220,15 +220,15 @@ export class BlocksService extends AbstractService {
 					method: method.methodName,
 				},
 				signature: isSigned ? { signature, signer } : null,
-				nonce,
+				nonce: isSigned ? nonce : undefined,
 				args: this.parseGenericCall(method).args,
-				tip,
+				tip: isSigned ? tip : undefined,
 				hash,
 				info: {},
 				events: [] as ISanitizedEvent[],
 				success: defaultSuccess,
 				// paysFee overrides to bool if `system.ExtrinsicSuccess|ExtrinsicFailed` event is present
-				paysFee: null as null | boolean,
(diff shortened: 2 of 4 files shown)


joepetrowski:
I think I prefer `null` rather than just removing those field entirely. Imagine you have an sr25519 session key for `im-online` that submits heartbeats as unsigned transactions. Now, the way people are _supposed_ to use these keys is that they are only for the session and not to control funds. But someone could make the bad decision to get the address for that key and store some funds on it. Bad idea, but I can't really stop them.

So this account could actually have a nonce, it's just that for the _unsigned transaction,_ it doesn't really make sense. We want to express, "hey, this nonce thing just doesn't make sense for this type of transaction." AFAIK, `null` is a good way to do that, but really open to other opinions on it.

emostov:
> I think I prefer `null` rather than just removing those field entirely. Imagine you have an sr25519 session key for `im-online` that submits heartbeats as unsigned transactions. Now, the way people are _supposed_ to use these keys is that they are only for the session and not to control funds. But someone could make the bad decision to get the address for that key and store some funds on it. Bad idea, but I can't really stop them.
> 
> So this account could actually have a nonce, it's just that for the _unsigned transaction,_ it doesn't really make sense. We want to express, "hey, this nonce thing just doesn't make sense for this type of transaction." AFAIK, `null` is a good way to do that, but really open to other opinions on it.

@joepetrowski I just updated to so the fields are now null if not signed. However in the `im-online` edge case you mentioned it is unclear to me if you think we should show the nonce? I'm not familiar with how `im-online` works, but rn in this PR anything that `isSigned` will have a nonce etc.

Also would be good to get an opinion from someone like @shawntabrizi on how to best surface this info.

joepetrowski:
I mean I think we should show `nonce: null`. It says, "the key that signed this message might be an account, might not be, but either way a nonce is not relevant to this extrinsic."

emostov:
@joepetrowski ready to merge?

joepetrowski (approved):

PR #278, opened 2020-09-11
Implement `/pallets/{palletId}/storage/{storageItemId}`

### LINKED ISSUE #277: Implement `/pallets/{storageItemId}/storage/{storageItemId}`
**Proposed Change or Idea**
Implement `/pallets/{storageItemId}/storage/{storageItemId}`

**One use case: Getting the block number when referenda will enact**

democracy.ReferendumInfoOf 

1) Monitor chain for democracy.Started events and log index.
2) Get ReferendumInfoOf(index) to find ending block number (end).
3) Execution block number will be end + delay.
4) Monitor chain for democracy.Passed, democracy.NotPassed, or democracy.Cancelled events citing the index.
    - If Passed, need to look at scheduler.Scheduled event in same block for enactment block.
5) Monitor chain for democracy.PreimageNoted events with the same hash as the ReferendumInfoOf(index) item, may be up to the last block before execution, but will not work if this is missing.
6) Monitor chain for democracy.Executed events for actual execution.

Future related work:
Add a tool to decode preimages to say what the referendum is, but that's more advanced, for a start we can just say that a referendum will execute

emostov:
Closes #277 

This PR introduces a dynamic way to query pallet storage items (as declared in `decl_storage`).

Major points for review:
- General user contract
- Do we want pallet index/ storage item index in response?
- Do we want the `pallet` and `storageItem` field to show up as they do in substrate src (current impl), or should we have them show up as they would in polkadot-js?
- Should we require the user to opt into metadata / should we have the option at all? We could just pull out the docs. Or just wait until later to add this feature.

Follow up PRs:
- When implementing new pallets endpoints we may want create an abstract pallet class with some common methods.
- Instantiate controllers directly in App controller array
- validation middleware for identifiers
---
Sample response for `/pallets/democracy/storage/referendumInfoOf?key1=2&metadata=true`
```
    "at": {
        "hash": "0xb765cefda9230375ec33d3758a429db9a8d71eb431536d2a4f06b40518c3ac08",
        "height": "1540143"
    },
    "pallet": "Democracy",
    "storageItem": "ReferendumInfoOf",
    "key1": "2",
    "value": {
        "Ongoing": {
            "end": "1612800",
            "proposalHash": "0x7de70fc8be782076d0b5772be77153d172a5381c72dd56d3385e25f62abf507e",
            "threshold": "Supermajorityapproval",
            "delay": "403200",
            "tally": {
                "ayes": "41906823320400000",
                "nays": "214535586500000",
                "turnout": "34301429248000000"
            }
        }
    },
    "metadata": {
        "name": "ReferendumInfoOf",
        "modifier": "Optional",
        "type": {
            "Map": {
                "hasher": "Twox64Concat",
                "key": "ReferendumIndex",
                "value": "ReferendumInfo",
                "linked": false
            }
        },
        "fallback": "0x00",
        "documentation": [
            " Information concerning any given referendum.",
            "",
            " TWOX-NOTE: SAFE as indexes are not under an attacker’s control."
        ]
    }
}
```


docs/src/openapi-v1.yaml
```diff
@@ -1365,6 +1434,47 @@ components:
...
+          type: string
+        type:
+          $ref: '#/components/schemas/PalletStorageType'
+        fallback:
+          type: string
+        documentation:
+          type: array
+          items:
+            type: string
+      description: Metadata of a storage item from a FRAME pallet.
```
dvdplm, line 1455: This is a bit vague; an example of a return value would work well here imo.

src/controllers/pallets/PalletsStorageItemController.ts
```diff
@@ -0,0 +1,63 @@
...
+			api,
+			'/pallets/:palletId/storage',
+			new PalletsStorageItemService(api)
+		);
+
+		this.initRoutes();
+	}
+
+	protected initRoutes(): void {
+		// TODO in future PR, look into middleware validation of storageId and palletId
```
dvdplm, line 31: Open an issue and add reference to it here?

src/main.ts
```diff
@@ -142,6 +142,7 @@ async function main() {
 			transactionFeeEstimateController,
 			transactionSubmitController,
 			palletsStakingProgressController,
+			new controllers.palletsStorageItem(api), // TODO: Future PR: change remaining controllers to this style in order to reduce redundancy.
```
dvdplm, line 145: Same: add issue and reference it here.

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -14,7 +14,8 @@ export class AccountsBalanceInfoService extends AbstractService {
 		hash: BlockHash,
 		address: string
 	): Promise<IAccountBalanceInfo> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm:
```suggestion
```

src/services/accounts/AccountsStakingInfoService.ts
```diff
@@ -15,7 +15,8 @@ export class AccountsStakingInfoService extends AbstractService {
 		hash: BlockHash,
 		stash: string
 	): Promise<IAccountStakingInfo> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm:
```suggestion
```

src/services/accounts/AccountsVestingInfoService.ts
```diff
@@ -14,7 +14,8 @@ export class AccountsVestingInfoService extends AbstractService {
 		hash: BlockHash,
 		address: string
 	): Promise<IAccountVestingInfo> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm:
```suggestion
```

src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -15,7 +15,8 @@ export class PalletsStakingProgressService extends AbstractService {
 	async derivePalletStakingProgress(
 		hash: BlockHash
 	): Promise<IPalletStakingProgress> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm:
```suggestion
```

src/services/pallets/PalletsStorageItemService.ts
```diff
@@ -0,0 +1,118 @@
...
+		hash: BlockHash;
+		palletId: string;
+		storageItemId: string;
+		key1: string | undefined;
+		key2: string | undefined;
+		metadata: boolean;
+	}): Promise<IPalletStorageItem> {
+		if (!this.api.query[palletId]) {
+			throw new BadRequest(
+				`Invalid palletId: "${palletId}". Currently only pallet names are expected as a palletId.`
```
dvdplm, line 29: The error message sounds like there are potentially other ways of providing the `palletId`, what would those other ways be?
Also isn't it possible that the `palletId` is technically valid but there is no such pallet?

src/services/pallets/PalletsStorageItemService.ts
```diff
@@ -0,0 +1,118 @@
...
+			);
+		}
+
+		if (!this.api.query[palletId][storageItemId]) {
+			throw new BadRequest(
+				`Invalid storageItemId: "${storageItemId}". Currently only camel case storage item names are expected as storageItemId.`
```
dvdplm, line 35:
```suggestion
				`Invalid storageItemId: "${storageItemId}". Storage item names are expected to be in camel case, e.g. 'storageItemId'.`
```

src/services/transaction/TransactionFeeEstimateService.ts
```diff
@@ -15,7 +15,8 @@ export class TransactionFeeEstimateService extends AbstractService {
 		hash: BlockHash,
 		transaction: string
 	): Promise<RuntimeDispatchInfo> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm: remove

src/services/transaction/TransactionMaterialService.ts
```diff
@@ -13,7 +13,8 @@ export class TransactionMaterialService extends AbstractService {
 		hash: BlockHash,
 		noMeta: boolean
 	): Promise<ITransactionMaterial> {
-		const api = await this.ensureMeta(hash);
+		// const api = await this.ensureMeta(hash);
```
dvdplm: remove

dvdplm (changes requested):

dvdplm:
>     * Should we require the user to opt into metadata

I think this is the best option, but I am also unsure of what the argument is for the other options.

src/services/pallets/PalletsStorageItemService.ts
emostov: > The error message sounds like there are potentially other ways of providing the palletId, what would those other ways be?

We are planning on supporting look ups by index for `pallets/id/events`, `pallets/id/errors`, and `pallets/id/calls`. I was thinking at some point in the future we might use the same for storage items. But thinking about it I am actually  not sure that makes sense / is a paradigm used in substrate.

> Also isn't it possible that the palletId is technically valid but there is no such pallet?
Good point. I'll try and think of some better phrasing

src/services/transaction/TransactionMaterialService.ts
emostov: Not sure why github showed this diff to you. This PR fixed these: https://github.com/paritytech/substrate-api-sidecar/pull/279/files, and all that got merged into master and then here prior to your review.

src/services/pallets/PalletsStorageItemService.ts
emostov: Updated the error messages. Let me know what you think.

dvdplm:
> * Do we want pallet index/ storage item index in response?

I think not, but not sure I understand what this means. Unless it has explicitly been requested I think it's preferable to leave features out. Is possible to query the backend using index instead of `palletId`? If not, what is the use-case for the index?

joepetrowski:
> > * Do we want pallet index/ storage item index in response?
> 
> I think not, but not sure I understand what this means. Unless it has explicitly been requested I think it's preferable to leave features out. Is possible to query the backend using index instead of `palletId`? If not, what is the use-case for the index?

Probably not much use for it in storage item, but definitely useful in `pallets/{palletId}/errors` because dispatchables that fail return an index of the era. I would actually say that `palletId` should be accessible by index of the pallet in the runtime and _errors_ should be accessible by their index as well. So if you get `module: 6, error: 3` you should be able to go to `pallets/6/errors/3` and get the error. 

For storage item I don't think it's really useful or needed though.

emostov:
> Is possible to query the backend using index instead of palletId? If not, what is the use-case for the index?

I think errors, calls, and events, are sometime identified by `[palletIndex, {error, call, event}Index]`, but I have never seen that for a storage item before. For example, (if I remember correctly) I have seen substrate error messages that simply have the indexes of the pallet and the error - in this case its useful to uses index for lookup. Polkadot-js even has `findMetaCall`, `findMetaError`, and `findMetaEvent`, that each take a tuple of indexes and do a lookup in the metadata. See here: https://github.com/polkadot-js/api/blob/master/packages/types/src/create/registry.ts#L164

And as I was writing just saw @joepetrowski 's comment. I agree not useful for storage. I also don't know how we would even filter pallets. For others, like callIndex, you don't count the pallets that don't have calls.

emostov:
>> * Should we require the user to opt into metadata

>I think this is the best option, but I am also unsure of what the argument is for the other options.

I also think it is the best option because it results into us defaulting to sending a smaller payload and I imagine most use cases don't need the metadata.

emostov:
response for `/pallets/DEMOCRACY/storage/ReferendumInfoOf?key1=2&metadata=true`:
```
{
    "at": {
        "hash": "0x768af35e8ab50ce56d7fcee2a779de36dba8a44da0383f2b5497b44165305cc6",
        "height": "1551412"
    },
    "pallet": "democracy",
    "storageItem": "referendumInfoOf",
    "key1": "2",
    "value": {
        "Ongoing": {
            "end": "1612800",
            "proposalHash": "0x7de70fc8be782076d0b5772be77153d172a5381c72dd56d3385e25f62abf507e",
            "threshold": "Supermajorityapproval",
            "delay": "403200",
            "tally": {
                "ayes": "41925212461400000",
                "nays": "214535586500000",
                "turnout": "34485320658000000"
            }
        }
    },
    "metadata": {
        "name": "ReferendumInfoOf",
        "modifier": "Optional",
        "type": {
            "Map": {
                "hasher": "Twox64Concat",
                "key": "ReferendumIndex",
                "value": "ReferendumInfo",
                "linked": false
            }
        },
        "fallback": "0x00",
        "documentation": " Information concerning any given referendum.\n\n TWOX-NOTE: SAFE as indexes are not under an attacker’s control."
    }
}

**Comment by dvdplm** [comment]
So to sum it up:

> Do we want pallet index/ storage item index in response?

Pallet index: yes (and thank you for the explanation, makes sense)
Storage index: no (let's wait for someone to ask for it)

> Do we want the pallet and storageItem field to show up as they do in substrate src (current impl), or should we have them > show up as they would in polkadot-js?

polkadot.js, i.e. Javascript standard item names.

> Should we require the user to opt into metadata / should we have the option at all? We could just pull out the docs. Or just wait until later to add this feature.

Metadata: included if the user opts in.

File: src/services/pallets/PalletsStorageItemService.ts [hunk 12]
```diff
@@ -0,0 +1,145 @@
...
+				// extend the latest. So since the intro of ModuleMetadataV12 all versions have `index` in
+				// polkadot-js, but at the substrate level, only versions >= 12 have pallet `index`.
+				// https://github.com/polkadot-js/api/pull/2599
+				// https://github.com/paritytech/substrate/pull/6969
+				// https://github.com/polkadot-js/api/issues/2596
+				palletIdx = section.index.eqn(255)
+					? _sectionIdx
+					: section.index.toNumber();
+				palletMeta = section;
+				break;
```
emostov, line 133: Just wanted to note the changes here coming with polkadot-js `ModuleMetadataV12`

See:
https://github.com/polkadot-js/api/issues/2596
https://github.com/paritytech/substrate/pull/6969
https://github.com/polkadot-js/api/pull/2599

emostov:
Updated response with palletIndex:
```
{
    "at": {
        "hash": "0x6cce94d33193e1bbf51fe8c2a561400fbbf51c9638f49d131283432150f59dac",
        "height": "1597300"
    },
    "pallet": "democracy",
    "palletIndex": "15",
    "storageItem": "referendumInfoOf",
    "key2": "1",
    "value": {
        "Finished": {
            "approved": true,
            "end": "1003528"
        }
    },
    "metadata": {
        "name": "ReferendumInfoOf",
        "modifier": "Optional",
        "type": {
            "Map": {
                "hasher": "Twox64Concat",
                "key": "ReferendumIndex",
                "value": "ReferendumInfo",
                "linked": false
            }
        },
        "fallback": "0x00",
        "documentation": " Information concerning any given referendum.\n\n TWOX-NOTE: SAFE as indexes are not under an attacker’s control."
    }
}

dvdplm (approved):

PR #319, opened 2020-10-15
feat: graceful shutdown server

islishude:


src/main.ts
```diff
@@ -18,6 +18,7 @@
 import { ApiPromise } from '@polkadot/api';
 import { WsProvider } from '@polkadot/rpc-provider';
 import { json } from 'express';
+import * as shutdown from 'http-graceful-shutdown';
```
dvdplm, line 21:
```suggestion
import gracefulShutdown from 'http-graceful-shutdown';
```

dvdplm (changes requested):
The description is lacking; I'm not sure I understand what problem is solved by this PR? Can you please provide a motivation for why you need this?

src/main.ts
islishude: the package has no default export,so here can't use default import

islishude: the compiled result

屏幕快照 2020-10-15 下午6 14 16

islishude:
close,there's not much need for this

PR #322, opened 2020-10-20
fix: Catch failed opaque call decodes

emostov:
Calls as arguments in the `/blocks` endpoints are recursively parsed and decoded in order to give consistent responses. In the case of `as_multi`, we also parse [OpaqueCall](https://github.com/paritytech/substrate/blob/b49b666043c61891faa3b56ce7fc4d74c40ebad8/frame/multisig/src/lib.rs#L66) (the scale encoded call) to a polkadot-js `Call`. With [block 871651](https://polkascan.io/polkadot/transaction/0xf5f743643c3d2f88d976087beb630ec1794697aef684e70d389e6dd5d2f6adc8), the decoding of a giant `utility.batch` fails when trying to decode the `OpaqueCall` to a `Call` and the URI returns a 500. 

This PR introduces a fix by wrapping the conversion in try/catch block and return the `OpaqueBytes` as is in the event the conversion fails.

Ideally the conversion should not be failing and the exact cause of polkadot-js decode failure should be investigated thoroughly. So this is a temporary fix and perhaps does not need to be merged if a solution is found soon.


src/services/blocks/BlocksService.ts
```diff
@@ -437,8 +437,15 @@ export class BlocksService extends AbstractService {
...
-					const call = this.api.createType('Call', argument.toHex());
-					newArgs[paramName] = this.parseGenericCall(call);
+					try {
+						const call = this.api.createType(
+							'Call',
+							argument.toHex()
+						);
+						newArgs[paramName] = this.parseGenericCall(call);
+					} catch {
+						newArgs[paramName] = argument;
```
maciejhirsz, line 447: If we are not expecting the error to happen mostly, it might make sense to log the errors?

emostov:
Relates to https://github.com/polkadot-js/api/issues/2753

src/services/blocks/BlocksService.ts
```diff
@@ -437,8 +437,15 @@ export class BlocksService extends AbstractService {
...
 					// multiSig.asMulti.args.call is an OpaqueCall (Vec<u8>) that we
 					// serialize to a polkadot-js Call and parse so it is not a hex blob.
-					const call = this.api.createType('Call', argument.toHex());
-					newArgs[paramName] = this.parseGenericCall(call);
+					try {
+						const call = this.api.createType(
+							'Call',
+							argument.toHex()
+						);
+						newArgs[paramName] = this.parseGenericCall(call);
```
dvdplm, line 445: How deep can this go? It would be good to have a recursion limit and give up after `n` levels of recursion or someone could submit crazy payloads and take down sidecar when the stack blows?

src/services/v0/v0blocks/BlocksService.ts
```diff
@@ -404,8 +404,17 @@ export class BlocksService extends AbstractService {
...
-					const call = this.api.createType('Call', argument.toHex());
-					newArgs[paramName] = this.parseGenericCall(call);
+					// multiSig.asMulti.args.call is an OpaqueCall (Vec<u8>) that we
+					// serialize to a polkadot-js Call and parse so it is not a hex blob.
+					try {
+						const call = this.api.createType(
+							'Call',
+							argument.toHex()
+						);
+						newArgs[paramName] = this.parseGenericCall(call);
```
dvdplm, line 414: Same question as above

src/services/blocks/BlocksService.ts
emostov: This goes to an arbitrary depth. However it only processes extrinsics in blocks that the node may have. So presumably Substrate would never have an extrinsic with an insane depth of nested call arguments that would blow a node.js stack.

PR #323, opened 2020-10-21
fix: Use correct registry when parsing extrinsic `call` arguments

emostov:
closes #322 

relates to https://github.com/polkadot-js/api/issues/2753

Previously we were using the `ApiPromise` registry to decode `OpaqueCalls`. It turns out this registry only has the latest metadata and will not have historic metadata. So when trying to parse some historic `Call`s, the bit of code changed by this PR would fail because it had the incorrect metadata for decoding.

This PR introduces the use of the blocks registry, which should always have the correct metadata to decode the extrinsics within. In theory we could use the registry off of any type from one of the historic queries, but I think using the block's registry is the is the most intuitive and readable.

Other note is that I updated all deps to latest versions.


maciejhirsz (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -396,11 +396,12 @@ export class BlocksService extends AbstractService {
 	 * @param argsArray array of `Codec` values
 	 */
```
insipx, line 397: Should we put doc-comment for `block` too?

src/services/blocks/BlocksService.ts
```diff
@@ -414,7 +415,10 @@ export class BlocksService extends AbstractService {
 	 *
 	 * @param genericCall `GenericCall`
 	 */
```
insipx, line 417: same here

src/services/v0/v0blocks/BlocksService.ts
```diff
@@ -363,11 +363,12 @@ export class BlocksService extends AbstractService {
 	 * @param argsArray array of `Codec` values
 	 */
```
insipx, line 364: and here

src/services/v0/v0blocks/BlocksService.ts
```diff
@@ -381,8 +382,11 @@ export class BlocksService extends AbstractService {
 	 *
 	 * @param genericCall `GenericCall`
 	 */
```
insipx, line 384: block @ param doc comment

insipx:
LGTM apart from the doc-comments I think

src/services/v0/v0blocks/BlocksService.ts
```diff
@@ -381,8 +382,11 @@ export class BlocksService extends AbstractService {
...
-	private parseGenericCall(genericCall: GenericCall): ISanitizedCall {
-		const { sectionName, methodName, callIndex } = genericCall;
+	private parseGenericCall(
+		genericCall: GenericCall,
+		block: Block
```
dvdplm, line 387: Would it be possible to pass around the `registry` rather than the block? It would clarify the intent a bit (maybe it’s more efficient too?). Without knowing the context I think I would have had to read a fair amount of code before figuring out what the second params is for.

src/services/v0/v0blocks/BlocksService.ts
```diff
@@ -395,26 +399,34 @@ export class BlocksService extends AbstractService {
...
+					newArgs[paramName] = this.parseGenericCall(call, block);
 				} else {
 					newArgs[paramName] = argument;
 				}
 			}
 		}
 
 		return {
-			method: `${sectionName}.${methodName}`,
-			callIndex,
```
dvdplm, line 417: Why remove the callIndex here? It’s a private method so I guess it can change at any time, but curious why?

dvdplm (approved):
This works. It’s not pretty but it gets the job done. Left a couple of questions but feel free to merge.
My more general concern about unbounded recursion on the other hackfix-PR still stands but shouldn’t block this ofc.

src/services/v0/v0blocks/BlocksService.ts
emostov: Thats a mistake! Good catch!

src/services/v0/v0blocks/BlocksService.ts
emostov: I will change to just the registry. I actually did the whole block just for readability and to make it clear the registry is associated with the block

PR #345, opened 2020-11-17
chore(release): 1.1.0

emostov:
Corresponds to tag v1.1.0 and the most recently published NPM package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## [1.1.0](https://github.com/paritytech/substrate-api-sidecar/compare/v1.0.0-rc4...v1.1.0) (2020-11-17)
+## [1.1.0](https://github.com/paritytech/substrate-api-sidecar/compare/v1.0.0...v1.1.0) (2020-11-17)
 
 
 ### ⚠ BREAKING CHANGES
 
 * Bump polkadot-js and document runtime/metadata API regression (#338)
-* Remove all v0 routes to prepare for v1 release
 
 ### Features
 
 * Bump polkadot-js and document runtime/metadata API regression ([#338](https://github.com/paritytech/substrate-api-sidecar/issues/338)) ([effc5eb](https://github.com/paritytech/substrate-api-sidecar/commit/effc5eb159587b2b3c333f0f545b8a3fe793c789))
-* Remove all v0 routes to prepare for v1 release ([410a2e9](https://github.com/paritytech/substrate-api-sidecar/commit/410a2e9251bf341b9e0f151bccf9c83617c7673f))
 
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### Features
 
-* Bump polkadot-js and document runtime/metadata API regression ([#338](https://github.com/paritytech/substrate-api-sidecar/issues/338)) ([effc5eb](https://github.com/paritytech/substrate-api-sidecar/commit/effc5eb159587b2b3c333f0f545b8a3fe793c789))
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [1.1.0](https://github.com/paritytech/substrate-api-sidecar/compare/v1.0.0...v1.1.0) (2020-11-17)
+
+
+### ⚠ BREAKING CHANGES
+
+* Bump polkadot-js and document runtime/metadata API regression (#338)
```
dvdplm, line 10: Ideally users would be able to learn what is breaking exactly; maybe link to the pd.js changelog?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0",
+  "version": "1.1.0",
```
dvdplm, line 2: Hmm, if this is a semver breaking release shouldn't this be `2.0`?

emostov: Technically, yes. But since its such a small breaking change that may be reverted in the near future I was thinking of having a note but keeping bumping minor like we do here.

CHANGELOG.md
emostov: Added a link

package.json
dvdplm: I'm of two minds on this. One voice in my head says "Last famous words! A change is either breaking or it's not, no such thing as a 'small' breaking change! Don't stray from semver." and the other is saying "Release 2.0 for this is just silly; and not releasing is also silly, so what do you want to do?".

As this is a fix for a regression downstream maybe we should treat this as a patch release and change the wording in the change log from `**BREAING**` to something less alarming sounding and instead spend some time explaining the circumstance(s) where this could break user code?

maciejhirsz: I'm with David that this should be rather 2.0.0 or 1.0.1, pending on how we judge the breakiness:

+ If the breaking change is actually fixing a bug, rather than affecting a contract with end users, then it's 1.0.1.
+ If it's actually breaking, no matter how small, then it's 2.0.0.

I get that going from 1.0.0 to 2.0.0 quickly _feels_ weird, but I also don't think we necessarily guarantee 100% api correctness for end users, so they should expect breaking changes down the road. If we make it a practice now then switching from 19.0.0 to 20.0.0 later will feel less weird.

danforbes: Zeke, can you please describe exactly why you consider this change "breaking"? As Maciej says, think in terms of fixing unintended/"buggy" behavior versus breaking "valid" end-user expectations, even if they are not stated explicitly. We shouldn't support people relying on buggy behavior in production, but we should endeavor not to break existing workflows. If we believe this change will break valid, existing workflows, we should bump the major version.

emostov: I guess the issue is that the intended behavior in the first place is not clear IMO. We are going from having metadata with pokadot-js specific type aliases to having the plain substrate metadata (at least from what I understand.) What I am not clear about is if our users expect pokadot-js specific metadata (or if anyone is even programmatically using this endpoint).
 
The other factor here is that we already released a v1.1.0 on NPM to quickly fix an issue with dependencies, and NPM does not allow you to release a decreased semver version, so 1.0.1 is out of the option there and I think would be confusing to have something else here

dvdplm: > What I am not clear about is if our users expect pokadot-js specific metadata (or if anyone is even programmatically using this endpoint).

Sounds like it is a breaking change, just one that likely affects very few people. If so, I think we should go with 2.0 (and not be coy about using major numbers; after all integers are plentiful and cheap).

DQ: doesn't npm allow releasing patches to older versions? Like, relase v3.2.0 and also v2.9.1?

danforbes: We could release as a v2-rc or something as well

emostov: >. DQ: doesn't npm allow releasing patches to older versions? Like, release v3.2.0 and also v2.9.1
I think this is correct

emostov:
Closing in favor of #349

Issue #343, opened 2020-11-16
westend: Error: findMetaCall: Unable to find Call with index 0x9004


islishude:
**Description**

code: lastest commit
nodejs: v14 LTS
network: westend testnet

**Steps to Reproduce**

request for `/blocks/2223354` and get `Error: findMetaCall: Unable to find Call with index 0x9004`

**Expected vs. Actual Behavior**

should no error here

emostov:
Thanks for the details on how to reproduce the issue.

**tl;dr** I have a temporary fix with a working branch here: https://github.com/paritytech/substrate-api-sidecar/tree/zeke-try-opaque

It looks like Sidecar is error-ing here: https://github.com/paritytech/substrate-api-sidecar/blob/eeea29b74ef50eb45356e4a7e1ea04344097cc00/src/services/blocks/BlocksService.ts#L451 
when it tries to decode the opaque call.

We have had issues here before, but they should have been fixed when we started using the registry specific to the block with this PR: https://github.com/paritytech/substrate-api-sidecar/pull/323/files. So I am not sure exactly what the issue is

PR #347, opened 2020-11-18
fix: Catch errors decoding opaque calls

emostov:
Currently we decode `OpaqueCall`s in order to give more meaningful responses.

Recently, we ran into a scenario where the `OpaqueCall` argument simply appears to be invalid (see #343, https://github.com/polkadot-js/api/issues/2858). In this case the decoding will always fail. 

In the current release the whole endpoint will fail when the `OpaqueCall` is bad. This PR introduces a `catch` for the throw, and will now respond with the raw, hex encoded scale when the decoding fails.

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -937,7 +937,7 @@ components:
         args:
           type: object
           description: >-
-            Object of arguments keyed by parameter name. Note: if you are expecting a [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
+            Object of arguments keyed by parameter name. Note: if you are expecting an [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
             and it is not decoded in the response (i.e. it is just a hex string), then Sidecar was
             not able to decode it and likely that it is not a valid call for the runtime.
         tip:
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -936,7 +936,10 @@ components:
           format: unsignedInteger
         args:
           type: object
-          description: Object of arguments keyed by parameter name.
+          description: >-
+            Object of arguments keyed by parameter name. Note: if you are expecting a [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
+            and it is not decoded in the response (i.e. it is just a hex string), then Sidecar was
+            not able to decode it and likely that it is not a valid call for the runtime.
         tip:
           type: string
           description: Any tip added to the transaction.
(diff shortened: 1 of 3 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -448,8 +448,18 @@ export class BlocksService extends AbstractService {
...
+						const call = registry.createType(
+							'Call',
+							argument.toHex()
+						);
+						newArgs[paramName] = this.parseGenericCall(
+							call,
+							registry
+						);
+					} catch {
+						newArgs[paramName] = argument;
```
dvdplm, line 461: Would be good to log a warning here maybe? Or is there a valid scenario where users pass in a bag of bytes that we're not supposed to understand how to parse?

Can we make `parseGenericCall` return a richer error type and use `catch(err)` to present something that makes sense to users?

dvdplm (approved):

src/services/blocks/BlocksService.ts
emostov: I think logging a message to stdout could be ok, but I would think many users are not checking logs unless the endpoint returns 400 + codes.

We could also put something into the response noting the bytes are not decodable. However, I am not sure where that message would go. Right now the response maps to exactly in the block, and in this case we would need to add a field somewhere. I also think adding the field right here could be strange since these are supposed to be the arguments to extrinsic. 

A good middle ground could be putting a note in the docs that if you are expecting a decoded OpaqueCall and you instead get an encoded one it is likely that it is not decodable. The user could then try decoding themselves.

emostov: Ok added docs

docs/src/openapi-v1.yaml
```diff
@@ -936,7 +936,10 @@ components:
           format: unsignedInteger
         args:
           type: object
-          description: Object of arguments keyed by parameter name.
+          description: >-
+            Object of arguments keyed by parameter name. Note: if you are expecting a [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
```
dvdplm, line 940:
```suggestion
            Object of arguments keyed by parameter name. Note: if you are expecting an [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -38,7 +38,9 @@ import AbstractController from '../AbstractController';
  *   - `method`: Extrinsic method.
  *   - `signature`: Object with `signature` and `signer`, or `null` if unsigned.
  *   - `nonce`: Account nonce, if applicable.
- *   - `args`: Array of arguments.
+ *   - `args`: Array of arguments. Note: if you are expecting a [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
```
dvdplm, line 41:
```suggestion
 *   - `args`: Array of arguments. Note: if you are expecting an [`OpaqueCall`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/type.OpaqueCall.html)
```

dvdplm (approved):

niklasad1 (approved):
LGTM

PR #355, opened 2020-11-30
fix: Use http BadRequest (400) when balance-info not found

### LINKED ISSUE #346: Use `http-errors` in `accounts/{accountId}/balance-info`
To align with the rest of the service, this should use `http-errors`: 

https://github.com/paritytech/substrate-api-sidecar/blob/eeea29b74ef50eb45356e4a7e1ea04344097cc00/src/services/accounts/AccountsBalanceInfoService.ts#L49

Should look similar too:

https://github.com/paritytech/substrate-api-sidecar/blob/eeea29b74ef50eb45356e4a7e1ea04344097cc00/src/services/accounts/AccountsStakingInfoService.ts#L31

emostov:
Closes #346 

Also uses `.unwrap()` on the `Compact` block number, which should be more idiomatic then using `.toNumber()`.

### DIFF
--- a/src/services/accounts/AccountsBalanceInfoService.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.ts
@@ -1,4 +1,5 @@
 import { BlockHash } from '@polkadot/types/interfaces';
+import { BadRequest } from 'http-errors';
 import { IAccountBalanceInfo } from 'src/types/responses';
 
 import { AbstractService } from '../AbstractService';
@@ -29,7 +30,7 @@ export class AccountsBalanceInfoService extends AbstractService {
 
 		const at = {
 			hash,
-			height: header.number.toNumber().toString(10),
+			height: header.number.unwrap().toString(10),
 		};
 
 		if (account && locks && sysAccount) {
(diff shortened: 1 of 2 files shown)


src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -29,7 +30,7 @@ export class AccountsBalanceInfoService extends AbstractService {
 
 		const at = {
 			hash,
-			height: header.number.toNumber().toString(10),
+			height: header.number.unwrap().toString(10),
```
dvdplm, line 33: afaiu this isn't anything like the Rust `unwrap()` right? It literally removes the wrapping?

dvdplm (approved):

src/services/accounts/AccountsBalanceInfoService.ts
niklasad1: haha, yeah this is confusing when `unwrap` is a nasty habit in Rust :)

niklasad1 (approved):

PR #351, opened 2020-11-24
feat: chainSpec based controller config; Types from apps-config

emostov:
Closes #334 

## Overview 

This PR introduces the foundations of allowing Sidecar to work smoothly with FRAME based chains. It accomplishes this by 
1) including type definitions from `@polkadot/apps-config` which are injected into the API
2) Creating a way to specify which controllers to mount based on a chains `specName` and hard coded config that aligns with the `specName`. 

These changes will allow most FRAME chain builders to use Sidecar by just specifying end points their chain supports in the controllers config and keeping there chain types up to date in `@polkadot/apps-config`.

In this PR I also include controller configuration for Kulupu since it was requested in an issue (#350) and makes sense to have a live example to try out the setup.

Would be nice to get this through in the near future, since it will be a blocker for meaningful work in this repo.

#### Chain types
- Include chains types from `@polkadot/apps-config` and use the in `ApiPromise`

#### Controller config

- Controller configuration object that allows to specify the controllers to use (see `ControllerConfig` in `src/types/chains-config`)
- Mandala controller config (read down for more about controller configs)
- controller config for Kulupu to close #350 
- Create default controller config (primarily for Polkadot and Kusama)
- Dynamically instantiate controllers base on specified controller config (see `getControllersForSpec` and `getControllersFromConfig ` from `src/chains-config/index.ts`)
- `src/controllers/index.ts` now exports a concrete `controllers` object so we can have better type safety when dynamically creating the controllers with `getControllersFromConfig ` in `src/chains-config/index.ts`
- `finalizes` filed that is used to decide if the `getFinalizedHead` should be used in the blocks controller https://github.com/paritytech/substrate-api-sidecar/blob/0ac5573b25c5d4ddcbca705b439bf1b83f12cbf6/src/controllers/blocks/BlocksController.ts#L94

#### Endpoints (services/ controllers)

- ~~Creation of `KulupuBlocksController` which has new `/kulupuBlocks/{head, :number}` endpoint. I had to do this because there is no notion of finalized blocks since it is a simple PoW consensus (someone please correct if I am wrong). In the new controller only a couple lines here taken out to remove the `?finalized=boolean` query param, but I created an entire other controller with a new endpoint name to keep things simple; However I am certainly open to other ideas, I just don't want the controllers to scale with a huge amount of conditionals as we add more chains.~~
- `accounts/{accountId}/balance-info` 
    - now supports a token query param that allows the user to specify tokens from ORML tokens pallet
[...]

PR #357, opened 2020-12-01
config: Enhance CUSTOM_TYPES to take Object too.

amarts:
If a pallet is updating a block with type Object (Record<string, string>),
there is currently no way to provide a structure/Object interface type.
Enhancing the type of the CUSTOM_TYPES to take json object with
`Record<string, string>` type would fix the issues.

Signed-off-by: Amar Tumballi <amar@dhiway.com>

### DIFF
--- a/src/types/config/SidecarConfig.ts
+++ b/src/types/config/SidecarConfig.ts
@@ -1,3 +1,5 @@
+import { AnyJson } from '@polkadot/types/types';
+
 /**
  * Object to house the values of all the configurable components for Sidecar.
  */
@@ -9,7 +11,7 @@ export interface ISidecarConfig {
 
 interface ISidecarConfigSubstrate {
 	WS_URL: string;
-	CUSTOM_TYPES: Record<string, string> | undefined;
+	CUSTOM_TYPES: AnyJson;
 }
 
 interface ISidecarConfigExpress {
--- a/src/types/config/SidecarConfig.ts
+++ b/src/types/config/SidecarConfig.ts
@@ -1,4 +1,4 @@
-import { AnyJson } from '@polkadot/types/types';
+import { RegistryTypes } from '@polkadot/types/types';
 
 /**
  * Object to house the values of all the configurable components for Sidecar.
@@ -11,7 +11,7 @@ export interface ISidecarConfig {
 
 interface ISidecarConfigSubstrate {
 	WS_URL: string;
-	CUSTOM_TYPES: AnyJson;
+	CUSTOM_TYPES: RegistryTypes | undefined;
 }
 
 interface ISidecarConfigExpress {


amarts:
Looks like this has issues with `yarn build`. Things worked for me when I have `Record<string, any> | undefined` as the type, but `yarn lint` warns about using 'any' as the type (well, I too don't like it). 

As I am not an expert with TypeScript yet, would like to get advise on how to handle a object in the `config/types.json` file. Thanks in advance.

src/types/config/SidecarConfig.ts
```diff
@@ -9,7 +9,10 @@ export interface ISidecarConfig {
 
 interface ISidecarConfigSubstrate {
 	WS_URL: string;
-	CUSTOM_TYPES: Record<string, string> | undefined;
```
emostov, line 12: I think here it would make sense to use the polkadot-js helper type `AnyJson`: https://github.com/polkadot-js/api/blob/94a4594d320753054faf7f4ec92a362cc6603d7e/packages/types/src/types/helpers.ts#L10

This essentially is a recursively defined type that should align with any valid JSON.

We could import it at the top with
```
import { AnyJson } from '@polkadot/types/types';
```

This line could then change to
```
CUSTOM_TYPES: AnyJson
```

emostov (changes requested):

emostov:
Thank you for the PR! As I mentioned in the code comment, I think `AnyJson` from polkadot-js would be a good type to use here.

You might also be interested in #351, as that makes sidecar much more friendly to non polkadot/kusama substrate based chains by pulling in types from `@polkadot/apps-config` and allowing chain builders to config which routes to mount.

src/types/config/SidecarConfig.ts
amarts: Thanks. This is cool, and solves the problem. Will resubmit in a bit.

amarts:
Sure. #351 looks good. Will keep a watch.

meantime, did a `--force` push even though CONTRIBUTE guide suggests otherwise mainly to fix the commit message to reflect the actual change.

emostov:
Looks like CI is failing because we can't assume `AnyJson` is an object. To get around that I think we could do `Record<string, AnyJson>`

amarts:
Even that gives error:

```
amar@bulde:~/work/dhiway/substrate-api-sidecar$ yarn lint --fix
yarn run v1.22.0
$ tsc && eslint . --ext ts --fix
src/main.ts:60:3 - error TS2322: Type '{} | { [x: string]: AnyJson; }' is not assignable to type 'Record<string, string | Record<string, string> | Constructor<Codec> | { _enum: string[] | Record<string, string | null>; } | { _set: Record<string, number>; }> | undefined'.
  Type '{ [x: string]: AnyJson; }' is not assignable to type 'Record<string, string | Record<string, string> | Constructor<Codec> | { _enum: string[] | Record<string, string | null>; } | { _set: Record<string, number>; }>'.
    Index signatures are incompatible.
      Type 'AnyJson' is not assignable to type 'string | Record<string, string> | Constructor<Codec> | { _enum: string[] | Record<string, string | null>; } | { _set: Record<string, number>; }'.
        Type 'undefined' is not assignable to type 'string | Record<string, string> | Constructor<Codec> | { _enum: string[] | Record<string, string | null>; } | { _set: Record<string, number>; }'.

60  	types: {
    	~~~~~


Found 1 error.
```

amarts:
@emostov as you mentioned #351 fixes the issue in much cleaner way, I am happy to wait for that to get merged and released. It is ok to not spend time on this if it gets changed soon anyways.

Feel free to close this in that case.

emostov (approved):

emostov:
I pushed directly to the branch. Not sure why I didn't think of it earlier, but just pulled the exact type polkadot-js uses for the object - `RegistryTypes`. I think it should be all good to go now.

amarts (commented):
Thanks, Working for me :-)

insipx (approved):

Issue #367, opened 2020-12-17
Staking-payouts with large depths not successful


wpank:
Queries with large depths do not successfully return, such as:
(polkadot)
`/staking-payouts?depth=83&&era=197&unclaimedOnly=true`

They will return an error such as:

```
  message: 'request to http://<endpoint>/accounts/<account_address>/staking-payouts?depth=83&&era=197&unclaimedOnly=true failed, reason: socket hang up',
  type: 'system',
  errno: 'ECONNRESET',
  code: 'ECONNRESET'
  ````

This generally happens for any depth greater than 10. This used to work with pretty large depths.

**Comment by emostov** [comment]
Noting here that querys of depth 83 are consistently taking about 5 minutes against `wss://rpc.polkadot.io` using postman on my local machine. 

In google chrome, with `depth=3` and I above I consistently get the error:
```
Your connection was interrupted
A network change was detected.
ERR_NETWORK_CHANGED
```

My hypothesis is that various clients have some variant of a time out error and Sidecar it self is not necessarily failing or even hanging.

My next steps will include closely investigating the exact error messages, looking into optimzations of the endpoint and potentially redesigning the end point interface in a way that would be less susceptible to time errors.

One potential design brainstormed by @danforbes essentially involves a POST endpoint to create a resource that is the status of the query (returns an ID) and then a GET endpoint to poll the resource, returning the status of the query and possibly the result when done.

PR #372, opened 2020-12-20
refactor: Optimize `accounts/{accountId}/staking-payouts` blocking query complexity

emostov:
Likely closes: #367

In the previous iteration of `AccountsStakingPayoutsService` we had M * N batched queries (read: queries `await`-ed within a`Promise.all`) , where M is the depth of eras and N is maximum number of validators nominated by `address` at any given era.

This PR dramatically reduces the endpoint complexity and response time by always only waiting for two batches of queries to resolve. So we have simplified the blocking query complexity from O(m*n) => O(1).

We achieve this by taking advantage of the fact that we can dynamically build arrays of promises wrap them in `await Promise.all`.

Below is a picture of Sidecar logs, left is master and right is this branch (look at the end of log message for response time in milliseconds):

Screen Shot 2020-12-19 at 9 07 59 PM


src/services/accounts/AccountsStakingPayoutsService.ts
```diff
@@ -55,105 +88,202 @@ export class AccountsStakingPayoutsService extends AbstractService {
...
+			era
+		);
+
+		// With the general data, we can now fetch the commission of each validator `address` nominates
+		const allErasCommissions = await this.fetchAllErasCommissions(
+			api,
+			hash,
+			address,
+			startEra,
+			allErasGeneral.map((el) => el[0])
```
dvdplm, line 114: A bit lost here: what is in `el[0]` here? 
Did I get this right? 
- `allErasGeneral` is a `Promise.all` of a collection of `Promise.all`s, each making 3 RPC queries; 
- we `await` it so all promises are resolved and at this point `el[0]` is the result of calling `api.derive.staking.eraExposure(eraIndex)`
-  the other two RPCs are used later and returned as part of `allEraData`

It's a bit unfortunate that we have to loop over this collection twice, once here and then again in the `map()` call on line 118. Not sure what to suggest though.

dvdplm (approved):
10x improvement! 🍰

src/services/accounts/AccountsStakingPayoutsService.ts
```diff
@@ -339,4 +454,35 @@ export class AccountsStakingPayoutsService extends AbstractService {
...
+		deriveEraExposure: DeriveEraExposure
+	): DeriveEraExposureNominating[] | undefined {
+		let nominatedExposures: DeriveEraExposureNominating[] | undefined =
+			deriveEraExposure.nominators[address];
+		if (deriveEraExposure.validators[address]) {
+			// We treat an `address` that is a validator as nominating itself
+			nominatedExposures = nominatedExposures
+				? nominatedExposures.concat({
+						validatorId: address,
+						validatorIndex: 0,
```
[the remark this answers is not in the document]
emostov: We don't actually, but it's never used - we just do it so the array has the expected type.  I suppose we could do some number like 9999 to show its never used

src/services/accounts/AccountsStakingPayoutsService.ts
emostov: > It's a bit unfortunate that we have to loop over this collection twice, once here and then again in the map() call on line 118. Not sure what to suggest though.

1) I am not too concerned. ATM this array will never be over 84, so while technically linear to depth its complexity has a hard bound

src/services/accounts/AccountsStakingPayoutsService.ts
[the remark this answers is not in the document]
emostov: Added comment and put in `9999` instead

src/services/accounts/AccountsStakingPayoutsService.ts
```diff
@@ -466,21 +467,15 @@ export class AccountsStakingPayoutsService extends AbstractService {
 		address: string,
 		deriveEraExposure: DeriveEraExposure
 	): DeriveEraExposureNominating[] | undefined {
-		let nominatedExposures: DeriveEraExposureNominating[] | undefined =
-			deriveEraExposure.nominators[address];
+		let nominatedExposures: DeriveEraExposureNominating[] =
+			deriveEraExposure.nominators[address] ?? [];
```
dvdplm, line 471: dq: why is `??` better here than plain old `||`?

emostov: No amazing reason, just figured it was more descriptive because it shows we just expect undefined/null.

TarikGul (approved):

TarikGul:
lgtm

PR #376, opened 2021-01-05
fix: Account for polkadot-js changes; Harden `createCalcFee`

emostov:
This PR bumps deps, accounts for changes from polkadot-js/api, and hardens `createCalcFee` which was affected by the polkadot-js changes.

Changes introduced by PR:
- `createCalcFee`
    - Forcibly serialize `AbstractInt`s with `toString(10)`,  making sure they are decimal form
    - Use native `toBigInt` on relevant `AbstractInt` 
- Manually create each extrinsic as a `Call` type before processing args
- Use `section` and `method` instead of the no longer available `sectionName` and `methodName` 
- wasm-pack introduced a new auto-generated method into `calc.js`, resulting a big diff there


dvdplm (approved):
lgtm

calc/src/calc_fee.rs
```diff
@@ -98,6 +98,10 @@ impl CalcFee {
         spec_version: u32,
     ) -> Option<CalcFee> {
         debug::setup();
+	    info!(
```
TarikGul, line 101: Spacing

TarikGul (approved):
Spacing in that one file `calc/pkg/calc_fee.rs`

PR #383, opened 2021-01-13
feat: State tracing for balance reconciliation 

emostov:
Relates to:
- https://github.com/paritytech/substrate/pull/7780
- https://github.com/paritytech/substrate-reconciler
- [`state_TraceBlock` RPC trace hierarchy for Polkadot & Kusama](https://docs.google.com/drawings/d/1vZoJo9jaXlz0LmrdTOgHck9_1LsfuQPRmTr-5g1tOis/edit?usp=sharing)

# Overview

This PR introduces the following endpoints:

- `/experimental/blocks/{blockId}/traces/operations?actions=false`
-  `/experimental/blocks/head/traces/operations?actions=false`
- `/experimental/blocks/{blockId}/traces`
- `/experimental/blocks/head/traces`

The first class goal is to provide an account balance state transition primitive that can be used for balance reconciliation. We do that by modeling each state transition to an account's balance with an `Operation`.

Sample response:

- [`/experimental/blocks/4403377/traces/operations`](https://gist.github.com/emostov/3df825eec2748052c2e0a1dcd0130be2)
- [`/experimental/blocks/4403377/traces`](https://gist.github.com/emostov/45b989c98a2b475bb6929992a97b3ff2)

## Why balance reconciliation is needed

Some known Substrate chain integrators and their auditors need to be able reconcile the chain's state at some time point against all the operations that took place leading up to it. 

## Why event based balance reconciliation is not feasible

Current known attempts ([primitive example](https://github.com/emostov/balance-reconcile/blob/95f88cd8f29b27febd3ba4451971ca08bac1fb63/src/classes/one_time_reconciler.ts#L121-L134)) at balance reconciliation rely on transaction parameters and emitted events, but this poses two primary problems:

1) Event coverage is not sufficient to account for all state changes, and
1) It would be an anti-pattern to make them sufficient. A core principle of Substrate is that the runtime itself is on-chain so any WASM host can validate a state transition.

## The solution

The solution to this problem is to attribute every state change to a pallet hook (`on_finalize` & `on_initialize`) or extrinsic application. A non-event-based, state tracing solution can have the following benefits:

- Let chain builders use events strictly for things that applications/parachains would care about.
- Reduce the bloat of events that only exist for reconciliation.

Beyond balance reconciliation, state tracing offers the potential to track changes to arbitrary storage items. A few examples:

- Generic solution to getting the `partialFee` of an extrinsic
- Tracking a particular storage item that belongs to an account

(above text is based off of original project outline from @joepetrowski ) 

## Accounting balance changes with `Operation`s

This PR introduces the notion of an `Operation` to model balance change for accounts in an audit ready form. On an abstract level an `Operation` is a single state change to an account and represents one side of a transaction.

For example, If Alice transfers 100unit to Bob, this would be modeled with two `Operation`s:

1) Alice = Alice - 100unit
1) Bob = Bob + 100unit

By summing the `Operation`s belonging to a single account from genesis up until a certain block, we should have the same account balance as stored in an archive chains DB for that block height.

In this PR we expose `Operation`s as 

```JSON
// Note: `eventIndex`, `parentSpandId` and `primarySpanId` are not strictly necessary
// but are left in to make it easier to understand how a `Operation` corresponds to
// the raw trace data.
        {
            "phase": {
                "variant": "applyExtrinsic",
                "extrinsicIndex": "2"
            },
            "parentSpanId": {
                "name": "approve_as_multi",
                "target": "pallet_multisig",
                "id": "21"
            },
            "primarySpanId": {
                "id": "20",
                "name": "apply_extrinsic",
                "target": "frame_executive"
            },
            "eventIndex": "12",
            "address": "GCMo2fMUXEsP6FJ2qyxGd9KdnejbtyRZJ3ACUd8jdNsgWGM",
            "storage": {
                "pallet": "system",
                "item": "Account",
                "field1": "data",
                "field2": "free"
            },
            "amount": {
                "value": "-3358666656520",
                "currency": {
                    "symbol": "KSM"
                }
            }
        },
```

**Note:** This PRs creates `Operation`s in a way that is hard coded to Kusama/Polkadot, which will likely break and is not generically applicable. However, the schema for an `Operation` and the process for reconciliation introduced here should be generic across an `FRAME`-based chain

### Terms

- `span`: [see tracing crate.](https://docs.rs/tracing/0.1.25/tracing/span/index.html)
- `event`: [see tracing crate.](https://docs.rs/tracing/0.1.25/tracing/event/struct.Event.html)
- `executeBlockSpan`: The span that all other trace spans and events are within.
- `primarySpan`: A span that is a direct descendant (i.e. depth of 1) of `executeBlockSpan`.
- `actionGroup`: A group of spans and events that conceptually count as one action in block execution i.e a pallet hook or an extrinsic.

### state_traceBlock RPC

We use the `state_traceBlock` RPC to re-excute a block and get state tracing data. The response result is identical to the object keyed by the `traces` field of the  `/experimental/blocks/{head, blockId}/traces`. ([see here](https://gist.github.com/emostov/45b989c98a2b475bb6929992a97b3ff2)). Consult the [docs](https://github.com/paritytech/substrate/pull/7780/files#diff-a470f7227faf378aa4b8cf63f5b3d35018da37803f6bf0456327e4f6d2e15250R140-R245) for detailed info. 

### Creating `Operations`

Pseudo proces:

1) find the `executeBlock` span. This can be thought of as the root span.
1) find the `primarySpan`s, which are direct descendants of the `excuteBlockSpan`
1) for each `primarySpan` create an `actionGroup` by gathering all its descendant spans and events.
1) for each `actionGroup` determine the phase of block execution by looking at the primary and secondary spans metadata.
1) for each `actionGroup` that is of phase `apply_extrinsic`, find the extrinsic index by looking at the extrinsic index storage read events.
1) for each `actionGroup` find all the events that are reads of the `frame_system::AccountInfo` storage item and organize them by address.
1) for each set of `AccountInfo` events organized by address for each `actionGroup` create Operation's by
    1) since events are linearly sequential, compare each field of `AccountInfo.data` from the previous event to the current event. 
    1) If there is a delta for a field create an `Operation` where the `Operation.amount.value` is the delta. We iterate through all the `AccountInfo` events for an address in an `actionGroup`. Note: we use the `actionGroup` metadata to create the `Operation`s metadata
1) for each `actionGroup`, sort its `Operation`s by their corresponding event index (Can be skipped if we just sort all the `Operation`s at then end or if we don't care that `Operation` lines up to event order). 
1) merge together `Operations` from all `actionGroup`s. If `actionGroups` are sorted by time entered of their primary span, we can merge `Operation`s in the order of `actionGroup`s primary spans. If `actionGroup`s are not sorted we need to sort `Operation`s by their underlying event index.

### Should we expose `actionGroups` ?

I don't think `actionGroup`s need an exact schema because they are instead a conceptual step in the process of attributing transactions to a certain point in block execution. However we are easily able to return what we use internally as the `actionGroup`. I think having it accessible is at the very least nice for understanding and debugging and could even be useful for auditing the creation of `Operation`s. 

### How to organize `actionGroups`?

Phases are used to link a `Operation` to some part of the block execution pipeline. 
Normally one might divide this into three phases: `on_initialize`, `apply_extrinsic`, and `on_finalize`. 

At first I assumed actionGroups (which are formed around primarySpans) would naturally fit into one of the aforementioned phases. However I found that while actionGroups for `apply_extrinsic` and `on_finalize` have primary spans that directly correspond to those phases, no `primarySpans` directly correspond to `on_initialize`. Instead, `on_initialize` spans are children of the `init_block` `primarySpan`.

Currently, I assigned the `init_block` `actionGroup` with the phase `on_initialize`, thus giving all the resulting `Operation`s the phase `on_initialize`. I definitely think we should use the `on_initialize`, but the **question** I have is:
 - should we break the current rule that all `actionGroup`s correspond 1-to-1 with a `primarySpan` and instead make `on_initialize` `actionGroups` that correspond to `on_initialize` spans? 
 
 The upside is that I think the `actionGroup`s would make more sense conceptual; i.e. a single pallet's hook would be have an `actionGroup` and a single extrinsic would have an `actionGroup`. As it stands a single pallet's `on_finalize` hook and a single extrinsic would have their own `actionGroup`, but a single pallet's `on_initialize` hook is grouped in together with all other pallet's `on_initialize` hooks under the `init_block` `actionGroup`.
 
The downside is that the `actionGroup` stray from closely aligning to the trace structure (may not actually be negative) and it would add some complexity to `actionGroup` generation.

Either way though, in the current design, it should not affect the `Operation`'s that the user sees - so we could push this decision to later without breaking things.

## Future work

- create storage keys based on metadata of current chain #528 
- generic storage item change tracking. i.e given a storage item name (and maybe a map key), give all the deltas to that storage item in a single block trace. Not sure if this would go in api-sidecar.
- add rpc def to polkadot-js

TODO
- [x] Better PR description
- [x] Docs
- [x] Clean up `Trace` - better comments, look for performance improvements
- [x] Tests
- [x] Expand on open questions
- [x] Investigate different strategy for sorting `Operation`s (maybe can do after tests are implemented to see if anything changes)
- [x] Break up confusing one liners in `Trace`
 -  ~~Create ActionGroups where `onInitialize` is a primary span.~~ nixing for now, too much complexity for something we are not even sure will add value and will make reviewing tougher.
- [x] Determine why we cannot do a sort when going through an individual account's events
- [x]  Improve types for events as they are now mutated in place quite a bit
- [x] Add a notes somewhere in docs that 0 value Ops do not exist. i.e a balance transfer to/from the same account will not show up as an Op, except for the fees.


emostov:
Here is where I am currently at: https://gist.github.com/emostov/831045a9d55bb41adcb1cb11981a0852

The first field represents a `Map<Address, Array<Transition>>`. `Transition` is taking sequentially events that contain the `system::account` for the address and calculating the delta for the account.

src/services/blocks/BlocksService.spec.ts
```diff
@@ -475,6 +479,7 @@ describe('BlocksService', () => {
 			checkFinalized: false,
 			queryFinalizedHead: false,
 			omitFinalizedTag: false,
+			operations: false,
```
emostov, line 482: Remove this

src/main.ts
```diff
@@ -39,10 +39,76 @@ async function main() {
...
+						isHistoric: true,
+					},
+				],
+				type: 'Json',
+			},
+		},
+	};
+
+	// Note: These are not working - currently just using `Json` type
+	// const _types = {
```
emostov, line 73: Remove these commented out types

src/main.ts
```diff
@@ -39,10 +39,76 @@ async function main() {
 
 	logger.info(`Version: ${packageJSON.version}`);
 
+	// curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "state_traceBlock", "params": ["0xb246acf1adea1f801ce15c77a5fa7d8f2eb8fed466978bcee172cc02cf64e264"]}' http://localhost:9933/
+	const rpc = {
+		state: {
+			traceBlock: {
+				description: 'Blcok state trace prototype',
```
TarikGul, line 46:
```suggestion
				description: 'Block state trace prototype',
```

src/main.ts
TarikGul: Friendly reminder that these have to get removed.

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+import { InternalServerError } from 'http-errors';
+
+import {
+	BlocksTrace,
+	BlocksTraceOperations,
+} from '../../types/responses/BlocksTrace';
+import { AbstractService } from '../AbstractService';
+import { BlockTraceResponse, isBlockTrace, isTraceError, Trace } from './trace';
+
+const DEFAULT_TARGETS = 'pallet,frame,state';
```
TarikGul, line 11: This seems extremely odd as an input, is this how `this.api.rpc.state.traceBlock()` expects the input?

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+// :extrinsic_index & frame_system::Account
+const DEFAULT_KEYS =
+	'3a65787472696e7369635f696e646578,26aa394eea5630e07c48ae0c9558cef7b99d880ec681799c0cf30e8886371da9';
+const UNEXPECTED_RPC_RESPONSE = 'Unexpected response to state_traceBlock RPC';
```
TarikGul, line 15: Just double checking but should it be `state_trace_block`, just looks a little weird but good chance I am wrong here?

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,203 @@
...
+
+export interface TraceError {
+	error: string;
+}
+
+export function isTraceError(
+	thing: unknown
+): thing is { traceError: TraceError } {
+	return (
+		typeof (thing as { traceError: TraceError })?.traceError?.error === 'string'
```
TarikGul, line 94: At first I was confused by this, and the return type of the traces call, but now it semi makes sense. The one thing that still confuses me a little is why would the `traceError` be undefined if it was returned in the object instead of `blockTrace`. Is this just a nuance of the type returned from polkadot-js?

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+			return {
+				at: {
+					hash,
+					height: block.header.number.unwrap().toString(10),
+				},
+				operations,
+				actions: includeActions ? actions : undefined,
+			};
+		} else {
+			throw new InternalServerError(UNEXPECTED_RPC_RESPONSE);
```
TarikGul, line 99: What kind of situation would cause this conditional to happen where `traceError` wouldn't catch it?

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	/**
+	 * Get the balance changing operations induced by a block.
+	 *
+	 * @param hash `BlockHash` to get balance transfer operations at.
+	 * @param includeActions wether or not to include `actions` field in the reponse.
```
TarikGul, line 57:
```suggestion
	 * @param includeActions whether or not to include `actions` field in the response.
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		name: 'on_initialize',
+		// Targets vary by pallet
+	},
+	onFinalize: {
+		name: 'on_finalize',
+		// Targets vary by pallet
+	},
+};
+
+// In the future this will need a more chain specific solution that creates
```
TarikGul, line 51: Should we make an issue for this?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+function getKeyNames(api: ApiPromise): Record<string, KeyInfo> {
+	const moduleKeys = Object.keys(api.query).reduce((acc, mod) => {
+		Object.keys(api.query[mod]).forEach((item) => {
+			const queryObj = api.query[mod][item];
+			let key;
+			try {
+				// Slice off '0x' prefix
+				key = queryObj.key().slice(2);
+			} catch {
+				key = queryObj.keyPrefix().slice(2);
```
TarikGul, line 67: I didnt know this but if the `key` doesnt exist, there will always be a `keyPrefix`?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+	// IMPORTANT for downstream logic: Make sure spans are sorted.
+	spans.sort((a, b) => a.id - b.id);
+
+	const spansById = new Map<number, SpanWithChildren>();
+	const parsedSpans = [];
+	let executeBlockSpanId;
+	for (const span of spans) {
+		const spanWithChildren = { ...span, children: [] };
+
+		if (
```
TarikGul, line 123: Is it expected to have only one span that fulfills this, or by sorting are you looking for the most recent span that fulfills this conditional?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		parsedSpans.push(spanWithChildren);
+		spansById.set(span.id, spanWithChildren);
+	}
+
+	if (executeBlockSpanId === undefined) {
+		throw new InternalServerError('execute_block span could not be found');
+	}
+
+	// Go through each span, and add its Id to the `children` array of its
+	// parent span, creating tree of spans starting with `execute_block` span as
```
TarikGul, line 139:
```suggestion
	// parent span, creating a tree of spans starting with `execute_block` span as
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		}
+
+		spansById.get(span.parentId)?.children.push(span.id);
+	}
+
+	return { spansById, parsedSpans, executeBlockSpanId };
+}
+
+/**
+ * Find the Ids of al the descendant spans of `root`.
```
TarikGul, line 159:
```suggestion
 * Find the Ids of all the descendant spans of `root`.
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+ *
+ * @param root span which we want all the descendants of
+ * @param spansById map of span id => span with children
```
TarikGul, line 162: This could use a better description. Little confusing for documentation.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+	// Note: traversal order doesn't matter here
+	const stack = spansById.get(root)?.children;
+	if (!stack) {
+		return [];
+	}
+
+	const descendants = [];
+	while (stack.length) {
+		const curId = stack.pop();
+		if (!curId) {
```
TarikGul, line 177: `if (!curId) break;` ?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		if (!extrinsicIndexBySpanId.has(id)) {
+			return acc;
+		}
+
+		acc.push(extrinsicIndexBySpanId.get(id) as BN);
+		return acc;
+	}, [] as BN[]);
+
+	if (extrinsicIndex.length === 0) {
+		// Since we assume the `spanIds` are for span from an apply extrinsic action group,
```
TarikGul, line 212:
```suggestion
		// Since we assume the `spanIds` are for a span from an apply extrinsic action group,
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		// we assume there is always an `:extrinsic_index` event
+		throw new InternalServerError('Expected at least one extrinsic index');
+	}
+
+	if (extrinsicIndex.length > 1) {
```
TarikGul, line 217: Interesting sanity check. Could an issue like the originate from the substrate rpc client, or would it always be the logic in this endpoints scope?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+
+/**
+ * Class for processing traces from the `state_traceBlock` RPC endpoint.
+ *
+ * Assumptions:
+ * - Spans do not start in sorted order.
+ * - Events start in sorted order.
+ *
+ * For a conceptual overview on traces from block excution and `Operation` construction
+ * consult [this diagram.](https://docs.google.com/drawings/d/1vZoJo9jaXlz0LmrdTOgHck9_1LsfuQPRmTr-5g1tOis/edit?usp=sharing)
```
TarikGul, line 241: 👍

src/services/blocks/BlocksTraceService.ts
TarikGul: resolved, i understand this is what it is now.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+			const secondarySpans: SpanWithChildren[] = [];
+			secondarySpansIds.forEach((id) => {
+				const events = eventsByParentId.get(id);
+				const span = spansById.get(id);
+
+				events && secondarySpansEvents.push(...events);
+				span && secondarySpans.push(span);
+			});
+
+			let phase: PhaseOther, extrinsicIndex: BN | undefined;
```
TarikGul, line 307:
```suggestion
			let phase?: PhaseOther, extrinsicIndex: BN;
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+				// This is an action group for an extrinsic application
+				extrinsicIndex = extractExtrinsicIndex(
+					// Pass in all the relevant spanIds of this action group
+					secondarySpansIds.concat(primary.id),
+					extrinsicIndexBySpanId
+				);
+
+				phase = Phase.ApplyExtrinsic;
+			} else {
+				// This likely an action group for a pallet hook or some initial
```
TarikGul, line 318: Confusing wording.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		const indexEncodedOption = event.data.stringValues.result;
+		if (!(typeof indexEncodedOption == 'string')) {
+			throw new InternalServerError(
+				'Expected there to be an encoded extrinsic index for extrinsic index event'
+			);
+		}
+
+		const scale = indexEncodedOption
+			.slice(2) // Slice of leading `Option` bits beacuse they cause trouble
+			.match(/.{1,2}/g) // Reverse the order of the bytes for correct endian-ness
```
TarikGul, line 406: Good job here!

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+	 * 2) Create a map of span id => array of children events
+	 * 3) Create a map of span id => to extrinisic index
+	 *
+	 * @param spansById map of span id => to span with its children for all spans
+	 * from tracing the block.
+	 */
+	private parseEvents(
+		spansById: Map<number, SpanWithChildren>
+	): {
+		eventsByParentId: Map<number, PEvent[]>;
```
TarikGul, line 427: Could this be something you would want to make into an Interface type?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+			if (isSome(extrinsicIndexOpt)) {
+				extrinsicIndexBySpanId.set(parsed.parentId, extrinsicIndexOpt);
+			}
+		}
+
+		return { eventsByParentId, extrinsicIndexBySpanId };
+	}
+
+	/**
+	 * Create a mapping adress => Account events for that address based on an
```
TarikGul, line 463:
```suggestion
	 * Create a mapping address => Account events for that address based on an
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+				(storagePath as PalletKeyInfo).item == 'account'
+			)
+		) {
+			throw new InternalServerError(
+				'Event did not have system::account path as expected'
+			);
+		}
+
+		// storage key = h(system) + h(account) + h(address) + address
+		//								^^^^^^^^^^^^^^^^^^^^storage item prefix
```
TarikGul, line 522: Not sure if this is just my Github but it looks very off spacing wise.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+		// Since this is a transparent key with the address at the end we can pull out
+		// the address from the key. Here we slice off the storage key + account hash.
+		const addressRaw = event.data.stringValues.key?.slice(96);
```
TarikGul, line 525: Is the key size fixed, might be worth noting its size in the comment above if possible.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,666 @@
...
+	 * on the previous event vs the current event.
+	 *
+	 * @param accountEventsByAddress map of address => events of that addresses `accountInfo`
+	 * events.
+	 */
+	private deriveOperations(
+		accountEventsByAddress: Map<string, PAccountEvent[]>
+	): Operation[] {
+		const ops = [];
+		for (const events of accountEventsByAddress.values()) {
```
TarikGul, line 561: You would know best, but just throwing it out there. Is a double for loop here absolutely necessary? (Assuming it is), but if it wasn't and there was a way to optimize it would it be worth it? As in is the current size of event's always going to be so small that it wont matter, or is the potential size of events large enough to cause issues scaling.

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,203 @@
...
+	 */
+	primarySpan: SpanWithChildren;
+	/**
+	 * Descendant spans of a primary span.
+	 *
+	 * For extrinsic application this will the actual extrinsic + any nested calls.
+	 */
+	secondarySpans: SpanWithChildren[];
+	/**
+	 * Events from the primary  and secondary spans.
```
TarikGul, line 169:
```suggestion
	 * Events from the primary and secondary spans.
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,203 @@
...
+	/**
+	 * Primary span is a direct descendant `executeBlock` span. Secondary spans are
+	 * a descendants of the primary span. (Secondary spans can be a descendant of
+	 * arbitrary depth.)
+	 */
+	primarySpan: SpanWithChildren;
+	/**
+	 * Descendant spans of a primary span.
+	 *
+	 * For extrinsic application this will the actual extrinsic + any nested calls.
```
TarikGul, line 165:
```suggestion
	 * For extrinsic application this will be the actual extrinsic + any nested calls.
```

TarikGul:
Just did my first initial review. Lot's going on in this PR. It definitely gets pretty complicated especially in `Trace.ts`. I need to get this branch active, and working in conjunction with the state tracing PR in substrate in order to get a more accurate understanding of the logic in some places. Love the work so far though. I was able to follow along for a lot of it.

src/services/blocks/trace/types.ts
emostov: The return type is an enum - it can be one of two variants:

```rust
pub enum TraceBlockResponse {
	/// Error block tracing response
	TraceError(TraceError),
	/// Successful block tracing response
	BlockTrace(BlockTrace)
}
```
see [here](https://github.com/paritytech/substrate/pull/7780/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-ab95a09c45453652418e5ecb30473d0fbbef66243b74bdd3486d56f37b164847R93-R98)
Because we are keying in deeply to an `unknown`, I figure we have better runtime safety if we check at every depth of the object for `undefined`

src/services/blocks/BlocksTraceService.ts
emostov: This will throw whenever the response JSON structure differs from what we expect. Ideally won't happen, but could capture a breaking change to the RPC

src/services/blocks/trace/Trace.ts
emostov: You mean type alias? Potentially .. not sure how to decide though.

src/services/blocks/trace/Trace.ts
emostov: It's github

docs/src/openapi-v1.yaml
```diff
@@ -921,6 +922,137 @@ paths:
...
+  /experimental/blocks/{blockId}/traces:
+    get:
+      tags:
+      - trace
+      summary: |
+        [Experimental - subject to breaking change.] Get traces for the most
+        recently finalized block.
+      description: | 
+        Returns traces (spans and events) of the bloc block from
+        RPC `state_straceBlock`. Consult the [RPC docs](TODO add link when pr is
```
emostov, line 952: TODO

docs/src/openapi-v1.yaml
```diff
@@ -921,6 +922,137 @@ paths:
...
+  /experimental/blocks/head/traces:
+    get:
+      tags:
+      - trace
+      summary: |
+        [Experimental - subject to breaking change.] Get traces for the most
+        recently finalized block.
+      description: | 
+        Returns traces (spans and events) of the most recently finalize block from
+        RPC `state_straceBlock`. Consult the [RPC docs](TODO add link when pr is
```
emostov, line 934: TODO

docs/src/openapi-v1.yaml
```diff
@@ -2048,6 +2180,119 @@ components:
...
+          $ref: '#/components/schemas/OperationAmount'
+    BlocksTraceOperations:
+      type: object
+      properties:
+        at:
+          $ref: '#/components/schemas/BlockIdentifiers'
+        operations:
+          type: array
+          items:
+            $ref: '#/components/schemas/Operation'
```
emostov, line 2295: Note: intentionally not including action group in the response. We should discuss if we think action groups should be available at all and if we want to make any guarantees about their stability.

src/services/blocks/trace/Trace.ts
emostov: Kind of both. If the substrate logic changes this may need to change. On the other hand if I am making faulty assumptions (namely that last read to an extrinsic index within an action group will always be for the current extrinsic index and not the next or former) the way we get the extrinsic index would need to change.

src/services/blocks/trace/Trace.ts
emostov: Only works for interfaces and function params afaict.

src/services/blocks/trace/Trace.ts
emostov: Unless there is an error to the logic, any given event will only belong to one address. So we loop through each address, and for each address we loop through its events. Because an event can only belong to one address we will never see the same event twice. Thus the worst case complexity of these nested loops is equal to the number of events.

src/services/blocks/trace/Trace.ts
emostov: Only one span will fill this. Sorting is actually just to make sure the eventual Operations are in the right order.

src/services/blocks/BlocksTraceService.ts
emostov: Yup, check out the RPC docs: https://github.com/paritytech/substrate/pull/7780/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-a470f7227faf378aa4b8cf63f5b3d35018da37803f6bf0456327e4f6d2e15250R140-R244

src/services/blocks/trace/Trace.ts
emostov: https://github.com/paritytech/substrate-api-sidecar/issues/528

src/services/blocks/trace/Trace.ts
emostov: https://github.com/paritytech/substrate-api-sidecar/issues/528

src/services/blocks/trace/Trace.ts
emostov: I believe so. Reasoning is that anything under `query` is a storage item and all storage items live under some key or a key prefix if it something like a map where the non-prefix part of the storage key is the map key.

emostov: [Here](https://github.com/polkadot-js/api/blob/a1f07e32743eed1c27bf06ca6fb9118afdbfa839/packages/api/src/base/Decorate.ts#L381-L384) you can see `key` expects either 1 or 2 args (depending on if map or double map). Where as keyPrefix simply expects an optional args. (`creator.keyPrefix` as a signature of `keyPrefix: (arg?: any) => Uint8Array;`)

src/services/blocks/trace/Trace.ts
emostov: lmk if you the changes are any better

src/services/blocks/trace/Trace.ts
emostov: I prefer multi line with brackets for consistency and readability.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,674 @@
...
+
+	if (extrinsicIndex.length === 0) {
+		// Since we assume the `spanIds` are for a span from an apply extrinsic action group,
+		// we assume there is always an `:extrinsic_index` event
+		throw new InternalServerError('Expected at least one extrinsic index');
+	}
+
+	if (extrinsicIndex.length > 1) {
+		// If there are multiple reads we check that they are all the same. If they
+		// are this program has some incorrect assumptions and this error should
```
emostov, line 230:
```suggestion
		// are not this program has some incorrect assumptions and this error should
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,674 @@
...
+		// be reported to the maintainers.
+		for (const [i, extIdx] of extrinsicIndex.entries()) {
+			if (i > 0 && !extrinsicIndex[i]?.eq(extIdx)) {
```
emostov, line 233:
```suggestion
			if (i > 0 && !extrinsicIndex[i -1]?.eq(extIdx)) {
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+	target: string;
+	data: Data;
+}
+
+/**
+ * Span with direct descendants
+ */
+export interface SpanWithChildren extends TraceSpan {
+	/**
+	 * Id of the child spans.
```
emostov, line 61:
```suggestion
	 * Ids of the child spans.
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+	spans: TraceSpan[];
+}
+
+export function isBlockTrace(
+	thing: unknown
+): thing is { blockTrace: BlockTrace } {
+	return (
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace.blockHash ===
+			'string' &&
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace.tracingTargets ===
```
emostov, line 79:
```suggestion
		typeof (thing as { blockTrace: BlockTrace })?.blockTrace?.tracingTargets ===
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+	tracingTargets: string;
+	events: TraceEvent[];
+	spans: TraceSpan[];
+}
+
+export function isBlockTrace(
+	thing: unknown
+): thing is { blockTrace: BlockTrace } {
+	return (
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace.blockHash ===
```
emostov, line 77:
```suggestion
		typeof (thing as { blockTrace: BlockTrace })?.blockTrace?.blockHash ===
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+			'string' &&
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace.tracingTargets ===
+			'string' &&
+		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace.events) &&
```
emostov, line 81:
```suggestion
		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace?.events) &&
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace.spans)
```
emostov, line 82:
```suggestion
		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace?.spans)
```

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export interface CurrencyId {
+	symbol: string;
+}
+
+export interface StorageResourceId {
+	pallet: string;
+	item: string;
+	/**
+	 * field1 is nested in item.
```
emostov, line 127:
```suggestion
	 * field1 is nested field in item.
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,691 @@
...
+			} else {
+				acc.get(address)?.push(asAccount);
+			}
+
+			return acc;
+		}, new Map<string, PAccountEvent[]>());
+	}
+
+	/**
+	 * Convert an `PActionEvent` to a `PAccountEvent` by adding the decoded `accountInfo`
```
emostov, line 525:
```suggestion
	 * Convert a `PActionEvent` to a `PAccountEvent` by adding the decoded `accountInfo`
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,691 @@
...
+	 * and `address`.
+	 *
+	 * Notes: will throw if event does not have system::account storagePath.
```
emostov, line 528:
```suggestion
	 * Notes: will throw if event does not have `system` `account` `storagePath`
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,691 @@
...
+	private toAccountEvent(event: PActionEvent): PAccountEvent {
+		const { storagePath } = event;
+		if (
+			!(
+				(storagePath as PalletKeyInfo).module == 'system' &&
+				(storagePath as PalletKeyInfo).item == 'account'
+			)
+		) {
+			throw new InternalServerError(
+				'Event did not have system::account path as expected'
```
emostov, line 541:
```suggestion
				'Event did not have system::account key prefix as expected'
```

TarikGul:
Thanks for the explanation's and the responses to my question's. It definitely cleared up a lot for me, and gave me a way clearer picture of a lot of the code. Let me know when you want me to do another run through of the PR, and I'll review it again.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+					: // This case catches `onFinalize` since `onFinalize` spans are
+					  // identified by the priamry spans name.
+					  stringCamelCase(primary.name);
+			}
+
+			const events =
+				eventsByParentId
+					.get(primary.id)
+					?.concat(secondarySpansEvents)
+					.sort((a, b) => a.eventIndex - b.eventIndex) || [];
```
emostov, line 363: Can probably remove this sort and the sort when parsing spans. Instead we can just sort all the Operations at the end by eventIndex before returning. Although it may be less stable for in terms of Operation index in the case when we get multiple Operations for an event

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+	spansById: SpansById;
+	parsedSpans: SpanWithChildren[];
+	executeBlockSpanId: number;
+} {
+	// IMPORTANT for downstream logic: Make sure spans are sorted.
+	// If this gets taken out, then **all** `Operation`s will need to be sorted on
+	// `eventIndex` to ensure integrity in there ordering. (Ordering integrity
+	// is not strictly neccesary for MVP balance reconciliation, but operating
+	// under the assumption it may be neccesary for closer audits.)
+	spans.sort((a, b) => a.id - b.id);
```
emostov, line 137: Realizing we probably don't need to do this. Instead we can avoid sorting here and sorting the events of an action group if we just sort the operations by eventIndex before returning.

The current strategy is sort spans, so actionGroups are in the order of there primary span. And sort events within an actionGroup so Operations within an actionGroup are in order. We merge together operations in the order of action groups so it ends up all sorted.

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+import {
+	BlocksTrace,
+	BlocksTraceOperations,
+} from '../../types/responses/BlocksTrace';
+import { AbstractService } from '../AbstractService';
+import { BlockTraceResponse, isBlockTrace, isTraceError, Trace } from './trace';
+
+const DEFAULT_TARGETS = 'pallet,frame,state';
+// :extrinsic_index & frame_system::Account
+const DEFAULT_KEYS =
```
niklasad1, line 13: what does `DEFAULT_KEYS` mean?

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,102 @@
...
+	 */
+	async traces(hash: BlockHash): Promise<BlocksTrace> {
+		const [{ number }, traceResponse] = await Promise.all([
+			this.api.rpc.chain.getHeader(hash),
+			// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+			// @ts-ignore
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-call
+			this.api.rpc.state.traceBlock(
+				hash,
+				DEFAULT_TARGETS,
```
niklasad1, line 31: So why is this hardcoded to `DEFAULT_TARGETS` and `DEFAULT_KEYS`? 

Is it possible for users to enter their own log targets and storage keys?

src/services/blocks/trace/types.ts
```diff
@@ -0,0 +1,206 @@
...
+): thing is { blockTrace: BlockTrace } {
+	return (
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace?.blockHash ===
+			'string' &&
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace?.tracingTargets ===
+			'string' &&
+		typeof (thing as { blockTrace: BlockTrace })?.blockTrace?.storageKeys ===
+			'string' &&
+		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace?.events) &&
+		Array.isArray((thing as { blockTrace: BlockTrace })?.blockTrace?.spans)
```
dvdplm, line 85: Repeating `(thing as { blockTrace: BlockTrace })?.blockTrace?` many times seems inefficient. Can't you

```
let bt_thing = (thing as { blockTrace: BlockTrace })?.blockTrace?;
return ['blockHash', 'tracingTargets', 'storageKeys'].all( member => typeof bt_thing[member] === 'string') && Array.isArray(… … …
```
?
(rewritten in actual js ofc)

src/services/test-helpers/mock/data/getKeyNames.ts
```diff
@@ -0,0 +1,747 @@
+export const keyNames = {
```
dvdplm, line 1: What would you need for this list to be automatically generated?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,688 @@
...
+	SpecialKeyInfo,
+	TraceSpan,
+} from './types';
+
+const EMPTY_KEY_INFO = {
+	error: 'key not found',
+};
+
+/**
+ * Known spans
```
dvdplm, line 30: Could use some expansion imo. What they are and why we need them.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,688 @@
...
+type SpansById = Map<number, SpanWithChildren>;
+
+/**
+ * Extrinsic indexs decoded from storage read events, keyed by the span Id to which
+ * the original storage event belonged to.
+ */
+type ExtrinsicIndexBySpanId = Map<number, BN>;
+
+/**
+ * Mapping of span id => array of the events belongin to the span.
```
dvdplm, line 63:
```suggestion
 * Mapping of span id => array of the events belonging to the span.
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,688 @@
...
+				acc[key.toString()] = {
+					module: mod,
+					item,
+				};
+			});
+
+			return acc;
+		}, {} as Record<string, KeyInfo>);
+
+		const wellKnownKeys = {
```
dvdplm, line 132: Couldn't this be a top-level `const`, like `SPANS`?

dvdplm (commented):
I'll be back, my brain hurts.

src/services/blocks/BlocksTraceService.ts
emostov: LMK what you think about the new comment.

docs/src/openapi-v1.yaml
```diff
@@ -921,6 +922,153 @@ paths:
...
+
+        To learn more about operation and action group creation please consult
+        [this diagram](https://docs.google.com/drawings/d/1vZoJo9jaXlz0LmrdTOgHck9_1LsfuQPRmTr-5g1tOis/edit?usp=sharing)
+      parameters:
+      - name: blockId
+        in: path
+        description: Block identifier, as the block height or block hash.
+        required: true
+        schema:
+          pattern: 'a-km-zA-HJ-NP-Z1-9{8,64}'
```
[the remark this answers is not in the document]
emostov: Regex for a hex block hash. I think @danforbes helped me with it.

src/main.ts
```diff
@@ -39,10 +39,41 @@ async function main() {
...
+	const rpc = {
+		state: {
+			traceBlock: {
+				description: 'Block state trace prototype',
+				params: [
+					{
+						name: 'at',
+						type: 'BlockHash',
+						isOptional: true,
+						isHistoric: true,
```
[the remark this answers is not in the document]
emostov: Great question, AFAICT its not covered in the [custom rpc](https://polkadot.js.org/docs/api/start/rpc.custom), but it [looks like](https://github.com/polkadot-js/api/blob/a6fcba95de8a1a6dc8cce1ca2444d43c61cf6db6/packages/rpc-core/src/index.ts#L212-L214) it is so we can do historical `.at(<blockHash>)` queries by specifying the `blockHash`.

See an example in [polkadot-js](https://github.com/polkadot-js/api/blob/a1f07e32743eed1c27bf06ca6fb9118afdbfa839/packages/types/src/interfaces/state/definitions.ts#L24-L26) here

src/services/blocks/BlocksTraceService.ts
emostov: I don't want to link to the `env_logger` docs since we do custom filtering (prefix based), but I will link the RPC docs.

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,127 @@
...
+import { InternalServerError } from 'http-errors';
+
+import {
+	BlocksTrace,
+	BlocksTraceOperations,
+} from '../../types/responses/BlocksTrace';
+import { AbstractService } from '../AbstractService';
+import { BlockTraceResponse, isBlockTrace, isTraceError, Trace } from './trace';
+
+const DEFAULT_TARGETS = 'pallet,frame,state';
```
emostov, line 11: TODO link RPC doc section on targets and key filters.

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,105 @@
...
+	 * @param hash `BlockHash` to get balance transfer operations at.
+	 * @param includeActions whether or not to include `actions` field in the response.
+	 */
+	async operations(
+		hash: BlockHash,
+		includeActions: boolean
+	): Promise<BlocksTraceOperations> {
+		const [{ block }, traceResponse] = await Promise.all([
+			// Note: this should be getHeader, but the type registry on chain_getBlock is the only
+			// one that actually has the historical types. This is a polkadot-js bug
```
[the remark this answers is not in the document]
emostov: No issue AFAIK, will make one.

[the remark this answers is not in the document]
emostov: https://github.com/polkadot-js/api/issues/3487

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,105 @@
...
+			this.api.rpc.chain.getBlock(hash),
+			// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+			// @ts-ignore
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-call
+			this.api.rpc.state.traceBlock(
+				hash,
+				DEFAULT_TARGETS,
+				DEFAULT_KEYS
+			) as Promise<BlockTraceResponse>,
```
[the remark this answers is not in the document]
emostov: Since the `traceBlock` RPC has not been added to polkadot-js we had to put in a custom definition that is created at runtime. Because the def is created at runtime tsc does not seem to have a way to know of that - so it thinks we are calling a function that does not exist.

`Promise<BlockTraceResponse>` => the return type

src/services/test-helpers/mock/data/getKeyNames.ts
emostov: I could generate off the Metadata, but would be a bit more code so I figured for the sake of keeping our sanity for these tests it would be easier to just copy and paste this.

src/services/blocks/trace/Trace.ts
emostov: Removing this sort breaks things even if we sort events for each address in `accountEventsByAddress`, not sure why.

docs/src/openapi-v1.yaml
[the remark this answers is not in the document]
dvdplm: That's unusuale way of expressing a regexp, the syntax is new to me; do you know where the block hash format comes from? It seems to allow block hashes to be 8 chars long? But what if it's a number, say `2`?

It's not important for the PR, I'm just curious and trying to understand.

src/services/blocks/BlocksTraceService.ts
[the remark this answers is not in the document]
dvdplm: lol, ok, I'd never have figured that out on my own, ty!

src/services/test-helpers/mock/data/getKeyNames.ts
dvdplm: Fair enough.

docs/src/openapi-v1.yaml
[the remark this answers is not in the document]
niklasad1: The regex enforces that `a-km-zA-HJ-NP-Z1-9` is matched between 8-64 times looks quite weird to me too.

src/controllers/blocks/BlocksTraceController.ts
```diff
@@ -0,0 +1,66 @@
...
+			['/:number/traces', this.getBlockTraces],
+			['/:number/traces/operations', this.getBlockOperations],
+			['/head/traces/operations', this.getLatestBlockOperations],
+		]);
+	}
+
+	private getLatestBlockTraces: RequestHandler = async (
+		_req,
+		res
+	): Promise<void> => {
```
niklasad1, line 26: dq: how are the error propagated here? is it caught when the user calls `resolve/then` and `catch`?

I got a little confused with the return type `void`

src/services/blocks/BlocksTraceService.ts
niklasad1: yeah or just write a comment like:

> we use the same default tracing targets as substrate, for further tracing targets [link to substrate or something]

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Equivalent of the storage keys for :extrinsic_index & frame_system::Account.
+ * These are the storage keys we use to filter events to reduce payload size and
+ * computational complexity for processing the data. For creating operations we
+ * only need events for storage Get/Put to these two storage items.
+ *
+ * Note: frame_system::Account is the storage prefix for items in a map. In the
+ * storage events we will get the actual entries of the map and use the key suffix
+ * to extract the address.
+ * ref: https://github.com/paritytech/substrate/blob/a604906c340c90e22fb20a8d77bcb3fee86c73c1/frame/system/src/lib.rs#L530-L538
+ * learn about transparen keys: https://www.shawntabrizi.com/substrate/transparent-keys-in-substrate/
```
niklasad1, line 25:
```suggestion
 * learn about transparent keys: https://www.shawntabrizi.com/substrate/transparent-keys-in-substrate/
```

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * These are the keys we pass as an arg to the RPC to filter events. We will get
+ * storage events where they have a key whose prefix matches one of these.
+ *
+ * Equivalent of the storage keys for :extrinsic_index & frame_system::Account.
+ * These are the storage keys we use to filter events to reduce payload size and
+ * computational complexity for processing the data. For creating operations we
+ * only need events for storage Get/Put to these two storage items.
+ *
+ * Note: frame_system::Account is the storage prefix for items in a map. In the
```
niklasad1, line 21: maybe?!

```suggestion
 * Note: frame_system::Account is the storage prefix for items in the storage map. In the
```

src/services/blocks/BlocksTraceService.ts
niklasad1: It's great makes it more clear to me at least

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,127 @@
...
+ * storage events we will get the actual entries of the map and use the key suffix
+ * to extract the address.
+ * ref: https://github.com/paritytech/substrate/blob/a604906c340c90e22fb20a8d77bcb3fee86c73c1/frame/system/src/lib.rs#L530-L538
+ * learn about transparen keys: https://www.shawntabrizi.com/substrate/transparent-keys-in-substrate/
+ *
+ * Note: :extrinisc_index is the key for a single `u32` value that gets updated
+ * during block execution to reflect the index of the current extrinsic being excuted.
```
niklasad1, line 28:
```suggestion
 * during block execution to reflect the index of the current extrinsic being exceuted.
```

src/services/blocks/BlocksTraceService.ts
```diff
@@ -0,0 +1,127 @@
...
+		} else if (isBlockTrace(traceResponse)) {
+			return {
+				at: {
+					hash,
+					height: number.unwrap().toString(10),
+				},
+				storageKeys: traceResponse.blockTrace.storageKeys,
+				tracingTargets: traceResponse.blockTrace.tracingTargets,
+				events: traceResponse.blockTrace.events,
+				spans: traceResponse.blockTrace.spans.sort((a, b) => a.id - b.id),
```
niklasad1, line 71: underflow doesn't matter/can't happen here?

I think I recall that number in JS/TS is float.

src/services/blocks/trace/Trace.ts
niklasad1: it would be super-useful add this in wiki/documentation in the long run.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+	 * @param extrinsicIndexBySpanId Map of spanId => extrinsicIndex
+	 */
+	private static extractExtrinsicIndex(
+		spanIds: number[],
+		extrinsicIndexBySpanId: ExtrinsicIndexBySpanId
+	): BN {
+		// There are typically multiple reads of extrinsic index when applying an extrinsic,
+		// So we get all of those reads, which should all be the same number.
+		const extrinsicIndex = spanIds.reduce((acc, id) => {
+			if (!extrinsicIndexBySpanId.has(id)) {
```
niklasad1, line 170: can't you avoid one lookup by doing `get()` check whether is it undefined or not?

```
const maybe_id = extrinsicIndexBySpanId.get(id);

if maybe_id !== undefined) {
  acc.push(extrinsicIndexBySpanId.get(maybe_id) as BN);
}
return acc;

```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+	 * apply extrinsic action groups should contain read events to `:extrinsic_index`.
+	 *
+	 * @param spanIds List of spanIds for an apply extrinsic action group
+	 * @param extrinsicIndexBySpanId Map of spanId => extrinsicIndex
+	 */
+	private static extractExtrinsicIndex(
+		spanIds: number[],
+		extrinsicIndexBySpanId: ExtrinsicIndexBySpanId
+	): BN {
+		// There are typically multiple reads of extrinsic index when applying an extrinsic,
```
niklasad1, line 167: so I don't really understand why you have to `reduce` to a list when it's an error having multiple intrinsic numbers,

Thus, can't just assign that single variable that you check when iterating over the `spanIds`?

For example:

```
let uniq = undefined;
for (let id of spanIds) {

    if (uniq === undefined) {
       uniq = id;
    } 
    
    if (uniq != id) {
       throw new InternalServerError(
       'Expect extrinsic to only be applied at a single index.'
        );
    }
}
```

src/services/blocks/trace/Trace.ts
niklasad1: right, you might need to iterate over the extrinsic.entries() too.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+		while (stack.length) {
+			const curId = stack.pop();
+			if (!curId) {
+				break;
+			}
+
+			descendants.push(curId);
+
+			const curSpanChildren = spansById.get(curId)?.children;
+			curSpanChildren && stack.push(...curSpanChildren);
```
niklasad1, line 223: this syntax looks weird to me, is this an idiom in TS/JS? 

Personally I found an if, much more readable

```
 if (curSpan) {
    stack.push(...curSpanChildren);
 }
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+
+		// Go through each span, and add its `id` to the `children` array of its
+		// parent span, creating a tree of spans starting with `executeBlock` span as
+		// the root.
+		for (const span of spans) {
+			if (!span.parentId) {
+				continue;
+			}
+
+			if (!spansById.has(span.parentId)) {
```
niklasad1, line 279: same question here if `has() + get()` could be replace with `get()` and check !== undefined?

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+		} = Trace.parseSpans(this.traceBlock.spans);
+		const { eventsByParentId, extrinsicIndexBySpanId } = this.parseEvents(
+			spansById
+		);
+
+		const actionEvents: ParsedActionEvent[] = [];
+		const actions: ActionGroup[] = [];
+		// Create a list of action groups (`actions`) and a list of all `Operations`.
+		for (const primary of spans) {
+			if (!(primary.parentId === executeBlockSpanId)) {
```
niklasad1, line 308: easier to read?

```suggestion
			if (primary.parentId !== executeBlockSpanId) {
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+		const { value_encoded, result_encoded } = event.data.stringValues;
+		const indexEncodedOption = value_encoded ?? result_encoded;
+		if (!(typeof indexEncodedOption == 'string')) {
+			throw new InternalServerError(
+				'Expected there to be an encoded extrinsic index for extrinsic index event'
+			);
+		}
+
+		const scale = indexEncodedOption
+			.slice(2) // Slice of leading `Option` bits beacuse they cause trouble
```
niklasad1, line 423:
```suggestion
			.slice(2) // Slice of leading `Option` bits because they cause trouble
```

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+			const p = spansById.get(event.parentId);
+			const parentSpanId = p && {
+				name: p.name,
+				target: p.target,
+				id: p.id,
+			};
+
+			const parsed = { ...event, storagePath, parentSpanId, eventIndex: idx };
+
+			if (!eventsByParentId.has(parsed.parentId)) {
```
niklasad1, line 465: maybe do get() and either add new list or push to the existing one.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+					symbol: this.registry.chainTokens[0],
+				};
+				// Base information for all `storage` values
+				const systemAccountData = {
+					pallet: 'system',
+					item: 'Account',
+					field1: 'data',
+				};
+
+				if (!free.eqn(0)) {
```
niklasad1, line 621: shouldn't you check `free.gtn(0)`? 

or is a negative number valid here?

niklasad1, line 621: we just filter out balance operations that doesn't have any impact on the state (i.e, 0)?

niklasad1 (commented):
Looks generally according to my mad TS/JS knowledge.  

I added some comments on the use of `map`, where the lookup could be performed more efficient and some dumb/annoying questions.

Nice work.

src/services/blocks/trace/Trace.ts
```diff
@@ -0,0 +1,690 @@
...
+			const address = asAccount.address.toString();
+
+			if (!acc.has(address)) {
+				acc.set(address, [asAccount]);
+			} else {
+				acc.get(address)?.push(asAccount);
+			}
+
+			return acc;
+		}, new Map<string, ParsedAccountEvent[]>());
```
niklasad1, line 516: wouldn't Map<eventIndex, (string, ParsedAccountEvent[])> make more sense here?

Ideally you would insert that in a collection that's sorted by keys such as an MinHeap or something.

src/services/blocks/BlocksTraceService.ts
dvdplm: In "normal" langs the `sort()` function must return one of `-1`, `0`, and `1`, but I think in JS it's alright to return `something negative`, `something zero-ish` and `something positive`. If `a - b` is lower than whatever minimum 56 bits give you it'll be `-Infinity`, which is `something negative`, so it's all good.

I think. :)

docs/src/openapi-v1.yaml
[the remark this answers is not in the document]
emostov: Honestly I can't remember the origin of it. Possible I tried to learn regex and wrote it, but I would consider my regex illiterate so I can't even really tell you what it means. Practically it doesn't do anything and I am in favor of removing

src/controllers/blocks/BlocksTraceController.ts
emostov: Yeah great question actually. I don't know if there is any idiomatic types for return errors with a neat enum like Rust. This service basically relies on all errors being thrown.

The caveat with express v4 `RequestHandler` functions is that they do not catch thrown errors in `async` functions. So we end up wrapping the call in `try { .. } catch { .. }` blocks. This effectively happens in `AbstrctController.safeMountAsyncGetHandlers`. Specifically it happens in [catchWrap](https://github.com/paritytech/substrate-api-sidecar/blob/01ae7ac1cbcb06cd76d6895bf9ccbe98e99f0d43/src/controllers/AbstractController.ts#L73-L90)

src/services/blocks/BlocksTraceService.ts
emostov: If compareFunction(a, b) returns less than 0, leave a and b unchanged.
If compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements. Note: the ECMAScript standard only started guaranteeing this behavior in 2019, thus, older browsers may not respect this.
If compareFunction(a, b) returns greater than 0, sort b before a.
compareFunction(a, b) must always return the same value when given a specific pair of elements a and b as its two arguments. If inconsistent results are returned, then the sort order is undefined.

from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description

emostov: @niklasad1 tl;dr plain JS code typically has a way to just keep going until it absolutely can't

src/services/blocks/trace/Trace.ts
emostov: Nice, yeah that is a good point. Re-wrote it so its still a reduce, but now we just check the value is either undefined or matches the previous value

src/services/blocks/trace/Trace.ts
emostov: Yeah, basically if nothing changes it probably meant there where back to back reads (normal can happen in block set up and then again prior to extrinsic execution). But yes in cases where there is a put and get we effectively don't count it. I should maybe make a note of that somewhere because it could affect auditing.

src/services/blocks/trace/Trace.ts
emostov: Hmm so basically there is only ever one event with a given index, so this would just map to one event. 

I do think there is probably some room for complexity improvement in regards to sorting events. They start out sorted, but then when we create the action groups, we annotate the events with phase info and essentially create a new collection. 
```javascript
			const primarySpanEvents = eventsByParentId.get(primary.id);
			const events = primarySpanEvents?.concat(secondarySpansEvents) || [];

			events.forEach((e) => {
				actionEvents.push({
					...e,
					phase: {
						variant: phase,
						extrinsicIndex: extrinsicIndex,
					},
					primarySpanId: {
						id: primary.id,
						name: primary.name,
						target: primary.target,
					},
				});
			});
```

I suppose we could though just try out mutating the original events in place... will need to try that out to see if it makes ses

emostov: So I think this turned into a significant improvement. We removed nlogn complexity by removing the sort we used to have to do right before creating Operations. I get around this by just always mutating the original event object. Since the events start out ordered, we just have to not mess up the ordering when creating the array of events for each address in `Map<string, ParsedAccountEvent[]>()`.

I do think the mutations may make future debugging more difficult, but seems to reduce reasonably avoidable time & space complexity.

dvdplm (approved):
Good enough for an mvp imo. Let's merge it.

Issue #385, opened 2021-01-13
partialFee value is always 0 on local development with polkadot 1.8.27


islishude:
**Description**
nodejs version: v14 LTS
polkadot version: 1.8.27
sidecar version: latest commit

**Steps to Reproduce**

make a transfer at polkadot.js.org/app,and request block info

![截屏2021-01-13 下午6 09 40](https://user-images.githubusercontent.com/24730006/104438100-923a9680-55ca-11eb-9c22-30b8720a0fab.png)

polkadot 1.8.26-1 is ok

**Expected vs. Actual Behavior**

[A clear description of what you expected to happen and what actually happened.]

emostov:
rel: https://github.com/paritytech/substrate/pull/6629

emostov:
`ExtrinsicBaseWeight` is no longer a constant in the system pallet and the fee calculations need to be updated.

vae520283995:
I have the same problem

vae520283995:
> I have the same problem

Fortunately, this is the testnet. If it is the mainnet, the consequences will be disastrous.

PR #388, opened 2021-01-14
fix: Update fee calc to use `system::constants::BlockWeights.per_class.normal.base_extrinsic`

emostov:
closes: #385 

### For review discussion
Should we take into account `DispatchClass` when using `baseExtrinsic`? At the moment polkadot runtimes have the `baseExtrinsic` for each class is fixed: https://github.com/paritytech/polkadot/blob/cd0e9186a6620aab846a8a54969a0e5bbecbd060/runtime/common/src/lib.rs#L88

From my understanding the changes would roughly require a refactor of the `calc_fee.rs` logic to do something like take in the dispatch class for each `calc_fee()` call and pass a mapping of the `base_extrinsic` weight for each dispatch class when creating `CalcFee.from_params()`. 

My two cents is that we should ship this fix ASAP, and then get the more thorough fix out next week.


src/services/blocks/BlocksService.ts
```diff
@@ -340,13 +340,14 @@ export class BlocksService extends AbstractService {
 		block: Block
 	) {
 		const perByte = api.consts.transactionPayment?.transactionByteFee;
-		const extrinsicBaseWeight = api.consts.system
-			?.extrinsicBaseWeight as AbstractInt;
+		const extrinsicBaseWeightExists =
+			api.consts.system?.extrinsicBaseWeight ||
+			api.query.system.blockWeight;
```
athei, line 345:
```suggestion
			api.query.system.blockWeights;
```

src/services/blocks/BlocksService.ts
```diff
@@ -385,10 +386,31 @@ export class BlocksService extends AbstractService {
...
+					multiplier,
+					{
+						normal: { baseExtrinsic: extrinsicBaseWeight },
+					},
+				] = await Promise.all([
+					api.rpc.state.getRuntimeVersion(parentParentHash),
+					api.query.transactionPayment.nextFeeMultiplier.at(
+						parentHash
+					),
+					api.query.system.blockWeight.at(parentHash),
```
athei, line 411:
```suggestion
					api.query.system.blockWeights.at(parentHash),
```

athei (commented):
Please keep in mind that this only works for non-normal transactions if they share the same base extrinsic weight.

src/services/blocks/BlocksService.ts
```diff
@@ -340,8 +340,10 @@ export class BlocksService extends AbstractService {
 		block: Block
 	) {
 		const perByte = api.consts.transactionPayment?.transactionByteFee;
-		const extrinsicBaseWeight = api.consts.system
-			?.extrinsicBaseWeight as AbstractInt;
+		const extrinsicBaseWeight = api.consts.system.extrinsicBaseWeight
+			? (api.consts.system.extrinsicBaseWeight as AbstractInt)
+			: // We assume all dispatch classes have the same base_extrinsic value
+			  api.consts.system.blockWeights.perClass.normal.baseExtrinsic;
```
kianenigma, line 346: so the js api has the value `api.consts.system.extrinsicBaseWeight` set before the weight refactor and for a chain that has it, the latter one?

kianenigma:
Sensible to do as you said: 

> My two cents is that we should ship this fix ASAP, and then get the more thorough fix out next week.

dvdplm (approved):
LGTM; we'll revisit this more thoroughly in subsequent releases.

emostov:
@TarikGul , @dvdplm , @kianenigma - I just re-did the whole thing. I forgot that the api just always uses the latests metadata. If we are doing a query in a historical runtime we need to fetch the old metadata, decorate it, and grab the constant. If we are doing a query in the latest runtime we can save the effort and just use the decorated metadata built into the api. The logic is worth another look over

src/services/blocks/BlocksService.ts
```diff
@@ -338,13 +341,14 @@ export class BlocksService extends AbstractService {
 		block: Block
 	) {
 		const perByte = api.consts.transactionPayment?.transactionByteFee;
-		const extrinsicBaseWeight = api.consts.system
-			?.extrinsicBaseWeight as AbstractInt;
+		const extrinsicBaseWeightExists =
+			api.consts.system.extrinsicBaseWeight ||
+			api.consts.system.blockWeights.perClass.normal.baseExtrinsic;
```
emostov, line 346: We are making an assumption here that if a runtime has one of these in the metadata currently, all past runtimes will have it. Since we only support fee calcs on a limited number of chains I think this is ok for now

dvdplm:
> The logic is worth another look over

Still LGTM.

Issue #333, opened 2020-10-29
Look into publishing Docker image with releases


emostov:
**Proposed Change or Idea**

Automatically publish a docker image for every release. Currently I do not know of any up to date public images.

This would reduce the barrier for some users to run the service. (See #318)

PR #396, opened 2021-01-18
ci: build container on release

TriplEight:
- optimized Dockerfile; 
- labelled Dockerfile; 
- Github Action to build container; 
- edit docker-compose

Closes: paritytech/ci_cd/issues/51
Closes: paritytech/substrate-api-sidecar#333

### DIFF
--- a/.github/workflows/publish-release-container.yml
+++ b/.github/workflows/publish-release-container.yml
@@ -1,0 +1,42 @@
+name: Build and Publish container for new releases
+
+on: [push] # debug
+  # release:
+  #   types:
+  #     - published
+
+jobs:
+  build:
+    name: Build and push image
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_NAME: docker.io/parity/substrate-api-sidecar
+      IMAGE_TAG: latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build Image
+      uses: redhat-actions/buildah-build@v1
+      with:
+        image: ${{ env.IMAGE_NAME }}
+        tag: ${{ env.IMAGE_TAG }}
+        dockerfiles: |
+          ./Dockerfile
+        build-args: |
+          VERSION=${{ github.event.release.tag_name }}
+          VCS_REF=${{ github.ref }}
+          BUILD_DATE=${{ github.event.release.published_at }}
+
+    - name: Push To docker.io
+      id: push-to-dockerhub
+      uses: redhat-actions/push-to-registry@v1
+      with:
+        image: ${{ env.IMAGE_NAME }}
+        tag: ${{ env.IMAGE_TAG }}
+        registry: docker.io/parity
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_TOKEN }}
+
+    - name: Check the image
+      run: echo "New image has been pushed to ${{ steps.push-to-dockerhub.outputs.registry-path }}"
--- a/.github/workflows/publish-release-container.yml
+++ b/.github/workflows/publish-release-container.yml
@@ -19,7 +19,6 @@ jobs:
     env:
       REGISTRY:                   docker.io/parity
       IMAGE_NAME:                 substrate-api-sidecar
-      # IMAGE_TAG:                  ${{ matrix.tag }}
       VERSION:                    ${{ github.event.release.tag_name }}
 
     steps:
--- a/.github/workflows/publish-release-container.yml
+++ b/.github/workflows/publish-release-container.yml
@@ -9,20 +9,25 @@ jobs:
   build:
     name:                         Build and push image
     runs-on:                      ubuntu-latest
+    strategy:
+      matrix:
+        # tag: ['latest', '${{ github.event.release.tag_name }}']
+        tag: ['latest', '${{ GITHUB_REF }}']
     env:
       REGISTRY:                   docker.io/paritytech
       IMAGE_NAME:                 substrate-api-sidecar
-      IMAGE_TAG:                  latest
-      VERSION:                    ${{ github.event.release.tag_name }}
+      # IMAGE_TAG:                  ${{ matrix.tag }}
+      # VERSION:                    ${{ github.event.release.tag_name }}
+      VERSION:                    ${{ GITHUB_REF }}
 
     steps:
     - uses:                       actions/checkout@v2
 
-    - name:                       Build Image tagged ${{ env.IMAGE_TAG }}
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,42 @@
+name: Build and Publish container for new releases
+
+on: [push] # debug
```
emostov, line 3: Does this build and publish the image when I create a release through github ui?

emostov (approved):
Thanks @TriplEight !! I don't fully understand all the pieces so worth a look over by a set of eyes with more workflows/docker experience

.github/workflows/publish-release-container.yml
TriplEight: yup, the commented lines below do exactly this

TriplEight:
Sorry, it had to take so long to get here.

Dockerfile
```diff
@@ -1,18 +1,35 @@
-FROM node:15 as builder
-LABEL author="chevdor@gmail.com"
```
lovelaced, line 2: We should only add authors/maintainers, not remove them. ref:

https://github.com/paritytech/substrate-telemetry/pull/269#discussion_r455126513

lovelaced (changes requested):
lgtm otherwise!

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,59 @@
...
+  # release:
+  #   types:
+  #     - published
+
+jobs:
+  build:
+    name:                         Build and push image
+    runs-on:                      ubuntu-latest
+    env:
+      REGISTRY:                   docker.io/paritytech
```
lovelaced, line 13: hmm, actually, don't we want to release this under `parity` and *not* `paritytech`? I think this is intended to be something that other people are going to be using.

Dockerfile
TriplEight: :+1:

.github/workflows/publish-release-container.yml
TriplEight: It was decided with @dvdplm that for the first time it will live under `paritytech`.

TriplEight:
Now it can be merged.

this is a bit tricky with that 
- `buildah` build and push actions do not support more than one tag
- github actions maybe support expressions in `matrix`
- this expression can be checked only on release

So please ping me if this action will crash on release.

.github/workflows/publish-release-container.yml
dvdplm: I can be swayed either way though. Superficially it seems confusing to use two different collections of docker images, at least if they are labelled so confusingly similar. Sidecar **is** stable and it **is** an official product; it obviously does not have the same importance as polkadot.

lovelaced: Well, the rule of thumb is that images under the `parity` namespace are generally to be for end-users, and `paritytech` is for in-house CI images, tooling, automation and stuff like that. For example, the telemetry backend and frontend are also both in the `parity` namespace. I would recommend we put this in `parity` too.

TriplEight: Oh, I thought sidecar is still in beta. Then it totally makes sense to publish it as a product, in **parity** namespace.

lovelaced: I'm not sure it really matters if it's in beta or not; the idea is that it's intended to be used externally, and I think changing the namespace down the road is a bad idea, because we have no idea who is going to use it where to notify them of the change.

TriplEight: done

Dockerfile
```diff
@@ -1,18 +1,34 @@
...
-FROM node:15-alpine
+FROM docker.io/library/node:15-alpine
+
+# metadata
+ARG VERSION=""
+ARG VCS_REF=master
+ARG BUILD_DATE=""
+
+LABEL summary="Substrate-api-sidecar." \
+	name="paritytech/substrate-api-sidecar" \
```
lovelaced, line 20: Will this label need changing? Is it a reference to the repo or the docker namespace?

lovelaced (approved):

Dockerfile
TriplEight: oh thanks

.github/workflows/publish-release-container.yml
dvdplm: > Oh, I thought sidecar is still in beta. Then it totally makes sense to publish it as a product, in **parity** namespace.

To give you an idea: sidecar has been used in production at exchanges for ~1 year now and some, like Huobi, has it as a pre-requisite for listing new tokens. It's legit. :)

.github/workflows/publish-release-container.yml
```diff
@@ -0,0 +1,52 @@
...
+      matrix:
+        # this matrix runs the jobs twice, in parallel. This is due to
+        # both buildah actions can't manage more than one image tag.
+        tag:
+          - latest
+          - ${{ github.event.release.tag_name }}
+    env:
+      REGISTRY:                   docker.io/parity
+      IMAGE_NAME:                 substrate-api-sidecar
+      # IMAGE_TAG:                  ${{ matrix.tag }}
```
s3krit, line 22: Is this supposed to be commented out?

Dockerfile
```diff
@@ -1,18 +1,34 @@
...
-    cargo install wasm-pack && \
-    yarn install && \
+
+RUN yarn install && \
     yarn build
 
 # ---------------------------------
 
-FROM node:15-alpine
+FROM docker.io/library/node:15-alpine
```
s3krit, line 12: Am I right in thinking `/library/` is equivalent to `/_/` on hub.docker.com?

s3krit (approved):
Looks good! May buildah take over the world!!😎

.github/workflows/publish-release-container.yml
TriplEight: I left it as an explanatory comment

Dockerfile
TriplEight: yes, `/_/` is the alias for `/library/`

TriplEight: not sure about now, but earlier versions of buildah didn't have the docker registry configured by default and it was confusing. Buildah did not understand the docker-style shortened image addresses i.e. `node:15` or `parity/substrate-api-sidecar` and demanded a fqdn everywhere.
Then it softened and now buildah is able to deduct the image address (when you build manually with just `node:15` in `FROM` it will stop and ask where to take it from - QUAY or dockehub), but still it will throw an error on `push` if you built an image without the registry address in name.

Issue #387, opened 2021-01-13
Can I configure custom types via environment variables?


ntduan:
If I start service from the source code, I can modify the config/types file to add custom types, but if I want to start service from the command line, I don't find a way to configure custom types.

emostov:
At the moment the only way to modify the custom types is to go into `config/types.json`, which technically you could do by going to where the NPM package is installed and modifying it there. However that obviously is not ideal and would be overridden any time you go to update the package. 

So this feature sounds like a good idea to me.

@ntduan do you think an ok would alternative would be to have an environment variable like `SAS_SUBSTRATE_TYPES="path/to/my-custom-types.json"?

ntduan:
> At the moment the only way to modify the custom types is to go into `config/types.json`, which technically you could do by going to where the NPM package is installed and modifying it there. However that obviously is not ideal and would be overridden any time you go to update the package.
> 
> So this feature sounds like a good idea to me.
> 
> @ntduan do you think an ok would alternative would be to have an environment variable like `SAS_SUBSTRATE_TYPES="path/to/my-custom-types.json"?

I'm not quite sure what it should look like. Since the types obtained from `@polkadot/app-config` may not be up to date, I would like to be able to update the types myself. And it would be nice to be able to support typeBundle in the following form.
```
{
    minmax: [0, 499] ,
    type: {
      ...
    }
  },
  {
    minmax: [500, 599] ,
    type: {
      ...
    }
  },
 {
    minmax: [599, undefined] ,
    type: {
      ...
    }
  },
```

emostov:
Ok, in that case I think a solution that simultaneously supports dynamically specifying `typeBundle`,  `typeSpec`, `typesChain`, and `types` would be ideal. Essentially we want to allow a api-sidecar user to have the ability to have the ability to override any of these fields:

https://github.com/paritytech/substrate-api-sidecar/blob/9d89309bb27e5a7bd47d287d9dbb76e725dd5abf/src/main.ts#L38-L46

emostov:
On the topic of keeping `polkadot/apps-config` up to date, one of the TODOs for sidecar is to automate publishing so whenever any polkadot-js dependency updates we automatically publish a new patch release of api-sidecar (see https://github.com/paritytech/substrate-api-sidecar/issues/380)

emostov:
@ntduan let me know wether or not #399 looks like a solution that should work for your use case.

ntduan:
@emostov  I think it looks good

PR #399, opened 2021-01-19
feat!: Type definition specification with env vars and JSON files

emostov:
closes: #387 

## changes

BREAKING CHANGE remove /config/types.json; Functionality replaced with SAS_SUBSTRATE_TYPES.

Additionally the following environment variables are added: 
- SAS_SUBSTRATE_TYPES_BUNDLE
- SAS_SUBSTRATE_TYPES_CHAIN
- SAS_SUBSTRATE_TYPES_SPEC

## motivation

This will allow users to update chain types when there is a delay for type updates in the apps-config package supply chain or if a chain's type definitions simply do not exist in the apps-config.


README.md
```diff
@@ -142,20 +142,45 @@ For more information on our configuration manager visit its readme [here](https:
...
+Sidecar affords environment variables which allow the user to specify an absolute path to a JSON file
+that contains type definitions in the corresponding formats. Consult polkadot-js/api for more info on
+the type formats (see `RegisteredTypes`).
+
+**N.B** Types set from environment variables will override the corresponding types pulled from
+@polkadot/apps-config.
+
+-   `SAS_SUBSTRATE_TYPES_BUNDLE`: a bundle of types with versioning info, type aliases, derives, and
+    rpc definitions. Format: `OverrideBundleType` (see `typesBundle`).
+-   `SAS_SUBSTRATE_TYPES_CHAIN`: type definitions keyed by the `chainName`. Format: `Record<string, RegistryTypes>` (see `typesChain`).
```
dvdplm, line 158: Maybe these "(see `thing`)" could be links?

dvdplm (approved):
This seems to get the job done.

Do you think it'd be good long term to switch to using a proper config file, rather than using env variables? So we'd have `substrate-api-sidecar --config=my-config.[json|toml|yaml]` and have all the type overrides and other configuration as keys in there?

src/main.ts
```diff
@@ -16,8 +17,9 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { ApiPromise } from '@polkadot/api';
-import { typesBundle, typesChain, typesSpec } from '@polkadot/apps-config/api';
+import * as apps from '@polkadot/apps-config/api';
```
TarikGul, line 20: Whats the reason for importing the whole path here? I definitely think what you did looks a lot cleaner and way more readable.

TarikGul (approved):
LGTM. I like how it gives the end user more flexibility.

emostov:
> Do you think it'd be good long term to switch to using a proper config file, rather than using env variables? So we'd have substrate-api-sidecar --config=my-config.[json|toml|yaml] and have all the type overrides and other configuration as keys in there?

@dvdplm Right now the equivalent would be putting all your env vars in a `.env` file, like .`env.polkadot`. Then the user would start up sidecar with `NODE_ENV=polkadot yarn start`. Do you think a yml type config file would be an improvement to this?

dvdplm:
> @dvdplm Right now the equivalent would be putting all your env vars in a `.env` file, like .`env.polkadot`. Then the user would start up sidecar with `NODE_ENV=polkadot yarn start`. Do you think a yml type config file would be an improvement to this?

I do have a distaste for ENV vars, but my opinion on this is definitely a minority. My gut reaction to `.env` files is "well, why not use a proper config file then?". I think env vars are opaque to users, fiddly to set, even fiddlier to unset, tricky to document properly, no namespaces so they grow really long and fugly. 
I think using a good config file format, like TOML (and to a lesser extent YAML) or even `.ini` has a psychological impact on developers to document and test their configs.

But yeah, ENV vars are really popular with many smart people so I tend to keep my opinions to myself. They're good from the devops POV for sure, and has good tooling on that end.

When I wrote my first comment what I had in mind was the ability to merge configs from several places: a decent set of defaults from substrate, refined further from polkadot and finally overrideable by users for their own chains. Such a hierarchy of config values is tractable using config files in a sane format; with ENV vars not so much?

emostov:
> > @dvdplm Right now the equivalent would be putting all your env vars in a `.env` file, like .`env.polkadot`. Then the user would start up sidecar with `NODE_ENV=polkadot yarn start`. Do you think a yml type config file would be an improvement to this?
> 
> I do have a distaste for ENV vars, but my opinion on this is definitely a minority. My gut reaction to `.env` files is "well, why not use a proper config file then?". I think env vars are opaque to users, fiddly to set, even fiddlier to unset, tricky to document properly, no namespaces so they grow really long and fugly.
> I think using a good config file format, like TOML (and to a lesser extent YAML) or even `.ini` has a psychological impact on developers to document and test their configs.
> 
> But yeah, ENV vars are really popular with many smart people so I tend to keep my opinions to myself. They're good from the devops POV for sure, and has good tooling on that end.
> 
> When I wrote my first comment what I had in mind was the ability to merge configs from several places: a decent set of defaults from substrate, refined further from polkadot and finally overrideable by users for their own chains. Such a hierarchy of config values is tractable using config files in a sane format; with ENV vars not so much?

I really have minimal opinion on this, just what ever works for the most people. Happy to explore/implement different solutions in the future - I just would need a clear direction

src/main.ts
emostov: @TarikGul Just to clarify the difference between the keys with those names and the imports with those same names

Issue #382, opened 2021-01-07
Add get extrinsic by id API


gutenye:
**Proposed Change or Idea**

Right now to get extrinsic data, you have to first get the block, then get extrinsic in that block by index.

It would be very helpful if you can provide a simple endpoint, which can be something like `GET /transactions/<blockHeight>-<extrinsicIndex>`, returns extrinsic data. without the need to fetch the whole block data.

emostov:
This will be `/blocks/{blockId}/extrinsics/{extrinsicIndex}` as a collection of `extrinsics `belong-to` a block.

PR #400, opened 2021-01-21
feat: add route /blocks/{blockId}/extrinsics/{extrinsicIndex} 

TarikGul:
## Feat

Added the route `/blocks/{blockId}/extrinsics/{extrinsicsIndex}`

## Changelog 

`src/controllers/blocks/BlocksExtrinsicsController.ts`
Added: `BlocksExtrinsicsController`. This controller uses the `BlocksService` in order to fetch the block, then retrieve the extrinsic. 

`src/services/blocks/BlocksService.ts`
Added: `fetchExtrinsicsByIndex`. Fetch's the extrinsic that is requested in the params. 

`src/services/blocks/BlocksService.spec.ts`
Added a test 'describe'  suite a called `fetchExrinsicsByIndex`.

Additionals:
Added mock data in: `src/services/test-helpers/responses/blocks/blocks789629Extrinsic.json`

## Notes

Inside of the test suite, I had to reuse one line of code between 2 'it' statements. This is because jest was giving me issues with using async in a describe block. It treats it as a nested asynchronous function which will throw a unhandled promise, and abort the test or give inaccurate results. 

closes [#382 ](https://github.com/paritytech/substrate-api-sidecar/issues/382)


docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,56 @@ paths:
...
+      - name: blockId
+        in: path
+        description: Block identifier, as the block height or block hash.
+        required: true
+        schema:
+          pattern: a-km-zA-HJ-NP-Z1-9{8,64}
+          type: string
+      - name: extrinsicsIndex
+        in: path
+        description: The index in which the requested extrinsic exists.
```
emostov, line 293:
```suggestion
        description: The extrinsic's index within the block's body.
```

docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,56 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /blocks/{blockId}/extrinsics/{extrinsicsIndex}:
+    get:
+      tags:
+      - blocks
+      summary: Get an extrinsic by its extrinsicIndex and block height or hash.
```
emostov, line 280:
```suggestion
      summary: Get an extrinsic by its `Timepoint` (blockId, extrinsicIndex).
```
@danforbes @joepetrowski thoughts?

docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,56 @@ paths:
...
+      description: Returns a single extrinsic.
+      operationId: getExtrinsicsByExtrsinsicIndex
```
emostov, line 282:
```suggestion
      operationId: getExtrinsicByTimepoint
```
If we decide to use the `Timepoint` language

docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,56 @@ paths:
...
+      parameters:
+      - name: blockId
+        in: path
+        description: Block identifier, as the block height or block hash.
+        required: true
+        schema:
+          pattern: a-km-zA-HJ-NP-Z1-9{8,64}
+          type: string
+      - name: extrinsicsIndex
```
emostov, line 291:
```suggestion
      - name: extrinsicIndex
```

docs/src/openapi-v1.yaml
```diff
@@ -1033,6 +1083,15 @@ components:
...
             extrinsic is signed when reconciling old blocks.
+    ExtrinsicIndex:
+      type: object
+      properties:
+        at:
+          $ref: '#/components/schemas/BlockIdentifiers'
+        extrinsic:
+          $ref: '#/components/schemas/Extrinsic'
+      description: Object returned after keying into the correct extrinsic, 
+        and adding its blockHash, and block height in a `at` field.
```
emostov, line 1094:
```suggestion
      description: A single extrinsic at a given block.
```

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export default class BlocksExtrinsicsController extends AbstractController<BlocksService> {
+	constructor(api: ApiPromise, private readonly options: ControllerOptions) {
+		super(api, '/blocks/:blockId/extrinsics', new BlocksService(api));
+		this.initRoutes();
+	}
+
+	protected initRoutes(): void {
+		this.safeMountAsyncGetHandlers([
+			['/:extrinsicsIndex', this.getExtrinsicsByExtrsinsicIndex],
```
emostov, line 20: Should we change all mentions of `extrinsicsIndex` to `extriniscIndex` since we only get a single extrinsic?

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+		]);
+	}
+
+	/**
+	 *
+	 * @param _req Express Request
+	 * @param res Express Response
+	 */
+	private getExtrinsicsByExtrsinsicIndex: RequestHandler<INumberParam> = async (
```
emostov, line 29:
```suggestion
	private getExtrinsicByTimepoint: RequestHandler<INumberParam> = async (
```
?

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+			checkFinalized: true,
+			queryFinalizedHead,
+			omitFinalizedTag,
+		};
+
+		const block = await this.service.fetchBlock(hash, options);
+
+		this.parseNumberOrThrow(
+			extrinsicsIndex,
+			'ExstrinsicIndex is not a number'
```
emostov, line 56:
```suggestion
			'`exstrinsicIndex` uri param is not a number'
```

src/services/blocks/BlocksService.ts
```diff
@@ -249,6 +250,30 @@ export class BlocksService extends AbstractService {
 		};
 	}
 
+	/**
+	 *
+	 * @param block Takes in a block which is the result of fetchBlock
```
emostov, line 255:
```suggestion
	 * @param block Takes in a block which is the result of `BlocksService.fetchBlock`
```

src/services/blocks/BlocksService.ts
```diff
@@ -249,6 +250,30 @@ export class BlocksService extends AbstractService {
...
+	 * @param extrinsicsIndex Parameter passed into the request
+	 */
+	fetchExtrinsicsByIndex(
+		block: IBlock,
+		extrinsicsIndex: number
+	): IExtrinsicIndex {
+		if (extrinsicsIndex > block.extrinsics.length - 1) {
+			throw new BadRequest('Requested ExtrinsicIndex does not exist');
```
emostov, line 263:
```suggestion
			throw new BadRequest('Requested `extrinsicIndex` does not exist');
```

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+		);
```
emostov, line 57: We could add a test for `extrinsicIndex < 0`

src/types/responses/Extrinsic.ts
```diff
@@ -29,3 +30,11 @@ export interface ISignature {
 	signature: EcdsaSignature | Ed25519Signature | Sr25519Signature;
 	signer: Address;
 }
+
+export interface IExtrinsicIndex {
+	at: {
```
emostov, line 35: Should be able to use `IAt` interface here. Because of some legacy code height is a string, so will need to unwrap if its a compact and then `.toString(10)` to ensure it serializes as a base 10.

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const queryFinalizedHead = !this.options.finalizes ? false : true;
+		const omitFinalizedTag = !this.options.finalizes ? true : false;
+
+		const options = {
+			eventDocs: eventDocsArg,
+			extrinsicDocs: extrinsicDocsArg,
+			checkFinalized: true,
+			queryFinalizedHead,
+			omitFinalizedTag,
```
emostov, line 49: I think we should hardcode `omitFinalizedTag` too `true` in order to avoid any of the `isFinalizedBlock` logic. Then both `checkFinalized` and `queryFinalizedHead` can be whatever since (I think) they are not used.

emostov (commented):
LGTM modulo some nits mostly on just on naming/ docs.

docs/src/openapi-v1.yaml
dvdplm: Is "Timepoint" and established term in our documentation? If not I think it sounds a bit funny. I found the [`pallet_multisig::Timepoint`](https://substrate.dev/rustdocs/v2.0.0/pallet_multisig/struct.Timepoint.html) but not sure how widely used it is.

docs/src/openapi-v1.yaml
```diff
@@ -1604,4 +1663,4 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/Transaction'
-      required: true
+      required: true
```
dvdplm, line 1666:
```suggestion
      required: true
      
```

src/controllers/blocks/BlocksExtrinsicsController.ts
dvdplm: I think so yes; one could argue that it is one index (singular) within a collection of extrinsics and so `extrinsicsIndex` isn't wrong. To my ears `extrinsicIndex` sounds better though.

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,68 @@
...
+		const block = await this.service.fetchBlock(hash, options);
+
+		this.parseNumberOrThrow(
+			extrinsicsIndex,
+			'ExstrinsicIndex is not a number'
+		);
+
+		// Change extrinsicIndex from a type string to a number before passing it
+		// into any service.
+		const index = parseInt(extrinsicsIndex, 10);
```
dvdplm, line 61: We should probably check for `NaN` here and bail if so?

src/services/blocks/BlocksService.spec.ts
```diff
@@ -358,4 +360,41 @@ describe('BlocksService', () => {
...
+			omitFinalizedTag: false,
+		};
+
+		it('Returns the correct extrinisics object for block 789629', async () => {
+			const block = await blocksService.fetchBlock(
+				blockHash789629,
+				options
+			);
+
+			const extrinsic = blocksService['fetchExtrinsicsByIndex'](block, 0);
```
dvdplm, line 380: Would it be better to fetch a non-zero index to prove that we can fetch xts out of order?

src/services/blocks/BlocksService.ts
```diff
@@ -249,6 +250,30 @@ export class BlocksService extends AbstractService {
 		};
 	}
 
+	/**
+	 *
+	 * @param block Takes in a block which is the result of fetchBlock
+	 * @param extrinsicsIndex Parameter passed into the request
+	 */
+	fetchExtrinsicsByIndex(
```
dvdplm, line 258: This should be `fetchExtrinsicByIndex` (singular).

src/services/test-helpers/responses/blocks/block789629Extrinsic.json
```diff
@@ -0,0 +1,37 @@
...
+                        "class": "Mandatory",
+                        "paysFee": "Yes"
+                    }
+                ]
+            }
+        ],
+        "success": true,
+        "paysFee": false
+    }
+}
```
dvdplm, line 37: Most editors can be configured to ensure there is a newline ending each file.

dvdplm (changes requested):
Makes sense to me. I think we should switch to use singular form (`extrinsicIndex` etc), otherwise lgtm.

joepetrowski:
> I think we should switch to use singular form (extrinsicIndex etc)

I agree

docs/src/openapi-v1.yaml
TarikGul: Yea `Timepoint` does sound a little funny, but that being said a lot of terminology can be weird sometimes, so if this is the correct verbiage I agree with it.

src/controllers/blocks/BlocksExtrinsicsController.ts
TarikGul: Yea I agree with you both. I actually had to say it out loud a few times but `extrinsicIndex` definitely makes it sound way clearer.

src/controllers/blocks/BlocksExtrinsicsController.ts
TarikGul: Yea I totally agree there is actually a function inside the `AbstractController`  called `parseNumberOrThrow` which in essence takes care of this issue. Thanks for reminding me. :)

src/services/blocks/BlocksService.spec.ts
TarikGul: Yea do you think adding another test case would suite useful here, or just changing to a non-zero index?

src/services/test-helpers/responses/blocks/block789629Extrinsic.json
TarikGul: Interesting, i thought the linter would catch that. I'll make sure to configure my editor as well.

src/types/responses/Extrinsic.ts
TarikGul: Thanks for pointing that out. The beauty of `unwrap`

docs/src/openapi-v1.yaml
joepetrowski: Yeah, only usage is in Multisig. I think we can also use this to unify the concept. "Get an extrinsic by its `extrinsicIndex` and block height or hash. The pair `blockId, extrinsicIndex` is sometimes referred to as a `Timepoint`."
https://github.com/paritytech/substrate/blob/d1923a4bc1eded37e57062f8b501c646e5321221/frame/multisig/src/lib.rs#L105

docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,56 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /blocks/{blockId}/extrinsics/{extrinsicsIndex}:
```
joepetrowski, line 276:
```suggestion
  /blocks/{blockId}/extrinsics/{extrinsicIndex}:
```

src/services/blocks/BlocksService.spec.ts
TarikGul: After a quick thought about this one, I will just change the index for this instead of adding an extra test that in essence will do the same exact thing. I like the insight though, and Ill make sure to comment that in the test itself.

src/controllers/blocks/BlocksExtrinsicsController.ts
TarikGul: Yea I agree. `omitFinalizedTag` and `queryFinalizedHead`  are not connected to the output of this route, so we definitely don't need to assume the `this.options` might be fulfilled.

src/services/test-helpers/responses/blocks/block789629Extrinsic.json
dvdplm: I actually don't know why it's being signaled by github. Like why is not having a final LF a concern anyway? But there it is and my eye is drawn to "red thingies on screen", so might as well obey. :)

dvdplm (approved):

docs/src/openapi-v1.yaml
TarikGul: @joepetrowski So would you think keeping the same name `getExtrinsicsByExtrsinsicIndex` for the new API route is the way to go, or changing it too `getExtrinsicsByTimepoint`?

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -0,0 +1,66 @@
...
+			['/:extrinsicIndex', this.getExtrinsicByExtrsinsicIndex],
+		]);
+	}
+
+	/**
+	 *
+	 * @param _req Express Request
+	 * @param res Express Response
+	 */
+	private getExtrinsicByExtrsinsicIndex: RequestHandler<INumberParam> = async (
```
dvdplm, line 25:
```suggestion
	private getExtrinsicByIndex: RequestHandler<INumberParam> = async (
```
…seems less of a tongue twister?

docs/src/openapi-v1.yaml
dvdplm: I'm warming up to `Timepoint` but if we go that route we should be prepared to shill it across the board.

TarikGul: Yup exactly. I totally agree, we would definitely have to make it completely consistent across the board.

joepetrowski: Yeah, I'm on board with `getExtrinsicByTimepoint`. (Note the _singular_ `Extrinsic`. You either get all _extrinsics_ from a block, but a `Timepoint` always refers to a single extrinsic.)

docs/src/openapi-v1.yaml
```diff
@@ -273,6 +273,57 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/Error'
+  /blocks/{blockId}/extrinsics/{extrinsicIndex}:
+    get:
+      tags:
+      - blocks
+      summary: Get an extrinsic by its extrinsicIndex and block height or hash. 
+      The pair blockId, extrinsicIndex is sometimes referred to as a Timepoint.
+      description: Returns a single extrinsic.
+      operationId: getExtrinsicByExtrsinsicIndex
```
emostov, line 283:
```suggestion
      operationId: getExtrinsicByIndex
```

emostov (approved):
LGTM modulo a nit

docs/src/openapi-v1.yaml
TarikGul: Yea I like this as well. Will change throughout the code.

gutenye:
@TarikGul Thank you very much!

TarikGul:
@gutenye Of course, thank you for bringing up the issue.

Issue #380, opened 2021-01-06
Continuous package publishing


emostov:
Ideally, each Monday, once polkadot-js scoped deps publish a new release we can try to automatically merge the new polkadot-js deps and then do an automated release with changelog generation, new tag, and publish to NPM.

dvdplm:
I wonder how this would jive with 2-reviews-to-merge. What we want is some sort of override for that setting such that iff we get a clean merge and CI runs green it's ok to automerge.

TarikGul:
From what I have looked into, we can create a `dependabot.yml` and schedule pull requests for every monday to update packages we specify. We have the ability to assign reviewers as well. The only issue is how to bypass the reviewers.

emostov:
@TarikGul do you know how we can generate a `github_token`? If not I think @dvdplm would know who to ask.

emostov:
Also worth noting the current release workflow with the `deploy` script is broken. We manually edited the auto generated change log awhile back and I think that broke it. Maybe we can deprecate the current changelog and start fresh with a new changelog the complies without whatever release tool we decide to use?

TarikGul:
> @TarikGul do you know how we can generate a `github_token`? If not I think @dvdplm would know who to ask.

I believe the person with the admin rights to the repository can make a `github_token`. 

> Also worth noting the current release workflow with the `deploy` script is broken. We manually edited the auto generated change log awhile back and I think that broke it. Maybe we can deprecate the current changelog and start fresh with a new changelog the complies without whatever release tool we decide to use?

Yea i think thats worth updating as well. In the long run it will save us a lot of time, if we can make it do the right things efficiently. Im into it. Yea I don't know what the side effects would be if we deprecated the change log and started with a new one, I dont think it would be much. I will definitely check it out.

dvdplm:
> I believe the person with the admin rights to the repository can make a `github_token`.

FWIW you are both admins here.

TarikGul:
Part one of this issue is is merged. A pre-release dependabot that creates PR's every monday for updated polkadot dependencies. 

The next part will be to upgrade the current Changelog generation. One solution me and Zeke have talked about is deprecating the current changelog, and creating a new workflow that will automate future changelogs.

dvdplm:
>  One solution me and Zeke have talked about is deprecating the current changelog

Indeed. No intrinsic value in the git history of `CHANGELOG` I think. :)

TarikGul:
[...]

PR #401, opened 2021-01-25
chore(ci): pre-release dependabot

TarikGul:
This PR builds on issue [#380 ](https://github.com/paritytech/substrate-api-sidecar/issues/380).

This PR has changed from creating a automated publishing pipeline to a sole Dependabot script to create PR's every Monday with updated packaging from `@polkadot/..`. 

After some discussion and thought, it would make sense to do each section of automated publishing in seperate PR's and to build out the process piece by piece. 

```

 | Every Monday at XX:XX(am/pm) check for updates in the following 3 packages |
 |        1                   2                       3                       |
 | @polkadot/api => @polkadot/apps-config => @polkadot/util-crypto.           |
 | When the updates exists, create the PR.                                    |
  
```


emostov (approved):

dvdplm (approved):
Nice, it's more succinct than I thought it'd be.

cc @niklasad1 Maybe we can apply the same technique to other places? Nothing specific comes to mind, but good to know it's there. :)

.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      # 11 am PST
+      time: "11:00"
+      # PST timezone
+      timezone: "America/Los_Angeles"
+    rebase-strategy: "auto"
+    commit-message:
```
TriplEight, line 14:
```suggestion
    open-pull-requests-limit: 20
    commit-message:
```

TriplEight: In case you don't want to be spammed with these PRs, feel free to change the number.

TriplEight (approved):

.github/dependabot.yml
TarikGul: Good call, I'll definitely add it in, for now I think 10 will do. I saw that 5 is the default so I think having a little more room for open pull requests would definitely be beneficial. Thanks Denis.

PR #402, opened 2021-01-26
fix!: Update polkadot-js/api and account for chainProperties updates

emostov:
BREAKING CHANGE

- Document regression (breaking change) in `runtime/spec` response. `tokenSymbol` and `tokenDecimals` now return an `Arrary<string>` instead of a `string`. See the diff for `src/services/test-helpers/responses/runtime/spec.json` to view a concrete example of the change.
- Bumps dependencies, including polkadot-js
- Accounts for updates to `chainProperties.chainTokens`. Formerly, `chainProperties.chainToken` was just a `string`; now it is a `Array<string>`, so it can be extensible for chains that support multiple tokens. In this PR we make the assumption that the token at index 0 will be the native token. We consider a token native because its balance is stored in `system::Accounts`; opposed to a non-native token which has its balance accessible via the ORML `tokens` pallets.

### DIFF
--- a/src/controllers/accounts/AccountsBalanceInfoController.ts
+++ b/src/controllers/accounts/AccountsBalanceInfoController.ts
@@ -68,7 +68,8 @@ export default class AccountsBalanceController extends AbstractController<Accoun
 		const tokenArg =
 			typeof token === 'string'
 				? token.toUpperCase()
-				: this.api.registry.chainTokens[0];
+				: // We assume the first token is the native token
+				  this.api.registry.chainTokens[0].toUpperCase();
 
 		const hash = await this.getHashFromAt(at);
 
--- a/src/controllers/accounts/AccountsBalanceInfoController.ts
+++ b/src/controllers/accounts/AccountsBalanceInfoController.ts
@@ -68,7 +68,7 @@ export default class AccountsBalanceController extends AbstractController<Accoun
 		const tokenArg =
 			typeof token === 'string'
 				? token.toUpperCase()
-				: this.api.registry.chainToken;
+				: this.api.registry.chainTokens[0];
 
 		const hash = await this.getHashFromAt(at);
 
--- a/src/services/accounts/AccountsBalanceInfoService.spec.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.spec.ts
@@ -1,7 +1,6 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-return */
 import { AccountInfo, Address, Hash } from '@polkadot/types/interfaces';
 
 import { sanitizeNumbers } from '../../sanitize/sanitizeNumbers';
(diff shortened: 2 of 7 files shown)


TarikGul (commented):
LGTM.

TarikGul (approved):

src/controllers/accounts/AccountsBalanceInfoController.ts
```diff
@@ -68,7 +68,7 @@ export default class AccountsBalanceController extends AbstractController<Accoun
 		const tokenArg =
 			typeof token === 'string'
 				? token.toUpperCase()
-				: this.api.registry.chainToken;
+				: this.api.registry.chainTokens[0];
```
dvdplm, line 71: Don't we need to `.toUpperCase()` this one too?

Also: a comment documenting the assumption that `[0]` is the native token would be good?

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -26,7 +26,8 @@ export class AccountsBalanceInfoService extends AbstractService {
 		const { api } = this;
 
 		let locks, header, accountInfo, accountData;
-		if (token === api.registry.chainToken) {
+		// We assume the first token is the native token
```
dvdplm, line 29: 👍

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -26,7 +26,8 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+		if (token === api.registry.chainTokens[0]) {
```
dvdplm, line 30: Same question as above: are we positive `.chainTokens[0]` is uppercase?

dvdplm (changes requested):
I think the new format merits a test or two?

src/controllers/accounts/AccountsBalanceInfoController.ts
emostov: This is a good point. AFAIK there is nothing normalizng the token symbols to be uppercase, I just took it as convention. Changing it to always force upper case

emostov:
@dvdplm @TarikGul I added some tests and very slightly modified the service to (double) make sure when we compare `token` values they are all uppercase. Previously we forced it to uppercase in the controller but I think this change is more resilient to fat finger changes in the controller.

src/services/accounts/AccountsBalanceInfoService.spec.ts
```diff
@@ -18,5 +23,89 @@ describe('AccountsBalanceInfoService', () => {
 				)
 			).toStrictEqual(accountsBalanceInfo789629);
 		});
+
+		describe('token recognization', () => {
+			let tempQueryTokens: any,
+				tempQueryBalance: any,
+				mockTokensLocksAt: jest.Mock<any>,
+				mockTokenAccountAt: jest.Mock<any>,
+				mockBalancesLocksAt: jest.Mock<any>;
```
TarikGul, line 32: Is that standard for the linter to indent multiple line `lets` like that?

src/services/accounts/AccountsBalanceInfoService.spec.ts
```diff
@@ -18,5 +23,89 @@ describe('AccountsBalanceInfoService', () => {
 				)
 			).toStrictEqual(accountsBalanceInfo789629);
 		});
+
+		describe('token recognization', () => {
```
dvdplm, line 27:
```suggestion
		describe('token recognition', () => {
```

src/services/accounts/AccountsBalanceInfoService.spec.ts
```diff
@@ -18,5 +23,89 @@ describe('AccountsBalanceInfoService', () => {
...
+			let tempQueryTokens: any,
+				tempQueryBalance: any,
+				mockTokensLocksAt: jest.Mock<any>,
+				mockTokenAccountAt: jest.Mock<any>,
+				mockBalancesLocksAt: jest.Mock<any>;
+			beforeAll(() => {
+				// Important: these temp values should never be reassinged. They are used so we can assign
+				// the mockApi properties back to there orignal values after this sub-section of tests run.
```
dvdplm, line 35:
```suggestion
				// the mockApi properties back to their original values after this sub-section of tests run.
```

src/services/test-helpers/mock/index.ts
```diff
@@ -1,4 +1,16 @@
+import { Hash } from '@polkadot/types/interfaces';
+import { Codec, CodecArg } from '@polkadot/types/types';
+
 export * from './addresses';
 export * from './mockApi';
 export * from './mockBlock789629';
 export * from './transactions';
+
+/**
+ * Convience type for type casting mockApi storage querys.
```
dvdplm, line 10:
```suggestion
 * Convenience type for type casting mockApi storage querys.
```

dvdplm (approved):

src/services/accounts/AccountsBalanceInfoService.spec.ts
emostov: Yes and the 4 space tab looks pretty exagerated on github vs vscode. One of these days we can change all our repos to 2 space tabs to line up closer with the standard 2 spaces that most JS projects outside fo parity use.z

Issue #393, opened 2021-01-15
Discriminate `base_extrinsic` weight based on dispatch class


emostov:
rel: https://github.com/paritytech/substrate/pull/6629, 

The base weight for extrinsic can now be configured on a per dispatch class basis. The current fix introduced in #388 assumes the `base_extrinsic` weight will be the same across dispatch classes. While this holds true for polkadot runtimes at the moment, it is not an assumption we can safely use in the long run. 

A reasonable solution would be to modify `calc_fee` to take in the `base_extrinsic` weight each time it is called to calculate an extrinsic fee. We would then figure out the dispatch class and the `base_extrinsic` weight to use in TS.

PR #414, opened 2021-02-02
fix: discriminate `extrinsic_base_weight` based on dispatch class

TarikGul:
## Info: 

This PR addresses Issues [#393](https://github.com/paritytech/substrate-api-sidecar/issues/393). There were a number of changes to the structure of createCalcFee. Both on the typescript side and the rust-wasm package.

### @substrate/calc package (Rust-wasm)
`./calc/src/calc_fee.rs`

In this file I removed the `extrinsic_base_weight` from `from_params` which was getting called in `createCalcFee` once, and moved it to `calc_fee`. When fetch_block iterates through the extrinsic's it will call `calc_fee` and recalculate the fee based on the extrinsic dispatch class. 

### BlocksService (Typescript)
`./src/services/blocks/BlocksService.ts`

Added a private helper function `getParentParentHash`. 

Removed the `block` parameter from `createCalcFee` and added `parentParentHash` instead. This is because the function `createCalcFee` only took in the block parameter for the `parentParentHash` runtime work around (which I removed from the function). I needed to find the parentParentHash inside of `fetchBlock` instead, therefore for optimization purposes I removed it to avoid multiple rpc calls. 

I also transferred a lot of the `extrinsicBaseWeight` logic from `createCalcFee` and put it into fetchBlock where we iterate through the extrinsics. This is also why i make an rpc call for the metadata right before we iterate. 

## Extra

The one issue I am not too sure about is publishing `@substrate/calc`, currently I have the package set to `file:./calc/pkg`, not sure if i should change it back, but the tests will break if I do because it uses the older calc_fee.


emostov:
> The one issue I am not too sure about is publishing @substrate/calc, currently I have the package set to file:./calc/pkg, not sure if i should change it back, but the tests will break if I do because it uses the older calc_fee.

One way to handle this is once this PR is totally approved, `npm publish` `calc` from this branch, update the `package.json` to use that new version, make sure tests run and then merge this branch.

src/services/blocks/BlocksService.ts
```diff
@@ -524,6 +536,32 @@ export class BlocksService extends AbstractService {
 		};
 	}
 
+	/**
+	 * The block where the runtime is deployed falsely proclaims it would
```
dvdplm, line 540:
```suggestion
	 * The block where the runtime is enacted falsely proclaims it would
```

…or do I misunderstand. This is about the block where a new runtime is activated?

dvdplm (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -216,12 +227,62 @@ export class BlocksService extends AbstractService {
...
+				 * https://github.com/polkadot-js/api/issues/2365
+				 */
+				let extrinsicBaseWeight;
+				if (
+					specName !== api.runtimeVersion.specName.toString() ||
+					specVersion !== api.runtimeVersion.specVersion.toNumber()
+				) {
+					// We are in a runtime that does **not** match the decorated metadata in the api,
+					// so we must fetch the correct metadata, decorate it and pull out the constant
+					const decorated = expandMetadata(api.registry, metadata);
```
emostov, line 254: As mentioned above, I think this is expensive and only needs to be done once per block so we can move it out of the loop.

src/services/blocks/BlocksService.ts
```diff
@@ -156,12 +157,22 @@ export class BlocksService extends AbstractService {
...
 
+		const { calcFee, specName, specVersion } = await this.createCalcFee(
+			api,
+			parentHash,
+			parentParentHash
+		);
+
+		// Metadata used when the specVersion and specName do not match the
+		// runtimes specs.
+		const metadata = await api.rpc.state.getMetadata(parentParentHash);
```
emostov, line 174: This should only be done as need and parallel with other querys if possible.

emostov (changes requested):
Logic checks out with me with me. My main concern is just making sure we give a decent effort with optimization. I mention it in my code comments, but 
1) we should only fetch metadata as needed and it should be parallel with other querys if possible (I think this would mean doing it with the querys that are currently in `createCalcFee`. 
2) we should only need to create the decorated metadata once. (Potentially all of this could be done in `createCalcFee` and the decorated metadata could be returned if relevant)

The one other thing is just make sure to bump `calc` to `0.2.0`

src/services/blocks/BlocksService.ts
athei: I understand deployed and enacted as synonymous here: It is defined as the block that contains the `set_code` dispatchable that sets the new runtime code. The point is that when you query for the runtime version at that block via RPC you get the just newly deployed version. This is because on RPC a new wasm instance of the current code is created, which is the just deployed one.

However, block production still happens with the old instance which means that the block after the one with `set_code` is created with the old runtime. This is why we need this hack.

dvdplm: When I first wrote my comment I thought the new blob was already present on chain and activated at the given block. I later learned I was wrong, and your comment further clarifies things, ty. So "deployed" here is fine.

TarikGul: Thanks for the great explanation!

calc/src/calc_fee.rs
```diff
@@ -157,20 +159,20 @@ impl CalcFee {
...
+			adjusted_fee: {} base_fee: {} result: {}",
             weight,
             len,
             unadjusted_len_fee,
             unadjusted_weight_fee,
             adjustable_fee,
             adjusted_fee,
+            base_fee,
             result
         );
```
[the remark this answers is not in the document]
TarikGul: Rad, thanks for explaining. Yea was a little confused at first with the versioning.

src/services/blocks/BlocksService.ts
TarikGul: Yea as talked about earlier, i'll revert it to the original plan, and put metadata and the decorated variable back into `createCalcFee` and build some logic to stop any and all extra expensive calls.

src/services/blocks/BlocksService.ts
```diff
@@ -449,67 +503,36 @@ export class BlocksService extends AbstractService {
...
-			if (
+			runtimeDoesNotMatch =
 				specName !== api.runtimeVersion.specName.toString() ||
-				specVersion !== api.runtimeVersion.specVersion.toNumber()
-			) {
-				// We are in a runtime that does **not** match the decorated metadata in the api,
-				// so we must fetch the correct metadata, decorate it and pull out the constant
+				specVersion !== api.runtimeVersion.specVersion.toNumber();
+
+			if (runtimeDoesNotMatch) {
```
TarikGul, line 521: @emostov I could have put the metadata inside of the `Promise.all` but that being said It would have to be a ternary like [this](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/blocks/BlocksService.ts#L86). The problem with that is we don't have any `metadata` to resolve the promise with, and `expandMetaData` explicitly takes type `Metadata` only. Any thoughts?

src/types/responses/Block.ts
```diff
@@ -28,3 +30,16 @@ interface ILog {
...
+export interface ICalcFee {
+	calcFee:
+		| {
+				calc_fee: () => null;
+		  }
+		| CalcFee
+		| undefined;
+	specName: string | number;
+	specVersion: string | number;
+	decorated: DecoratedMeta | undefined;
```
emostov, line 43: prefer `decorated?: DecoratedMeta` when possible

emostov, line 43: prefer `decorated?: DecoratedMeta` when possible

src/types/responses/Block.ts
```diff
@@ -28,3 +30,16 @@ interface ILog {
...
+	runtimeDoesNotMatch: boolean | undefined;
```
emostov, line 44:
```suggestion
	decorated?: DecoratedMeta;
	runtimeDoesNotMatch?: boolean;
```

I read in the typescript docs a while back that this is preffered.

src/services/blocks/BlocksService.ts
```diff
@@ -243,13 +236,15 @@ export class BlocksService extends AbstractService {
...
-				if (
-					specName !== api.runtimeVersion.specName.toString() ||
-					specVersion !== api.runtimeVersion.specVersion.toNumber()
-				) {
-					// We are in a runtime that does **not** match the decorated metadata in the api,
-					// so we must fetch the correct metadata, decorate it and pull out the constant
-					const decorated = expandMetadata(api.registry, metadata);
+				if (runtimeDoesNotMatch) {
+					if (!decorated) {
+						extrinsics[idx].info = {
```
emostov, line 241: I am realizing I think we will still want the `RuntimeDispatchInfo` even if we can't calculate the partial fee. We are not planning on supporting fee calcs generically, but we should still support including the dispatch class and weight if possible. Originally, `calc_fee` just returned `null` for other chains, so the dispatch info would be 
```
{
    weight: 100,
    class: Normal,
    partialFee: null
}
```
I honestly forgot about this and I probably introduced a regression when I added in 
```
				if (!extrinsicBaseWeight) {
					throw new InternalServerError(
						'`extrinsicBaseWeight` is not defined when it was expected to be defined. File an issue at https://github.com/paritytech/substrate-api-sidecar/issues'
					);
				}
```
in the last PR. But I think going forward we should try to ensure we can deliver the runtime dispatch info as above and have a test to make sure a regression does not happen

emostov (changes requested):

src/services/blocks/BlocksService.ts
```diff
@@ -508,11 +503,29 @@ export class BlocksService extends AbstractService {
...
 			const [version, multiplier] = await Promise.all([
 				api.rpc.state.getRuntimeVersion(parentParentHash),
 				api.query.transactionPayment.nextFeeMultiplier.at(parentHash),
 			]);
 
+			runtimeDoesNotMatch =
+				specName !== api.runtimeVersion.specName.toString() ||
+				specVersion !== api.runtimeVersion.specVersion.toNumber();
+
+			if (runtimeDoesNotMatch) {
```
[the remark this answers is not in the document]
TarikGul: > Won't runtimeDoesNotMatch always be false? specName and specVersion look like they are undefined here. Looks like in this logic branch they only become defined on line 529:

Yes! great catch, thanks :)

src/types/responses/Block.ts
```diff
@@ -28,3 +30,11 @@ interface ILog {
 	index: number;
 	value: Codec;
 }
+
+export interface ICalcFee {
+	calcFee: null | CalcFee | undefined;
```
emostov, line 35:
```suggestion
	calcFee?: null | CalcFee;
```

emostov (approved):
LGTM!

All thats left to do is publish `calc` to npm, update sidecars package.json to use `calc` from npm, run tests and merge

PR #436, opened 2021-02-16
Fix the calc rust crate to work with the newest compiler

### LINKED ISSUE #430: Remove the `calc` crates dependence on `sp-arithmetic-legacy`
Currently, the `calc` crate relies on an old commit of substrate to reproduce a bug a in an old version of sp-arithmetic that affected fee calculations. Relying on old substrate commits will not work long term. Some options for tackling this issue that where proposed offline by @dvdplm include: 

* backport needed fixes to old substrate and maintain that fork forever
* copy the needed parts from old substrate and maintain only those forever, inside calc
* something else I haven't thought of

athei:
The calc package depends on two different version of `sp_arithmetic` which both no longer compile with the newest rust. This PR fixes that by doing this:

- Include `sp-arithmetic-legacy` inside this repository as a fork that includes the old buggy behaviour that we need to emulate for the first kusama runtime.
- Remove one used impl that prevented compilation of the package
- Fix the Cargo.toml to pull its dependencies from crates.io
- Update the "current sp_arithmetic" to v3 as `2.0.0-rc5` did not compile
- Update oudated dependencies while we at it

Fixes #430


dvdplm (approved):
This is great, ty.

I think we should slim down the legacy code to be as small as possible to make it easier for future devs to work with the code. Not necessary to do that in this PR, but if you're up for it, that would be great.

Also a unit test for `saturated_multiply_accumulate` would be good.

athei:
I have to disagree on "slimming it down". I see no reason for anyone to work with this code. We will certainly not do any feature enhancements to this code as its only for a single legacy runtime. Additionally, any bug in the calculation found will also not be fixed as this is the reason we have this crate in the first place: To emulate bugs.

emostov (approved):

IkerAlus:
The legacy `sp_arithmetic` is now removed as per this PR: https://github.com/paritytech/substrate-api-sidecar/pull/1146

PR #478, opened 2021-03-17
fix(`/blocks`): cache extrinsic base weight constants to improve performance

TarikGul:
This PR aims to fix a regression that happened when v2.1.2 was released. ie: [commit](https://github.com/paritytech/substrate-api-sidecar/commit/5ec24e63d571eaf348eb58053ccc8a7054276bd0). 

- [X] Dynamically generate hardcoded extrinsicBaseWeight, and blockWeight values depending on their grouped runtime.
    - [X] Polkadot
    - [X] Kusama 
- [X] Integrate `getBlockWeights` into `BlockService` with passing tests before passing in a cache config.
- [X] Create Caching system, saves on `expandMetadata` calls for unknown runtimes. 
    - [X] Change Cache object to only hold base weight to save on memory.
- [X] Add cache to the following services
    - [X] getExtrinsicByTimepoint
    - [X] getLatestBlock
    - [X] getBlockById
- [x] Add more test cases to make sure extrinsicBaseWeights is reacting properly. 

Update on benchmarks: After integrating `getBlockWeights` I made sure to do some benchmarking. I chose to target block `1907499` which has a staking bondExtra tx (Note I am querying a single block here). I ran a 30s Benchmark with 4 threads and 30 concurrent connections. (This was also done on my local Macbook pro which is extremely fast)

```
Before this PR:                     After:
--------------------- |           |----------------------
Avg Latency    : 1.13s            | Avg Latency   : 92.11ms
Completed Req  : 312              | Completed Req : 7392
```

To wrap this PR I ran a new set of benchmarks from our standard collection of 39 blocks that are used in `Sidecar-Bench`. I also ran it inside of our GCP instance, to recreate a standard environment. This benchmark is 2 minute's with 4 threads and 12 concurrent connections (Which is our standard test for Sidecar in GCP).

```
v4.0.1:                           This PR:
--------------------- |           |----------------------
Avg Latency    : 9.72s            | Avg Latency   : 5.28s
Completed Req  : 140              | Completed Req : 268
```


src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+ * a chain
+ *
+ * @param specName specName from the metadata of the current block being fetched
+ */
+export const getBlockWeight = (
+	specName: string
+): Record<string, BlockWeightDefinitions> | Object => {
+	switch (specName) {
+		case 'polkadot':
+			return Object.freeze(generateBlockWeightObject(polkadotDefinitions));
```
emostov, line 39: Just curious, why do `Object.freeze`?

TarikGul: It actually has no purpose at all, removing it :), started off with me testing it against changing types was all.

src/chains-config/metadata-consts/kusama-consts.ts
```diff
@@ -0,0 +1,62 @@
...
+			baseExtrinsic: 125000000,
+			maxExtrinsic: 371168000000001,
+			maxTotal: null,
+			reserved: 1500000000000,
+		},
+		operational: {
+			baseExtrinsic: 1,
+			maxExtrinsic: 32000000000,
+			maxTotal: null,
+			reserved: 1949875000000,
```
emostov, line 25: Do we need all these fields? Maybe just `baseExtrinsic` in each perClass?

src/chains-config/metadata-consts/kusama-consts.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * All Runtimes before v2027
+ */
+const extrinsicBaseWeight = 125000000;
+
+/**
+ * Runtimes after v2026
+ */
+const blockWeights = {
+	baseBlock: 5000000000,
```
emostov, line 12: Do these block base weights get used?

src/chains-config/metadata-consts/polkadot-consts.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * All Runtimes before v27
+ */
+const extrinsicBaseWeight = 125000000;
+
+/**
+ * Runtimes after v26
+ */
+const blockWeights = {
```
emostov, line 11: I am wondering if both kusama and polkadot should point at the same `blockWeights` and `extrinsicBaseWeights` object? Maybe we can make a shared file within this folder and import into these chain specific files?

src/services/blocks/BlocksService.ts
```diff
@@ -524,10 +575,52 @@ export class BlocksService extends AbstractService {
...
+	/**
+	 * Extracts necessary `extrinsicBaseWeight` or `blockWeights` data from
+	 * the decorated metadata in order to cache memory efficiently.
+	 *
+	 * @param decorated Decorated meta data to be parsed
+	 * @param runtimeVersion spec version of the blocks runtime
+	 * @returns
+	 */
+	private extractWeightFromDecorated(
+		decorated: DecoratedMeta,
```
TarikGul, line 590: I wonder if this should be `decorated.consts.system` and not the a pointer to the entire `DecorateMeta` object. I think that would make sense and allow for a testable `extractWeightFromDecorated`

src/types/chains-config/ControllerConfig.ts
```diff
@@ -25,3 +29,33 @@ export interface ControllerConfig {
 		finalizes: boolean;
 	};
 }
+
+/**
+ * Used in BlocksService, to persist decorated metadata to avoid expensive calls
+ */
+export interface CacheType {
```
emostov, line 36: Maybe name it `MetaConstsCache`

src/types/chains-config/ControllerConfig.ts
```diff
@@ -25,3 +29,33 @@ export interface ControllerConfig {
 		finalizes: boolean;
 	};
 }
+
+/**
+ * Used in BlocksService, to persist decorated metadata to avoid expensive calls
```
emostov, line 34: Change to something like `Cache for runtime version specific metadata constants.`

src/types/chains-config/ControllerConfig.ts
```diff
@@ -1,5 +1,9 @@
-import { controllers } from '../../controllers';
+// import { DecoratedMeta } from '@polkadot/metadata/decorate/types';
```
emostov, line 1: delete

src/services/blocks/BlocksService.ts
```diff
@@ -251,21 +265,35 @@ export class BlocksService extends AbstractService {
...
+				}
+
 				const len = block.extrinsics[idx].encodedLength;
 				const weight = weightInfo.weight;
 
 				const partialFee = calcFee.calc_fee(
 					BigInt(weight.toString()),
 					len,
-					extrinsicBaseWeight.toBigInt()
+					calcFeeExBaseWeight
```
emostov, line 296: Remove `Ex`?

src/services/blocks/BlocksService.ts
```diff
@@ -466,7 +502,7 @@ export class BlocksService extends AbstractService {
 			calcFee = null;
 
 			const version = await api.rpc.state.getRuntimeVersion(parentHash);
-			[specVersion, specName] = [
+			[specName, specVersion] = [
```
emostov, line 505: Nice catch

src/services/blocks/BlocksService.ts
```diff
@@ -500,14 +536,29 @@ export class BlocksService extends AbstractService {
 				version.specVersion.toNumber(),
 			];
 
-			runtimeDoesNotMatch =
-				specName !== api.runtimeVersion.specName.toString() ||
-				specVersion !== api.runtimeVersion.specVersion.toNumber();
+			const hardcodedChainIncluded = hardcodedChains.includes(specName);
 
-			if (runtimeDoesNotMatch) {
+			if (!hardcodedChainIncluded) {
```
emostov, line 541: This isn't the check we want to be doing.

1) We only support polkadot based chains + dock for fee calc as of right now
2) We need to get and expand metadata whenever there is a new runtime for one of the chains we support, i.e a live runtime upgrade and we have not hardcoded the values

This being said this whole branch of logic can probably be removed. I think the simplest way to do this is lazy filling of the cache thing. So basically somewhere before we need any weights we do the equivalent of the following logic
```javascript
metaConstsCache[runtimeVersion] ||= this.getDecorateAndExtractWeight(blockNumber)
```

And then every time forward we need weights for the runtime version its already in the `metaConstsCache`, so we don't need al this extra logic.

src/controllers/AbstractController.ts
```diff
@@ -40,6 +44,14 @@ export default abstract class AbstractController<T extends AbstractService> {
 		return this._router;
 	}
 
+	public get cache(): CacheType {
```
emostov, line 47: This shouldn't be here? Looks like it should be in BlocksController.

src/services/blocks/BlocksService.ts
```diff
@@ -70,7 +74,8 @@ export class BlocksService extends AbstractService {
 			checkFinalized,
 			queryFinalizedHead,
 			omitFinalizedTag,
-		}: FetchBlockOptions
+		}: FetchBlockOptions,
+		that: That
```
emostov, line 78: I honestly think this is a code smell. The `metaConstsCache` should belong to the `BlocksService`. I know there was some issues with this but we should be able to figure something once we get everything else straigtened out.

This current design means for every controller that uses the `BlocksService`, it will need to have its own, redundant `metaconstsCache`, which seems like a waste of memory since sidecar is designed to only connect to one chain.

src/services/blocks/BlocksService.ts
```diff
@@ -173,7 +177,7 @@ export class BlocksService extends AbstractService {
 
 			if (calcFee === null || calcFee === undefined) {
 				extrinsics[idx].info = {
-					error: `Fee calculation not supported for ${specName}#${specVersion}`,
+					error: `Fee calculation not supported for ${specVersion}#${specName}`,
```
emostov, line 180: :thumbs up

src/services/blocks/BlocksService.ts
```diff
@@ -444,18 +472,26 @@ export class BlocksService extends AbstractService {
...
 			api.consts.system.blockWeights.perClass.normal.baseExtrinsic;
 
-		let calcFee, specName, specVersion, decorated, runtimeDoesNotMatch;
+		/**
+		 * Chains where extrinsicBaseweight data is hardcoded in Sidecar
+		 * to avoid using expandMetadata to grab it
+		 */
+		const hardcodedChains = ['polkadot', 'kusama'];
+
+		let calcFee, specName, specVersion, decorated;
```
emostov, line 494: Instead of returning decorated this should just muatate the `metaConstsCache`.

Pseudo logic is 
```javascript
metaConstsCache[runtimeVersion] ||= this.getDecorateAndExtractWeight(blockNumber)
```

src/services/blocks/BlocksService.ts
```diff
@@ -500,14 +536,29 @@ export class BlocksService extends AbstractService {
...
+					// Decorate current metadata to read baseweight used in calcFee
+					decorated = expandMetadata(api.registry, metadata);
+
+					const cachedObject = this.extractWeightFromDecorated(
+						decorated,
+						specVersion
+					);
+
+					// Calls the cache setter
+					that.cache = { ...cachedObject };
```
emostov, line 560: See comment above - I think this should be local to this service.

One idea is maybe just initialize the cache when the service starts. So in the config.options it could pass down the chains spec name and then based on the specName it grabs the correct `metaConstsCache` with `getBlockWeight`. We won't be able to inject the `metaConstsCache` the same way for testing, but it seems less convoluted

src/services/blocks/BlocksService.spec.ts
```diff
@@ -178,13 +190,104 @@ describe('BlocksService', () => {
...
+				mockApi,
+				('0xParentHash' as unknown) as Hash,
+				mockBlock789629,
+				that
+			);
+
+			expect(that.cache.runtimeVersion).toBe(20);
+
+			(mockApi.runtimeVersion
+				.specVersion as unknown) = polkadotRegistry.createType('u32', 16);
```
emostov, line 287: You could save these to temp variables at the beggining and then re-assign at the end instead of recreating

emostov: f

emostov (changes requested):
Left my initial review

- I reffer to the cache as `metaConstsCache` since it is more descriptive to me
- I don't think the cache getter/setter should live in the abstract controller;  if they are to be in a controller it should be `BlocksController`.
- I think the `metaConstsCache` should be a private member of the `BlocksService`. There should be a way to do this
- The `metaConstCache` should be lazily filled - so if we see a block that belongs to a runtime we don't have an entry for than we can add the entry but getting and decorating the metadata. Otherwise we should not need to fetch/decorate meatdata
    - The expanding and decorating should be avoided if we don't support fee calcs for the chain. Maybe this can be a config an explicit config option? (If so the docs and chain integrations guide should be updated to reflect

src/chains-config/metadata-consts/kusama-consts.ts
```diff
@@ -0,0 +1,62 @@
...
+	maxBlock: 2000000000000,
+	perClass: {
+		normal: {
+			baseExtrinsic: 125000000,
```
emostov, line 16: Just double checking these do not need to be BN or BigInt?
If we convert them to one or the other later maybe best just to store them that way to avoid the downstream conversion

src/services/blocks/BlocksService.ts
emostov: FWIU, your suggestion makes sense

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
+import { Definition } from '../../types/chains-config';
+import { kusamaDefinitions } from './kusama-consts';
+import { polkadotDefinitions } from './polkadot-consts';
+
+/**
+ * Creates an object that orders each runtime to their appropriate data
```
dvdplm, line 6: This is too abstract; the code manipulates concrete types (`Definition`) and operates over concrete types (runtimes). I think it's good to state more up front what this does.

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param definitions An array of objects that group data based on runtimes
+ * and their extrinsicBaseWeight metadata
+ */
+const generateBlockWeightObject = (
+	definitions: Definition[]
+): Record<string, Definition> => {
+	const blockWeightObject = {};
+
+	for (let i = 0; i < definitions.length; i++) {
+		const runtimes: number[] = definitions[i].runtimes;
```
dvdplm, line 17: Name this `runtime_version` maybe? It is just the number right?

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+		for (let p = 0; p < runtimes.length; p++) {
```
dvdplm, line 19: These double for-loops give me pause. Can you give me an intuition for how big these collections are we're looping over? And why we're convinced they will never ever be too large for this to matter?

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+			const version: number = runtimes[p];
+			blockWeightObject[version] = definitions[i];
+		}
+	}
+
+	return blockWeightObject;
+};
+
+/**
+ * Returns a set of runtime pointing to their blockWeightDefinitions specific to
```
dvdplm, line 29:
```suggestion
 * Returns a set of runtimes pointing to their blockWeightDefinitions specific to
```

src/chains-config/metadata-consts/kusama-consts.ts
```diff
@@ -0,0 +1,62 @@
...
+		},
+		operational: {
+			baseExtrinsic: 1,
+			maxExtrinsic: 32000000000,
+			maxTotal: null,
+			reserved: 1949875000000,
+		},
+		mandatory: {
+			baseExtrinsic: 512000000000001,
+			maxExtrinsic: '0x00746a5288000100',
```
dvdplm, line 29: Why is this one a string? Too big for JS?

src/chains-config/metadata-consts/polkadot-consts.ts
dvdplm: Hmm, can you elaborate on why the two are unlikely to use different weights?

src/services/blocks/BlocksService.spec.ts
```diff
@@ -51,6 +52,12 @@ interface ResponseObj {
  */
 const blocksService = new BlocksService(mockApi);
 
+/**
+ * Object that would be a pointer to AbstractController where the cache would
+ * be stored
```
dvdplm, line 57: Using the conditional here is a bit disorienting. `is a pointer…` right?

src/services/blocks/BlocksService.spec.ts
```diff
@@ -51,6 +52,12 @@ interface ResponseObj {
...
+ */
+const that = ({ cache: {} } as unknown) as That;
```
dvdplm, line 59: Can you ELI5 what a `That` is?

src/services/blocks/BlocksService.ts
```diff
@@ -251,21 +265,35 @@ export class BlocksService extends AbstractService {
...
+					};
+					continue;
+				}
+
+				let calcFeeExBaseWeight;
+				if (typeof extrinsicBaseWeight !== 'number') {
+					calcFeeExBaseWeight = extrinsicBaseWeight.toBigInt();
+				} else {
+					calcFeeExBaseWeight = BigInt(extrinsicBaseWeight);
+				}
```
dvdplm, line 288: Could this conversion be just `let calcFeeExBaseWeight = extrinsicBaseWeight.toBigInt();`? It seems we're casting to `BigInt` regardless of the value being string or number so maybe we can save a few lines of code?

Also: this happens inside a for-loop; does `extrinsicBaseWeight` really change for each extrinsic? Maybe there's some optimization here too if, say, 90% of extrinsics are balance transfers?

src/services/blocks/BlocksService.ts
```diff
@@ -251,21 +265,35 @@ export class BlocksService extends AbstractService {
...
 				const weight = weightInfo.weight;
 
 				const partialFee = calcFee.calc_fee(
 					BigInt(weight.toString()),
 					len,
-					extrinsicBaseWeight.toBigInt()
+					calcFeeExBaseWeight
 				);
 
 				extrinsics[idx].info = api.createType('RuntimeDispatchInfo', {
```
dvdplm, line 299: Commenting here because I can't comment on line 304 because GitHub. It used to be that try-catch blocks were fairly expensive performance and memory wise. It's not immediately obvious to me what it buys us here as opposed to having an `all_good` boolean and, if it's false, print all `extrinsics` with an error, in the end?

Or maybe V8 is good enough these days that it doesn't matter?

src/services/blocks/BlocksService.ts
```diff
@@ -444,18 +472,26 @@ export class BlocksService extends AbstractService {
...
 	 * @param parentHash Hash of the parent block
 	 * @param block Block which the extrinsic is from
+	 * @param block
 	 */
 	private async createCalcFee(
 		api: ApiPromise,
 		parentHash: Hash,
-		block: Block
+		block: Block,
+		that: That
```
dvdplm, line 481: I think this deserves a better name. I have no idea what this does.

src/services/blocks/BlocksService.ts
```diff
@@ -444,18 +472,26 @@ export class BlocksService extends AbstractService {
 	 * @param api ApiPromise
 	 * @param parentHash Hash of the parent block
 	 * @param block Block which the extrinsic is from
+	 * @param block
```
dvdplm, line 475: Missing docs?

dvdplm (changes requested):
I'll get back to reviewing this, but left a few comments and questions. 

One more general question I have is how you connected the general benchmark results to this refactoring, i.e. how did you figure out that this was the problem?

src/chains-config/metadata-consts/index.ts
emostov: Based on the current code, the number of iterations == number of known runtimes.

src/chains-config/metadata-consts/polkadot-consts.ts
emostov: At the moment they use the same weights, so to DRY things we can just have one definition. 

In the future it is entirely possible that the two will diverge, but we will always need these definitions for older runtimes and might as well just define it once. So we can move the current `extrinsicBaseWeight` and `blockWeights` def to a common file (perhaps with more descriptive names linking them to the chain and runtime version) and then re-import them into both `polkadotConsts.ts` and `kusamaConsts.ts` and line them up to the appropriate runtime. If/when there is an update that diverges we would just add another def specific to that chain/runtime.

src/services/blocks/BlocksService.ts
emostov: I think extrinsicBaseWeight should just be persisted as a BigInt to avoid repetitive conversion.

src/controllers/AbstractController.ts
TarikGul: So the first time around I actually had the cache getter, setter, and object inside of `BlocksController`, but `BlocksExtrinsicController` also needs access to the cache so I decided I would put it in AbstractController instead.

src/chains-config/metadata-consts/polkadot-consts.ts
```diff
@@ -0,0 +1,68 @@
+import { Definition } from '../../types/chains-config';
```
emostov, line 1: File name should be camelCase for consistency. so polkadotConsts.ts . On the other hand we do folder names are kebab-case.

TarikGul:
@dvdplm I would actually hold off on reviewing any further if you'd like. Zeke and I did a walkthrough together, and saw some design changes that could optimize the whole workflow, so a lot of the things are definitely going to get updated and simplified. So it will probably look very different a day or two from now. Im still going to answer your above questions/concerns too though because I think a lot of them are still important :)

TarikGul:
@dvdplm In regards to how the regression was pin pointed. 

The benchmarks were able to reveal a range of commits to where the issue was actually possible. Specifically between v2.1.0 and v2.1.2.

 Then I started doing some more targeted benchmarks on specific commits(PR's), and moved my way forward, until I was able to confirm the actual PR that was causing the issue. Once that happened Zeke and I looked through the commit, and the only expensive call that was added was `expandMetadata`. So to confirm it was the issue, I injected a hardcoded extrinsicBaseWeight value in its place and started running benchmarks again and things dramatically changed and went back to its previous state. So from there I ran specific manual tests against expandMetadata and other RPC calls, and the time it takes to execute is astronomically heavier than every thing else. 
 
 That leaves us with 3 possibilities 
 
* If a chain does not need to calculate a fee, have a opt out config option (This is a new idea)
* For polkadot and Kusama hardcode these weight values.
* If its a non polkadot-kusama chain perform `expandMetadata` and Cache the necessary decorated metadata for queried blocks, tied to their runtimes for future use.

src/chains-config/metadata-consts/index.ts
TarikGul: Agreed, this could use more detailed explanation, and maybe even an output example.

src/chains-config/metadata-consts/index.ts
TarikGul: Yea of course, double for loops are always a sight for caution. In this case I think it's justifiable, because `i == 2` and wont get any larger. It's is equal to the amount of chains that are hardcoded ie `polkadot`, and `kusama`. Then in the case of `p`, it will only ever be as big as the number of runtime's that exists for Polkadot or Kusama. So that number will only grow by probabaly no more than 12 a year.

TarikGul:
Improvements to be made. 

* Add a config options to opt of of partialFee calculation

* Move the Cache out of `AbstractController` and decreasing its scope to be limited to `BlocksService`

* Refactor hardcoded weight data.

* Consolidate temp variable's inside of the test's

* Consolidate all the metaConstsCache logic into one class function.

src/chains-config/metadata-consts/index.ts
dvdplm: If `i` is known to be 2 maybe we can just hardcode the two? Maybe it's easier to read that way?

dvdplm:
>     * If a chain does not need to calculate a fee, have a opt out config option (This is a new idea)

Another thought I had yesterday was to send over a full batch of extrinsics to calculate the fees for, rather than one by one. Would have to benchmark the overhead of calculating fees 1000 times vs calculating all fees for a batch of 1000 extrinsics.

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+	const blockWeightObject = {};
+
+	for (let i = 0; i < definitions.length; i++) {
+		const runtimes: number[] = definitions[i].runtimes;
+
+		for (let p = 0; p < runtimes.length; p++) {
+			const version: number = runtimes[p];
+			blockWeightObject[version] = definitions[i];
+		}
+	}
```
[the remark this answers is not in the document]
TarikGul: Fixed!

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,45 @@
...
+	definitions: Definition[]
+): Record<string, Definition> => {
+	const blockWeightObject = {};
+
+	for (let i = 0; i < definitions.length; i++) {
+		const runtimes: number[] = definitions[i].runtimes;
+
+		for (let p = 0; p < runtimes.length; p++) {
+			const version: number = runtimes[p];
+			blockWeightObject[version] = definitions[i];
```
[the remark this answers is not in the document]
TarikGul: So I updated this, changed the naming and also did some abstraction with Definition. The reason it changed was because I also simplified the return value of `generateBlockWeightObject`. I think it also allows for more specific type's and is a lot easier to read. Let me know what you think.

src/chains-config/metadata-consts/kusama-consts.ts
TarikGul: Fixed!

src/chains-config/metadata-consts/kusama-consts.ts
TarikGul: Fixed!

src/types/chains-config/ControllerConfig.ts
TarikGul: Fixed

src/services/blocks/BlocksService.ts
TarikGul: Fixed!

src/controllers/AbstractController.ts
TarikGul: Fixed!

src/services/blocks/BlocksService.ts
TarikGul: Fixed!

src/services/blocks/BlocksService.ts
TarikGul: Fixed!

src/types/chains-config/ControllerConfig.ts
TarikGul: Fixed!

TarikGul:
Finished up optimizing and updating the code. I just need to run through and update the docs above `getDecorateAndExtractWeight`, see if there are any needed clarifying comments, and make sure everything is tidy and smooth. 

The very last thing I will do is run a set of benchmarks and compare them to v4.0.2.

src/services/blocks/BlocksService.spec.ts
TarikGul: Good question, in Javascript there is a convention that is referred to as `this = that`. Originally, the getter and setter for `this.cache` was in `AbstractController`, which `BlocksController` had access too but `BlocksService` didnt. To put it in simple terms when binding or inheriting `this` in one object to another we referr it to `that` 

1. because it cant be called `this`, because `this` already exists
2. That's how the community decided to name it many many years ago :)

But that being said, in this case I agree that the naming is bad and can be way more specific and something like `BlocksControllerThis` could probably be more useful and also get the point across. 

So to answer this question directly as well, `That` here is just the type for the `this` that was getting passed into `BlocksService` from `BlocksController`.

src/chains-config/metadata-consts/kusama-consts.ts
TarikGul: Yup exactly, Jaco has a limit for number size in `@polkadot/api` which will then represent it as a hex when the cap is reached.

src/chains-config/metadata-consts/index.ts
TarikGul: So I actually completely re worked this function, I still use a double for loop, but I think you'll find it to be a lot more readable

src/services/blocks/BlocksService.spec.ts
TarikGul: But to furthermore I actually completely redesigned this, and that is now completely removed. Things are now tightly scoped, and things are way easier to test now as well.

TarikGul:
> Another thought I had yesterday was to send over a full batch of extrinsics to calculate the fees for, rather than one by one. Would have to benchmark the overhead of calculating fees 1000 times vs calculating all fees for a batch of 1000 extrinsics.

Yea I think something like this could definitely be possible. The loop does service some nuances though such as figuring out they type of weightClass each extrinsic is. But that being said if we were to somehow have a hash that had every extrinsicType that point's to their weightClassInfo (ie: normal, operational, mandatory) I think this is doable.

TarikGul:
From the latest updates and optimizations, this is the latest benchmarks I got. @dvdplm @emostov The PR is ready for review's again. 

### /blocks
```
v4.0.1:                           The latest commits:
--------------------- |           |----------------------
Avg Latency    : 9.72s            | Avg Latency   : 4.39s
Completed Req  : 140              | Completed Req : 327
```

What I did from the last time this was reviewed. 

* I removed the cache from `AbstractController` and placed it inside `BlocksService`. This allows our scope to stay small, and allows better testing, debugging, and less intertwined state. 

* I removed all the logic inside of `fetchBlock` when it comes to deciding where to extract our weight data from. Instead `createCalcFee` calls `getDecorateAndExtractWeight` and there everything is extracted and the object to be returned is then stored into the cache. The only logic that should be in `fetchBlock` is to decide whether it's pulling extrinsicBaseWeight or blockWeights. 

* I also introduced `shouldCalcFee` which decides whether or not to calculate the fee for extrinsic's based on the specVersion. This came about because I found that older versions of Westend do not calculate Fees, and would cause errors. Therefore there are 3 hardcoded chains that have their minimum runtime version stored to make sure this error does not happen for Polkadot, Kusama, and Westend. I also added a config option called `SAS_SUBSTRATE_MIN_CALC_FEE_VERSION`. This will allow the user to specify a minimum runtime version to start calculating fees for. 

* I made the cache persist previously stored runtime weight data

* I added full test coverage that makes sure to coverage all cases for `getDecorateAndExtractWeight`.

emostov (changes requested):
@dvdplm @TarikGul and whomever else may review this: I created a branch and PR with my review for change up to commit 5326f0d.

See #493

src/test-helpers/metadata/polkadotV29Metadata.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ *  Polkadot v16 metadata. Useful for testing
```
emostov, line 2:
```suggestion
 *  Polkadot v29 metadata. Useful for testing
```

src/types/chains-config/MetadataConsts.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export interface MetadataWeightDefinition {
+	/**
+	 * Polkadot runtime versions before v0.8.27
+	 * Kusama runtime versions before v2027
+	 */
+	extrinsicBaseWeight?: number | AbstractInt;
+	/**
+	 * Polkadot runtime versions after v0.8.26
```
emostov, line 17: 0.8.X refers to the client release and all the runtimes in it. Polkadot v26 refers to the specVersion of the polkadot runtime in client release 0.8.26.

src/chains-config/metadata-consts/kusamaConsts.ts
```diff
@@ -0,0 +1,28 @@
...
+			2022,
+			2023,
+			2024,
+			2025,
+			2026,
+		],
+		extrinsicBaseWeight,
+	},
+	{
+		runtimeVersions: [2027, 2028, 2029],
```
dvdplm, line 25: What happens for future versions here? `2030` comes along and does not appear here; shouldn't we assume it's `perClass` too until we know it's not? Or maybe that's taken care of elsewhere?

src/chains-config/metadata-consts/substrateConsts.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * WEIGHT NOTES (POLKADOT | KUSAMA)
+ * @constant extrinsicBaseWeight
+ * @constant blockWeights
+ *
+ * The following weights are used for Polkadot and Kusama chains. Both are using
+ * the same fee calculations currently, and share the same Integer values. Because
```
dvdplm, line 7: Is this not true for `Westend` also?

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -1,13 +1,27 @@
...
 import { RequestHandler } from 'express';
+import { BlockWeightStore } from 'src/types/chains-config';
 
 import { BlocksService } from '../../services';
 import { INumberParam } from '../../types/requests';
 import AbstractController from '../AbstractController';
 
+interface ControllerOptions {
+	blockWeightStore: BlockWeightStore;
+	minCalcFeeRuntime: null | number;
```
dvdplm, line 11: What is our convention for these cases, isn't there `Option<Number>` in TS? We don't like it?

src/services/blocks/BlocksService.ts
```diff
@@ -57,10 +63,19 @@ enum Event {
 }
 
 export class BlocksService extends AbstractService {
+	constructor(
+		api: ApiPromise,
+		private minCalcFeeRuntime: null | number,
```
dvdplm, line 68: Can this be `Option<number>`?

src/services/blocks/BlocksService.ts
```diff
@@ -234,29 +262,24 @@ export class BlocksService extends AbstractService {
...
-				if (runtimeDoesNotMatch) {
-					if (!decorated) {
-						extrinsics[idx].info = {
-							error: 'Failure retrieving necessary decorated metadata',
-						};
-
-						continue;
-					}
+				if (
+					// 0n is a falsy value so we need to check if undefined
```
dvdplm, line 266:
```suggestion
					// 0 is a falsy value so we need to check if undefined
```

TarikGul:
Benchmarks for the final additions. 

### /blocks
```
v4.0.1:                           The latest commits:
--------------------- |           |----------------------
Avg Latency    : 9.72s            | Avg Latency   : 4.48s
Completed Req  : 140              | Completed Req : 329
```

src/services/blocks/BlocksService.ts
dvdplm: @TarikGul ^^^ didn't pan out? Or am I wrong?

src/types/chains-config/ControllerConfig.ts
```diff
@@ -23,5 +23,7 @@ export interface ControllerConfig {
 		 * Wether or not the chain finalizes blocks
 		 */
 		finalizes: boolean;
+		minCalcFeeRuntime: null | number;
```
dvdplm, line 26: Could this be `Option<number>`?

src/services/blocks/BlocksService.ts
```diff
@@ -450,84 +473,122 @@ export class BlocksService extends AbstractService {
...
+
+			weightValue = { perClass };
+		} else if (system.extrinsicBaseWeight) {
+			weightValue = {
+				extrinsicBaseWeight: ((system.extrinsicBaseWeight as unknown) as AbstractInt).toBigInt(),
+			};
+		} else {
+			throw new InternalServerError(
+				'Could not find a extrinsic base weight in metadata'
+			);
```
dvdplm, line 586: @TarikGul  @emostov is throwing errors the kosher way in `sidecar` to return errors? Not a fan of `try-catch` as you know and it'd be better if we had a proper error type we use consistently, but I'm might see this with Rust-colored glasses. 
If this is how we do things typically I'll shut up about it (for now). :)

src/test-helpers/metadata/metadata.ts
```diff
@@ -20,3 +24,11 @@ export const kusamaMetadata = new Metadata(
 	kusamaRegistry,
 	substrateMetadataRpc
 );
+
+/**
+ * Metadata of polkadotRegistry (v28)
+ */
+export const polkadotMetadataV29 = new Metadata(
+	polkadotRegistryV29,
+	polkadotV29MetadataRpc
```
dvdplm, line 33: Funny naming here, not sure if it's your code or not though.

```
polkadotV29Registry,
polkadotV29MetadataRpc
```

**or**

```
polkadotRegistryV29,
polkadotMetadataRpcV29
```

src/test-helpers/registries/polkadotRegistry.ts
```diff
@@ -20,12 +26,24 @@ function createPolkadotRegistry(): TypeRegistry {
...
  * Polkadot v16 TypeRegistry.
  */
-export const polkadotRegistry = createPolkadotRegistry();
+export const polkadotRegistry = createPolkadotRegistry(
+	16,
+	polkadotV16MetadataRpc
+);
+
+/**
+ * Polkadot v18 TypeRegistry
```
dvdplm, line 43: v18? Should be v29?

src/chains-config/metadata-consts/kusamaConsts.ts
TarikGul: Thats a good point. Currently the idea is before every runtime upgrade we just add the new runtime in with the weekly releases. Fortunately with the way it works, and is designed, is even if its not there the service will notice that and it will decorate the meta data and then cache the value inside of the `blockWeightStore`.

src/chains-config/metadata-consts/substrateConsts.ts
TarikGul: Yea thats a good point we can add it in as well for Westend. The original thought was I found that early versions of Westend didn't support fee calculations. So I didn't add it in.

src/controllers/blocks/BlocksExtrinsicsController.ts
TarikGul: Yea we can use it here, and I updated the code to use it here. 

`export type Option<T> = T | null;` 

And then set it to be Option<number>

src/services/blocks/BlocksService.ts
TarikGul: Fixed

src/types/chains-config/ControllerConfig.ts
TarikGul: fixed

src/test-helpers/registries/polkadotRegistry.ts
TarikGul: Fixed

src/types/util/Option.ts
```diff
@@ -0,0 +1 @@
+export type Option<T> = T | null;
```
dvdplm, line 1: So there is no such type in Typescript itself?

src/chains-config/metadata-consts/kusamaConsts.ts
dvdplm: > Fortunately with the way it works, and is designed, is even if its not there the service will notice that and it will decorate the meta data and then cache the value inside of the `blockWeightStore`.

Oh I missed that, can you point me to where in the code this happens?

TarikGul: Yea ofcourse the exact line is [here](https://github.com/paritytech/substrate-api-sidecar/blob/fix-metadata-regression/src/services/blocks/BlocksService.ts#L523). The `||=` operation says if the statement on the left hand side does not exist set it equal to the right hand operation, else dont do anything as it already exists.

src/types/util/Option.ts
TarikGul: Yea it doesn't exist out of the box.

src/services/blocks/BlocksService.ts
TarikGul: I think moving forward having our own internal Error type and, Error handler would be a great improvement. I do think the idea holds a lot of merit. Currently this is how things are done, but that being said there is room for improvement. I have seen monkey-patched `try-catch` implementation's for error handling that have improved performance in some blogs, so this some good food for thought.

src/services/blocks/BlocksService.ts
TarikGul: So `try-catch` block's themselves is actually pretty fast these days if an exception is not being thrown and as far as I know, from the research I did it has almost zero penalty (This is excluding throwing exceptions). In this case I don't think its absolutely necessary to remove the `try-catch` because of all the checks before calcFee is called which significantly removes the chance of an exception being thrown. That being said though if we start seeing exceptions thrown in the `try-catch` it will reveal deeper issues with `calc_fee`, or interdependent logic, which in retrospect should be caught beforehand if we have the necessary testing.

src/chains-config/metadata-consts/substrateConsts.ts
TarikGul: I just have to map out the runtimes, but will have this added in the next 30 mins.

src/services/blocks/BlocksService.ts
emostov: Yes, the errors are explicitly caught inside of the services with

https://github.com/paritytech/substrate-api-sidecar/blob/415f0302b05aeac013c93227ac19e4928427206a/src/controllers/AbstractController.ts#L77-L88

Once express 5.0 stabilizes this wrapper for requests will no longer be needed since it will catch async throws out of the box.

Also important to keep in mind that certain types of error formats are explicitly supported by sidecar's middleware: https://github.com/paritytech/substrate-api-sidecar/tree/master/src/middleware/error - this includes errors from the HttpError library that you see here.

I was not aware of any penalties for `try catch` when our wrote our error handling pipeline. With a quick google search, the first stake overflow answer seems to indicated that `try catch` is no slower than any other code since node v8.3 (we require min v12): https://github.com/paritytech/substrate-api-sidecar/tree/master/src/middleware/error

However probably warrants a little further investigation just to make sure there are not some low hanging optimizations.

src/test-helpers/metadata/polkadotV29Metadata.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ *  Polkadot v26 metadata. Useful for testing
```
emostov, line 2:
```suggestion
 *  Polkadot v29 metadata. Useful for testing
```

src/chains-config/metadata-consts/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * to their classification of extrinsic weight data.
+ *
+ * Example return object:
+ * {
+ * 	   ...
+ *     '24': { extrinsicBaseWeight: 125000000 },
+ *     '25': { extrinsicBaseWeight: 125000000 },
+ *     '26': { extrinsicBaseWeight: 125000000 },
+ *     '27': { blockWeights: { perClass: [Object] } },
+ *     '28': { blockWeights: { perClass: [Object] } }
```
emostov, line 25:
```suggestion
 *     24: { extrinsicBaseWeight: 125000000 },
 *     25: { extrinsicBaseWeight: 125000000 },
 *     26: { extrinsicBaseWeight: 125000000 },
 *     27: { blockWeights: { perClass: [Object] } },
 *     28: { blockWeights: { perClass: [Object] } }
```
These are numbers now :)

src/services/blocks/BlocksService.ts
TarikGul: > Once express 5.0 stabilizes this wrapper for requests will no longer be needed since it will catch async throws out of the box.

Didn't know about that improvement is coming in 5.0. Pretty awesome!

src/chains-config/metadata-consts/substrateConsts.ts
emostov: I think we actually should not add Westend. I don't want to establish the expectation that we will always support Westend with max optimizations. The block weight store still works with westend, we just won't hardcode values on startup so the first query in an unseen runtime will be slow but subsequent queries in that runtime will be fast - this should save us some surface area for manual maintenance.

src/controllers/blocks/BlocksExtrinsicsController.ts
emostov: @dvdplm If you have seen an option in the code before it is from polkadot-js which has an `Option` class which extends `Codec` - we mainly see it as a return type from storage queries.

@TarikGul if you add `Option` just make sure there is not a name space collision with the polkadot-js ones. Maybe we can do `SASOption`, 'IOption` etc.

src/chains-config/metadata-consts/substrateConsts.ts
TarikGul: Added!

src/types/util/Option.ts
emostov: Maybe we should make a PR for `Option` to [TypeScript](https://github.com/microsoft/TypeScript) ;) Native support for this would be very nice.

emostov (commented):
LMK when its ready for a thorough review - just mostly focused on responding to open comments for this one

src/services/blocks/BlocksService.ts
```diff
@@ -450,84 +474,122 @@ export class BlocksService extends AbstractService {
...
 			api.consts.system.blockWeights.perClass.normal.baseExtrinsic;
 
-		let calcFee, specName, specVersion, decorated, runtimeDoesNotMatch;
 		if (
-			perByte === undefined ||
-			extrinsicBaseWeightExists === undefined ||
+			!perByte ||
+			!extrinsicBaseWeightExists ||
+			(this.minCalcFeeRuntime && specVersion < this.minCalcFeeRuntime) ||
 			typeof api.query.transactionPayment?.nextFeeMultiplier?.at !== 'function'
```
emostov, line 511: We try and query this on line 491, so we can should just be able to check multiplier

src/services/blocks/BlocksService.ts
```diff
@@ -450,84 +474,122 @@ export class BlocksService extends AbstractService {
...
+	 * @returns formatted block weight store entry
+	 */
+	private async getWeight(
+		api: ApiPromise,
+		blockHash: BlockHash
+	): Promise<WeightValue> {
+		const metadata = await api.rpc.state.getMetadata(blockHash);
+		const {
+			consts: { system },
+		} = expandMetadata(api.registry, metadata);
```
emostov, line 557: We should keep an eye on what registry we use here in the future. There is a chance it may be better to pull the registry off `block` we get from the rpc response since that should be guranteed to have the exact types for the runtime. For now I _think_ this is fine.

emostov (approved):
Everything looks good to me. 

I tried making suggestions via the github vscode extension but botched it somewhere along the way and ended up just making a commit: https://github.com/paritytech/substrate-api-sidecar/pull/478/commits/20feb8e4ab2a632a2d9203aa9bad7fbeb4f34516

Mostly just updates to comments, and variable names. But I did take out that giant `try catch` when we loop through extrinsics. I honestly think if that is catching we are not doing the correct checks prior so my proposal is to remove it and if it becomes a production issue we can address it

src/services/blocks/BlocksService.ts
dvdplm: Cool, I rest my case.

(That is, my main beef with `try-catch` and exceptions in general is that one can never be really sure where errors are caught without reading through the whole reverse call-chain. They call this "spooky action at a distance" I think.)

src/services/blocks/BlocksService.ts
```diff
@@ -508,7 +503,7 @@ export class BlocksService extends AbstractService {
 			!perByte ||
 			!extrinsicBaseWeightExists ||
 			(this.minCalcFeeRuntime && specVersion < this.minCalcFeeRuntime) ||
-			typeof api.query.transactionPayment?.nextFeeMultiplier?.at !== 'function'
+			!multiplier
```
dvdplm, line 506: 👍

src/test-helpers/metadata/metadata.ts
```diff
@@ -6,15 +6,15 @@ import {
 	polkadotRegistry,
 	polkadotRegistryV29,
 } from '../registries/polkadotRegistry';
-import { polkadotV16MetadataRpc } from './polkadotV16Metadata';
-import { polkadotV29MetadataRpc } from './polkadotV29Metadata';
+import { polkadotMetadataRpcV16 } from './polkadotV16Metadata';
+import { polkadotMetadataRpcV29 } from './polkadotV29Metadata';
```
dvdplm, line 10: 👍

src/types/util/Option.ts
```diff
@@ -1 +1,2 @@
-export type Option<T> = T | null;
+// Not to be confused with the polkadot-js `Codec` extending `Option` type.
```
dvdplm, line 1: good pt!

src/types/util/Option.ts
```diff
@@ -1 +1,2 @@
...
+export type IOption<T> = T | null;
```
dvdplm, line 2: So what is the TS convention for when to use the `I` prefix?

dvdplm (approved):
lgtm, good job.

src/types/util/Option.ts
emostov: I thinks its more just a me convention that I have implicitly picked up from reading other `TS` code - `I` is really for interface, but I use it both for interface or types to avoid name collisions with classes / functions etc.

emostov:
@lovesh you should take a look at the dock controller configs to make sure they are still valid. Additionally you may be interested in adding the equivalent of [this](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/chains-config/metadata-consts/polkadotConsts.ts) for the dock chains.

src/types/util/Option.ts
TarikGul: So in typescript (this is quite the debate apparently), "I" should refer to interface's only, and when its not there it will refer to an `impl`. Even though this example doesn't fit those parameters the use of 'I' here is to make sure there is a distinguishing difference as stated in the comment.

lovesh:
@emostov Thanks for the notification. Submitted change at https://github.com/paritytech/substrate-api-sidecar/pull/511

Issue #479, opened 2021-03-19
Unknown types found, no types for BlockLength, BlockWeights, ConsumedWeight, SeatHolder, Voter


phillux:
**Description**

Substrate API Sidecar does not connect due to BlockWeights issue

**Steps to Reproduce**

- run polkadot archive node 
- run substrate-api-sidecar

**Expected vs. Actual Behavior**

It should connect, but it doesn't.

`$ substrate-api-sidecar --rpc-cors=all
SAS:
  📦 LOG:
     ✅ LEVEL: "info"
     ✅ JSON: false
     ✅ FILTER_RPC: false
     ✅ STRIP_ANSI: false
  📦 SUBSTRATE:
     ✅ WS_URL: "ws://127.0.0.1:9944"
  📦 EXPRESS:
     ✅ BIND_HOST: "127.0.0.1"
     ✅ PORT: 8080
2021-03-19 13:18:23 warn: API/INIT: RPC methods not decorated: payment_queryFeeDetails, system_addLogFilter, system_resetLogFilter, system_unstable_networkState
2021-03-19 13:18:24 warn: METADATA: Unknown types found, no types for BlockLength, BlockWeights, ConsumedWeight, SeatHolder, Voter
2021-03-19 13:18:24 warn: REGISTRY: Unable to resolve type BlockWeights, it will fail on construction
2021-03-19 13:18:24 error: REGISTRY: Error: Cannot construct unknown type BlockWeights
    at new DoNotConstruct (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/primitive/DoNotConstruct.js:21:11)
    at new <anonymous> (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/primitive/DoNotConstruct.js:27:9)
    at new Type (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/create/registry.js:300:16)
    at initType (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/create/createType.js:49:19)
    at createTypeUnsafe (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/create/createType.js:68:12)
    at createType (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/create/createType.js:83:10)
[...]

PR #480, opened 2021-03-19
feat: Log version before creating ws connection

emostov:
This PR simply adds a log message of the version of api-sidecar prior to creating the `ApiPromise` and starting up the websocket connection.

If the `ApiPromise` fails to connect (due to type issues, node not running etc.) the logging prompts following will never show up. Thus, logging the version prior to the most common failure point for new users is prudent for helping debug ([example](https://github.com/paritytech/substrate-api-sidecar/issues/479#issuecomment-802928219)).

Below is an example of the startup logging when the api fails to connect because there is no websocket port at the specified address.

Screen Shot 2021-03-19 at 9 04 38 AM

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -22,6 +22,7 @@ import { WsProvider } from '@polkadot/rpc-provider';
 import { OverrideBundleType, RegistryTypes } from '@polkadot/types/types';
 import { json } from 'express';
 
+import packageJSON from '../package.json';
 import App from './App';
 import { getControllersForSpec } from './chains-config';
 import { consoleOverride } from './logging/consoleOverride';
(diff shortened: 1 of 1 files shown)


TarikGul (approved):
Nice addition, LGTM

emostov:
@dvdplm I like the idea. See new issue #484 

Makes me wonder if we should allow configuration through CLI as well instead of ENV vars?

dvdplm:
> Makes me wonder if we should allow configuration through CLI as well instead of ENV vars?

There be dragons. We'll have to handle people setting part of their configs with CLI args and some with ENV vars and write code to merge options from multiple sources. It's best if we don't go there unless we have a compelling reason to imo.

Issue #468, opened 2021-03-10
Update Staking Endpoints for *30 changes


joepetrowski:
https://github.com/paritytech/substrate/pull/8113 will remove the off-chain election and corresponding `Open`, `Closed` statuses. Will need to update the API and docs to handle the new staking API.

Should support the legacy (V5) staking for some period for any parachains that might still be using it.

dvdplm:
Fixed by #485 

(@TarikGul seems like the github AI isn't AI enough to autoclose issues with "resolves issue x"; need to use "fixes x" or "closes x" I think)

PR #485, opened 2021-03-23
fix: update eraElectionStatus for runtime upgrade v30

TarikGul:
This PR resolves issue [#468](https://github.com/paritytech/substrate-api-sidecar/issues/468).

Updated `eraElectionStatus` to work for parachains, but also to work with the upcoming runtime upgrade. 

With the upcoming runtime Upgrade v30, `eraElectionStatus` will no longer be supported and therefore deprecated.

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -1309,7 +1309,7 @@ components:
         status:
           type: object
           description: >-
-            [Deprecated](Works for runtimes before v30).   
+            [Deprecated](Works for polkadot runtimes before v0.8.30).   
             
             Era election status: either `Close: null` or `Open: <BlockNumber>`.
             A status of `Close` indicates that the submission window for solutions
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -1309,6 +1309,8 @@ components:
         status:
           type: object
           description: >-
+            [Deprecated](Works for runtimes before v30).   
+            
             Era election status: either `Close: null` or `Open: <BlockNumber>`.
             A status of `Close` indicates that the submission window for solutions
             from off-chain Phragmen is not open. A status of `Open` indicates that the
--- a/src/services/pallets/PalletsStakingProgressService.ts
+++ b/src/services/pallets/PalletsStakingProgressService.ts
@@ -20,16 +20,26 @@ export class PalletsStakingProgressService extends AbstractService {
 		const [
 			validatorCount,
 			forceEra,
-			eraElectionStatus,
 			validators,
 			{ number },
 		] = await Promise.all([
 			api.query.staking.validatorCount.at(hash),
 			api.query.staking.forceEra.at(hash),
-			api.query.staking.eraElectionStatus.at(hash),
 			api.query.session.validators.at(hash),
 			api.rpc.chain.getHeader(hash),
 		]);
+
+		let eraElectionStatus;
+		try {
+			eraElectionStatus = await api.query.staking.eraElectionStatus.at(hash);
+		} catch {
+			/**
+			 * Runtime v30 and above do not support eraElectionStatus, so use
+			 * a `try` to retrieve the eraElectionStatus, and if were running on a
+			 * runtime less than v30 it will return a successful result. If it doesn't
+			 * we do nothing and let the undefined `eraElectionStatus` dictate the below logic.
+			 */
+		}
 
 		const {
 			eraLength,
@@ -88,7 +98,10 @@ export class PalletsStakingProgressService extends AbstractService {
 		if (electionLookAhead.eq(new BN(0))) {
 			// no offchain solutions accepted
 			toggle = null;
-		} else if ((eraElectionStatus as { isClose?: boolean }).isClose) {
+		} else if (
+			eraElectionStatus &&
+			(eraElectionStatus as { isClose?: boolean }).isClose
+		) {
 			// election window is yet to open
 			toggle = nextCurrentEra.sub(electionLookAhead);
 		} else {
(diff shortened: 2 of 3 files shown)


src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -20,17 +20,27 @@ export class PalletsStakingProgressService extends AbstractService {
...
+		try {
+			eraElectionStatus = await api.query.staking.eraElectionStatus.at(hash);
+		} catch {
+			/**
+			 * Runtime v30 and above do not support eraElectionStatus, so use
+			 * a `try` to retrieve the eraElectionStatus, and if were running on a
+			 * runtime less than v30 it will return a successful result. If it doesn't
+			 * we do nothing and let the undefined `eraElectionStatus` dictate the below logic.
+			 */
+		}
```
dvdplm, line 42: What does `api.query.staking.eraElectionStatus` return on `v30` and beyond? I'd have thought it returns an error (or is even `undefined`?). I think it's better to avoid try-catch when we can, as they risk catching and silencing unexpected errors. So if something like this works, it's better imo:

```suggestion
eraElectionStatus = await api.query.staking.eraElectionStatus.at(hash);
if (!eraElectionStatus) {
  // report warning?
}
```

dvdplm (changes requested):

src/services/pallets/PalletsStakingProgressService.ts
TarikGul: Yea of-course. I think that's totally fair.

To build off your example it would look more like 

```typescript
let eraElectionStatus;
if(api.query.staking.eraElectionStatus) {
    eraElectionStatus = await api.query.staking.eraElectionStatus.at(hash);
}

// Else do nothing, allow `eraElectionStatus` to be undefined, and use a ternary to report it's been Deprecated.
```

But yes I agree this could be much simpler with potentially less overhead via ditching `catch`. The only thing to make sure is we don't check `api.query.staking.eraElectionStatus.at()` because the undefined `at` with the undefined `eraElectionStatus` will cause 500 error.

TarikGul:
@dvdplm Awesome I switched the conditional around to remove the `try catch`, I agree that it is way better now.

docs/src/openapi-v1.yaml
```diff
@@ -1309,6 +1309,8 @@ components:
         status:
           type: object
           description: >-
+            [Deprecated](Works for runtimes before v30).
```
emostov, line 1312:
```suggestion
            [Deprecated](Works for polkadot runtimes before v0.8.30).   
```

src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -20,17 +20,26 @@ export class PalletsStakingProgressService extends AbstractService {
...
 			api.query.staking.validatorCount.at(hash),
 			api.query.staking.forceEra.at(hash),
-			api.query.staking.eraElectionStatus.at(hash),
 			api.query.session.validators.at(hash),
 			api.rpc.chain.getHeader(hash),
 		]);
 
+		let eraElectionStatus;
+		/**
+		 * Runtime v30 and above do not support eraElectionStatus, so we check
```
emostov, line 34:
```suggestion
		 * Polkadot runtimes v0.8.30 and above do not support eraElectionStatus, so we check
```

src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -20,17 +20,26 @@ export class PalletsStakingProgressService extends AbstractService {
...
+		 * to see if eraElectionStatus is mounted to the api, and if were running on a
+		 * runtime less than v30 it will return a successful result. If it doesn't
```
emostov, line 36:
```suggestion
		 * runtime less than v0.8.30 it will return a successful result. If it doesn't
```

src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -100,7 +112,9 @@ export class PalletsStakingProgressService extends AbstractService {
 			...baseResponse,
 			nextActiveEraEstimate: nextActiveEra.toString(10),
 			electionStatus: {
```
emostov, line 114: I think it is more clear to have `electionStatus: "Deprecated, see docs"` instead of having the inner `status` property say that. RN it jus looks like the inner `status` field is deprecated and not the whole `electionStatus`

emostov (changes requested):
Core logic looks good but I think the `electionStatus` field should be left out or marked deprecated, not just the `status` sub property.

src/services/pallets/PalletsStakingProgressService.ts
TarikGul: Awesome, fixed it up so that electionStatus return the message instead.

src/services/pallets/PalletsStakingProgressService.ts
```diff
@@ -88,7 +97,10 @@ export class PalletsStakingProgressService extends AbstractService {
 		if (electionLookAhead.eq(new BN(0))) {
 			// no offchain solutions accepted
 			toggle = null;
-		} else if ((eraElectionStatus as { isClose?: boolean }).isClose) {
+		} else if (
+			eraElectionStatus &&
+			(eraElectionStatus as { isClose?: boolean }).isClose
```
emostov, line 102:
```suggestion
			(eraElectionStatus as { isClose?: boolean })?.isClose
```
Does that work?

emostov (approved):

src/services/pallets/PalletsStakingProgressService.ts
emostov: We should be able to eliminate that `eraElectionStatus &&` check with the `?` operator

emostov (approved):

src/types/responses/PalletStakingProgress.ts
```diff
@@ -10,9 +10,11 @@ export interface IPalletStakingProgress {
...
-	electionStatus?: {
-		status: AnyJson;
-		toggleEstimate: string | null;
-	};
+	electionStatus?:
+		| {
+				status: AnyJson;
+				toggleEstimate: string | null;
+		  }
+		| string;
```
dvdplm, line 18: dq: What do the `|`s do here?

dvdplm (approved):

src/types/responses/PalletStakingProgress.ts
TarikGul: So here we have a Union Type, the first `|` denotes that the type has not yet been decided but it will guaranteed be one of the following below. And all following `|`'s are there to allow `electionStatus` to fulfill multiple types.

PR #493, opened 2021-03-27
[do not merge into master] Review and suggestions for #478

emostov:
Review for #478 

Most of my suggestions for design where easier to write in code so I branched from commit `5326f0d`.

Overarching take aways

- Simplify arbitrarily nested types. I realize this was an attempt to future proof but I think it makes things harder to follow and get the type annotations right.
- Store types in the format they will be used in. If we need a BigInt just store a BigInt - not a number or something that extends the `Codec` class.
- We know ahead of time that some old runtimes don't support fee calcs, so we can just combine that information with the rest of the existing chain specific configs. This should not be configurable by the user - sidecar should (and currently does) default to not supporting fee calcs due to the complex nature.
- Try not to clobber typescripts type annotations unless we are testing it or we are certain it can be ducked type to the new type.
- I made the block store injectable to the service via the chain config options - hopefully easier to test and reduces the amount of logic in the service that needs to be intertwined with things outside of the class.

See the diff for specifics of my recommendations.

Notes: 
- I updated tests just enough to compile, but not to work.
- If you incorporate these changes we should do some more manually testing, particularly on the westend blocks you where having problems with before.


src/services/blocks/BlocksService.ts
```diff
@@ -62,20 +64,17 @@ enum Event {
...
+		this.blockWeightStore =
+			minCalcFeeRuntime !== null
+				? getBlockWeight(api.runtimeChain.toString())
+				: {};
+		this.minCalcFeeRuntime = minCalcFeeRuntime;
 	}
 
-	private set metaConstsCache(cacheObject: MetaConstsCache) {
-		const currentCache = Object.assign({}, this.metaConstsCache);
-		this._metaConstsCache = { ...currentCache, ...cacheObject };
```
emostov, line 77: What was the purpose of setting this way?

src/services/blocks/BlocksService.ts
```diff
@@ -451,220 +474,118 @@ export class BlocksService extends AbstractService {
...
+				},
+				mandatory: {
+					baseExtrinsic: mandatory.baseExtrinsic.toBigInt(),
+				},
+			};
 
-		// Set the runtime version key to `decoratedKeyValues`.
-		extractedWeight[specVersion] = decoratedKeyValues;
-
-		const doesCacheThisChain = specName === 'polkadot' || specName === 'kusama';
```
emostov, line 610: We shouldn't need to be hard coding stuff like this since we have the configs for each chain

PR #501, opened 2021-03-30
fix(BlocksService): check all materials prior to creating calcFrom.from_params

emostov:
This PR adds a single defensive check to `createCalcFee` in order to make sure we are not mapping over undefined `weightToFee`.


TarikGul (approved):
LGTM

dvdplm (approved):

PR #504, opened 2021-04-05
Add endpoints

emostov:
Adds endpoints:
- `paras/`: list of all registered parachains and their lifecycle stage.
- `/:paraId/lease-info`: current and future lease information for a para.
- `/:paraId/crowdloan-info`: info on a para's current crowdloan.
- `/leases/current`: list of all lease holders for the current lease period and general information about the current lease period.
- `/auctions/current`: information on the current auction.

Future PRs:
- Test for endpoints
- Docs
- Maybe merge `crowdloan-info` & `lease-info` into a single endpoint that just has all the plo info that is specific to a chain.

### DIFF
--- a/src/services/paras/ParasService.ts
+++ b/src/services/paras/ParasService.ts
@@ -26,8 +26,8 @@ export type AuctionPhase = 'PreEnding' | 'Ending';
 
 export class ParasService extends AbstractService {
 	/**
-	 * @param hash
-	 * @param paraId
+	 * @param hash `BlockHash` to make call at
+	 * @param paraId ID of para to get crowdloan info for
 	 * @returns crowdloan information for paradId
 	 */
 	async crowdloansInfo(
@@ -71,8 +71,8 @@ export class ParasService extends AbstractService {
 	}
 
 	/**
-	 * @param hash
-	 * @param includeFundInfo
+	 * @param hash `BlockHash` to make call at
+	 * @param includeFundInfo wether or not to include `FundInfo` for every crowdloan
 	 * @returns list of all crowdloans
 	 */
 	async crowdloans(
@@ -113,6 +113,12 @@ export class ParasService extends AbstractService {
 		};
 	}
 
+	/**
+	 * @param hash `BlockHash` to make call at
+	 * @param paraId ID of para to get lease info of
+	 * @returns current and future lease info as well as lifecycle information for
+	 * a given `paraId`
+	 */
 	async leaseInfo(hash: BlockHash, paraId: number): Promise<any> {
 		const [leases, { number }, paraLifeCycle] = await Promise.all([
 			this.api.query.slots.leases.at<
@@ -175,6 +181,12 @@ export class ParasService extends AbstractService {
 		};
 	}
 
+	/**
+	 *
+	 * @param hash `BlockHash` to make call at
+	 * @returns information on the current auction. Most fields will be null if
+	 * if there is no ongoing auction.
+	 */
 	async auctionsCurrent(hash: BlockHash): Promise<any> {
 		const [auctionInfoOpt, { number }, auctionCounter] = await Promise.all([
 			this.api.query.auctions.auctionInfo.at<Option<Vec<AbstractInt>>>(hash),
@@ -227,12 +239,21 @@ export class ParasService extends AbstractService {
 			beginEnd,
 			finishEnd,
 			phase,
+			// If there is no current auction, this will be the index of the previous auction
 			auctionIndex: auctionCounter,
 			leasePeriods,
 			winning,
 		};
 	}
 
+	/**
+	 *
+	 * @param hash `BlockHash` to make call at
+	 * @param includeCurrentLeaseHolders wether or not to include the paraIds of
+	 * all the curent lease holders. Not including is likely faster and reduces
+	 * response size.
+	 * @returns general information about the current lease period
+	 */
 	async leasesCurrent(
 		hash: BlockHash,
 		includeCurrentLeaseHolders: boolean
@@ -271,6 +292,11 @@ export class ParasService extends AbstractService {
 		};
 	}
 
+	/**
+	 *
+	 * @param hash `BlockHash` to make call at
+	 * @returns all the current registered paraIds and their lifecycle status
+	 */
 	async paras(hash: BlockHash): Promise<any> {
 		const [{ number }, paraLifecycles] = await Promise.all([
 			this.api.rpc.chain.getHeader(hash),
--- a/src/services/paras/ParasService.ts
+++ b/src/services/paras/ParasService.ts
@@ -33,6 +33,7 @@ export class ParasService extends AbstractService {
 	/**
 	 * Get crowdloan information for a `paraId`.
 	 *
+         *  `/:paraId/crowdloan-info` Service function to retrieve crowdloan information for a specific paraId.
 	 * @param hash `BlockHash` to make call at
 	 * @param paraId ID of para to get crowdloan info for
 	 */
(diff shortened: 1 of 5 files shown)


src/services/paras/ParasService.ts
```diff
@@ -1,41 +1,67 @@
...
-	 * @param paraId
-	 * @returns
-	 * {
-	 *   "at": {},
-	 *   "paraId": number,
-	 *   "fundInfo": {...}
-	 * }
+	 * @param hash `BlockHash` to make call at
+	 * @param paraId ID of para to get crowdloan info for
+	 * @returns crowdloan information for paradId
```
TarikGul, line 36: I don't think we need this, We don't document any of the `@returns` in any other services, we could get rid of this. Up to you though.

src/services/paras/ParasService.ts
```diff
@@ -45,51 +71,44 @@ export class ParasService extends AbstractService {
...
-	 *   "funds": [
-	 *     {
-	 *       "paraId": number,
-	 *       "fundInfo": {}
-	 *     }
-	 *   ]
-	 * }
+	 * @param hash `BlockHash` to make call at
+	 * @param includeFundInfo wether or not to include `FundInfo` for every crowdloan
+	 * @returns list of all crowdloans
```
TarikGul, line 83: Same as before. Maybe we could remove these `@returns` and write doc info above the `@param` notes.

src/services/paras/ParasService.ts
```diff
@@ -45,51 +71,44 @@ export class ParasService extends AbstractService {
 		return {
 			at,
 			fundInfo,
+			// TOOD would it make snese to merge the below derived info into `fundInfo`
```
TarikGul, line 74: Little confused on  this could you elaborate?

src/services/paras/ParasService.ts
```diff
@@ -1,41 +1,67 @@
...
 	/**
-	 * @param hash
-	 * @param paraId
-	 * @returns
-	 * {
-	 *   "at": {},
-	 *   "paraId": number,
-	 *   "fundInfo": {...}
-	 * }
+	 * @param hash `BlockHash` to make call at
```
TarikGul, line 34:
```suggestion
         *  `/:paraId/crowdloan-info` Service function to retrieve crowdloan information for a specific paraId.
	 * @param hash `BlockHash` to make call at
```

Maybe something like this instead of the `@returns`?

src/services/paras/ParasService.ts
TarikGul: I don't think this is really that important more so an observation of continuity between all services.

src/services/paras/ParasService.ts
```diff
@@ -102,4 +121,255 @@ export class ParasService extends AbstractService {
...
+			height: blockNumber.toString(10),
+		};
+
+		let leasesFormatted;
+		if (leases.length) {
+			const currentLeasePeriodIndex = this.currentLeasePeriodIndex(
+				blockNumber
+			).toNumber();
+
+			leasesFormatted = leases.reduce((acc, curLeaseOpt, idx) => {
```
TarikGul, line 152: Very Nice!

src/services/paras/ParasService.ts
```diff
@@ -102,4 +121,255 @@ export class ParasService extends AbstractService {
...
+					});
+				}
+
+				return acc;
+			}, [] as LeaseFormatted[]);
+		} else {
+			leasesFormatted = null;
+		}
+
+		let onboardingAs: ParaType | undefined;
```
TarikGul, line 169: Could we make an `IOption<T>` for undefined's?

src/services/paras/ParasService.ts
```diff
@@ -102,4 +121,255 @@ export class ParasService extends AbstractService {
...
+				this.api.query.slots.leases.entriesAt<
+					Vec<Option<ITuple<[AccountId, BalanceOf]>>>,
+					[ParaId]
+				>(hash),
+			]);
+
+			blockNumber = number.unwrap();
+
+			currentLeaseHolders = leaseEntries
+				.filter(([_k, leases]) => leases[0].isSome)
```
TarikGul, line 278: I like the use of `isSome`, I think it has the ability to also optimize code in a lot of other service, etc

src/services/paras/ParasService.ts
```diff
@@ -102,4 +121,255 @@ export class ParasService extends AbstractService {
...
+		const [{ number }, paraLifecycles] = await Promise.all([
+			this.api.rpc.chain.getHeader(hash),
+			this.api.query.paras.paraLifecycles.entriesAt<ParaLifecycle, [ParaId]>(
+				hash
+			),
+		]);
+
+		const parasPromises = paraLifecycles.map(async ([k, paraLifeCycle]) => {
+			const paraId = k.args[0];
+			let onboardingAs: ParaType | undefined;
```
TarikGul, line 311: Same thing here could we have a Option for undefined.

src/types/responses/Paras.ts
```diff
@@ -1,19 +1,154 @@
...
 
-export interface IEntries {
-    paraId: ParaId;
+export type AuctionPhase = 'preEnding' | 'ending';
+
+export type ParaType = 'parachain' | 'parathread';
+
+export interface IFund {
+	/**
+	 * Id of the para that has a crowloan.
```
TarikGul, line 22:
```suggestion
	 * Id of the parachain or parathread that has a crowdloan.
```

src/types/responses/Paras.ts
```diff
@@ -1,19 +1,154 @@
...
+	 */
+	paraId: ParaId;
+	/**
+	 * `FundInfo` for the para's crowdloan. Use query param `fundInfo=true` to include the `FundInfo`.
+	 * Inclusion may make query take longer.
```
TarikGul, line 27:
```suggestion
	 * Inclusion may make the query take longer.
```

src/types/responses/Paras.ts
```diff
@@ -1,19 +1,154 @@
...
+	 * in the `Ending` phase and using the `winning` bids.
+	 */
+	phase: IOption<AuctionPhase>;
+	/**
+	 * Auction number. If there is no current auction this will be the number of
+	 * the previous auction.
+	 */
+	auctionIndex: AbstractInt;
+	/**
+	 * Lease period indexs that may be bid on in this auction.
```
TarikGul, line 106:
```suggestion
	 * Lease period indexes that may be bid on in this auction.
```

src/types/responses/Paras.ts
```diff
@@ -1,19 +1,154 @@
...
+	endOfLeasePeriod: BN;
+	/**
+	 * ParaIds of current lease holders.
+	 */
+	currentLeaseHolders?: ParaId[];
+}
+
+export interface IPara {
+	/**
+	 * ParaId
```
TarikGul, line 134:
```suggestion
	 * Id of a parachain or parathread
```

TarikGul (approved):
LGTM, little grammar nit pics. 

Also comments about the `@returns`, up to you if you want them there or not.

src/services/paras/ParasService.ts
emostov: Basically should we add `leasePeriods` as a key within the `fundInfo` object?

src/services/paras/ParasService.ts
emostov: I would be open to it, but would want to call it something else. 
My only hold up though are that `undefined` fn params and interface properties already have the `?` operator which is normally preferred over having a union type with undefined.

src/services/paras/ParasService.ts
emostov: Yup - just to clarify though, this is `.isSome` is  a property on the polkadot-js `Option` enum which extends `Codec` - which is different than the `IOption` type alias we have defined within this repo.

PR #511, opened 2021-04-08
feat: Add weight for Dock mainnet

lovesh:
Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/src/chains-config/dockMainnetControllers.ts
+++ b/src/chains-config/dockMainnetControllers.ts
@@ -1,4 +1,5 @@
 import { ControllerConfig } from '../types/chains-config';
+import { getBlockWeight } from './metadata-consts';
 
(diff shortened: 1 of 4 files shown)


TarikGul (approved):
LGTM

emostov (approved):
lgtm - @lovesh would you like to make sure fee calcs are enabled on the test net as well? You could just changes the `null` to `1` here: 
https://github.com/paritytech/substrate-api-sidecar/blob/6294925d229208ed6fa755721ee99a81396e4fb6/src/chains-config/dockTestnetControllers.ts#L29

And then the block weight store would still be able to be dynamically built up as the querys are made.

lovesh:
@emostov Done. Thanks.

PR #512, opened 2021-04-09
feat: add testing for plo endpoints

TarikGul:
Added:

* Rococo Registry and metadata
* TypeFactory class
* Tests for the following methods
    * `ParasService.crowdloansInfo`
    * `ParasService.crowdloans`
    * `ParasService.leaseInfo`
    * `ParasService.leasesCurrent`
    * `ParasService.paras`
    * `ParasService.auctionsCurrent`


src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +289,188 @@ const referendumInfoOfAt = () =>
...
+
+const parasLifecyclesAt = () =>
+	Promise.resolve().then(() => {
+		return typeFactory.optionOf(paraLifecycleOne);
+	});
+
+/**
+ * Used for parachain leases
+ */
+const tupleOne = new Tuple(
```
TarikGul, line 398: I need to extract this call to `Tuple` and have it live in the `TypeFactory`, same for the following ones below.

TarikGul: Resolved!

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +289,187 @@ const referendumInfoOfAt = () =>
...
+			[paraId2, vectorLeases],
+		];
+	});
+
+/**
+ * Used for parachain Auctions
+ */
+const beingEnd = new BN(1000) as AbstractInt;
+const leasePeriodIndex = new BN(39) as AbstractInt;
+const vectorAuctions = typeFactory.vecOf([beingEnd, leasePeriodIndex]);
```
emostov, line 428: `auctions::AuctionInfo` is a `Option<(LeasePeriodOf<T>, T::BlockNumber)>` so `Option<[beginEnd, leasePeriodIndex]>`. I don't see the `Option` part here. In the service we have it modelled as `Option<Vec<AbstractInt>>` - should wrap this in an `Option`

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +289,187 @@ const referendumInfoOfAt = () =>
...
+			['AccountId', 'ParaId', 'BalanceOf']
+		);
+		const parasOptionsOne = typeFactory.optionOf(tupleOne);
+		const parasOptionsTwo = typeFactory.optionOf(tupleTwo);
+
+		const vectorWinnings = typeFactory.vecOf([
+			parasOptionsOne,
+			parasOptionsTwo,
+		]);
+		const optionWinnings = typeFactory.optionOf(vectorWinnings);
```
emostov, line 468: Should this be longer? like a length of 7 (Or whatever the `SLOT_RANGE_COUNT` is)? The rest of the values can just be `Option::None`, but I am concerned there could be behavior that relies on the length of `auctions::Winning`. 

It is defined as `[Option<(<T as frame_system::Config>::AccountId, ParaId, BalanceOf<T>)>; SlotRange::SLOT_RANGE_COUNT]` where `SLOT_RANGE_COUNT` is the length it will always be.

src/test-helpers/registries/rococoRegistry.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	registry.register(getSpecTypes(registry, 'Rococo', 'rococo', 228));
+
+	registry.setMetadata(new Metadata(registry, rococoMetadataV228));
+
+	return registry;
+}
+
+/**
+ * Kusama v228 TypeRegistry.
```
emostov, line 22:
```suggestion
 * Rococo v228 TypeRegistry.
```

src/test-helpers/typeFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	constructor(api: ApiPromise) {
+		this.#api = api;
+		this.#registry = this.#api.registry;
+	}
+
+	public storageKey = (index: number): StorageKey => {
+		const key = new StorageKey(
+			this.#registry,
+			this.#api.query.crowdloan.funds.key(this.paraIndex(index))
```
emostov, line 43: Oh cool, didn't know this would work. Does `.key` create the key?

src/test-helpers/typeFactory.ts
```diff
@@ -0,0 +1,77 @@
...
+		);
+
+		return key.setMeta(this.#api.query.crowdloan.funds.creator.meta);
```
emostov, line 46: What is `creator`?

emostov (changes requested):
Overall it looks good, but some minor grumbles

src/services/test-helpers/mock/mockApi.ts
TarikGul: Yea so this is actually wrapped in an Option inside of `auctionInfoAt`, but there were two methods using `vectorAuction's` so I didn't wrap it in an `Option` in the outer scope. But you raised a good point with a previous comment `typeFactory.optionOf(null)` so I fixed it all to be more on point, and now the variable are all inside of `auctionInfoAt` and don't need to be in the outer scope anymore. 

Inside of TypeFactory we now have a `nullOf`, and that allows you too create a Null primitive type and pass it into Option. Good catch and great addition.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +289,187 @@ const referendumInfoOfAt = () =>
...
+export const auctionsInfoAt = (): Promise<Option<Vec<AbstractInt>>> =>
+	Promise.resolve().then(() => {
+		const optionAuctions = typeFactory.optionOf(vectorAuctions);
+
+		return optionAuctions;
+	});
+
+export const nullAuctionsInfoAt = (): Promise<Vec<AbstractInt>> =>
+	Promise.resolve().then(() => {
+		return vectorAuctions;
```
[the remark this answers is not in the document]
TarikGul: Yup great catch. This is definitely a grumble for sure and I fixed it up. I added a `typeFactory.nullOf` that returns a 'Null' primitive type, and then you can pass that into `Option` to mock `Option<None>`. 

My logic here worked, but it wasn't optimal nor was it situationally correct.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +289,187 @@ const referendumInfoOfAt = () =>
...
+		return [
+			[paraId1, vectorLeases],
+			[paraId2, vectorLeases],
+		];
+	});
+
+/**
+ * Used for parachain Auctions
+ */
+const beingEnd = new BN(1000) as AbstractInt;
```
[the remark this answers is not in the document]
TarikGul: Agreed, after this PR merges, I can make a small PR that adds it to the typeFactory, and we can refactor some of the service code.

src/test-helpers/typeFactory.ts
TarikGul: `creator` allows you to access the metadata to decode the necessary args as well as passing necessary type information to unwrap the `StorageType` and set it to the output Type.

src/test-helpers/typeFactory.ts
TarikGul: Yea So `.key` will create the key but then in order to set the metadata for the key, `setMeta` needs to be called on it.

src/services/test-helpers/mock/mockApi.ts
TarikGul: So I couldnt find where it says the minimum `SLOT_RANGE_COUNT` is, but I know it's suppose to be a maximum of 10 per the older code about a week ago.

```
/// Total number of possible sub ranges of slots.
pub const SLOT_RANGE_COUNT: usize = 10;
```
This was the comment that they had for slot range. 

Now the codes changed a bit which includes a `generate_slot_range` macro. 

`/// * A constant `SLOT_RANGE_COUNT` will count the total number of enum variants.`

TarikGul:
@emostov I restructured the tests so that they all reflect 'user contract' tests. They all use `sanitizeNumbers` as well as `toMatchObject` to mock an actual user response.  I also added a `typeFactory.nullOf` for Null primitives that can be added to `Option`. Everything is now looking good, the only last piece is the Slot range constant that is a little confusing. Whats I guess a little confusing for me is the macro itself `generate_slot_range!`.

src/services/paras/ParasService.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+				paras: [
+					{
+						paraId: '199',
+						paraLifeCycle: 'Onboarding',
+						onboardingAs: 'parachain',
+					},
+					{
+						paraId: '200',
+						paraLifeCycle: 'Parathread',
+						onboardingAs: undefined,
```
emostov, line 176: I think you could completely leave out these undefined fields if you wanted to

src/services/paras/ParasService.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+					},
+				],
+			};
+
+			const response = await parasService.auctionsCurrent(blockHash789629);
+
+			expect(sanitizeNumbers(response)).toMatchObject(expectedResponse);
+		});
+
+		it('Should return the correct null values when auctionInfo is not of type Option', async () => {
```
emostov, line 221:
```suggestion
		it('Should return the correct null values when auctionInfo is `None`', async () => {
```
It should be of type `Option`: `Option::None`. In JSON though that gets converted into `null`.

emostov: The `sanitizeNumber` specs are exhaustive for this stuff. See

https://github.com/paritytech/substrate-api-sidecar/blob/ae74ef77cf05c68fc7a552b8cccc200262f721be/src/sanitize/sanitizeNumbers.spec.ts#L572-L575

src/services/paras/ParasService.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+	describe('ParasService.auctionsCurrent', () => {
+		it('Should return to correct data during an ongoing auction', async () => {
+			const expectedResponse = {
+				at: expectedAt,
+				beginEnd: '39',
+				finishEnd: '20039',
+				phase: 'ending',
+				auctionIndex: '4',
+				leasePeriods: ['1000', '1001', '1002', '1003'],
+				winning: [
```
emostov, line 196: I am leaning towards us adjusting `winning` to return a vec with length equal to the slot ranges, so length 10. The rest could just be `Option::None`

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +288,189 @@ const referendumInfoOfAt = () =>
...
+		const vectorAuctions = typeFactory.vecOf([beingEnd, leasePeriodIndex]);
+		const optionAuctions = typeFactory.optionOf(vectorAuctions);
+
+		return optionAuctions;
+	});
+
+export const nullAuctionsInfoAt = (): Promise<Option<Null>> =>
+	Promise.resolve().then(() => {
+		const nullPrimitive = typeFactory.nullOf();
+		const nullOption = typeFactory.optionOf(nullPrimitive);
```
emostov, line 438: Can you just pass in `null` here instead of creating this more complicated polkadot-js null thing?

src/services/test-helpers/mock/mockApi.ts
emostov: Lets hardcode it to 10 (I think we covered the rest in offline discussion.)

src/test-helpers/typeFactory.ts
```diff
@@ -0,0 +1,81 @@
...
+	};
+
+	public nullOf = (): Null => {
```
emostov, line 49: We can remove this if we don't need it for line 438.

emostov (approved):
LGTM % nits

src/services/test-helpers/mock/mockApi.ts
TarikGul: Well no is the simplest answer. The way its implemented to in polkadot/api is to create a `emptyOption` method. Which is what I just added. It's definitely better than adding the `nullOf` method as you mentioned. 

```
public emptyOption = <T extends Codec>(typeName: keyof InterfaceTypes): Option<T> => 
        new Option<T>(this.#registry, typeName);
```

And then we just add the Null primitive as the typeName

src/services/paras/ParasService.spec.ts
TarikGul: fixed

src/services/paras/ParasService.spec.ts
TarikGul: ahh gotcha, yea that makes sense. Empty Options are returned as null with sanitize numbers :)!

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -276,6 +290,198 @@ const referendumInfoOfAt = () =>
...
+		// Instead of using emptyOption here we use optionOf since we need the 
+		// winning data options to consist of tuples. This will also help satisfy the 10
+		// options we need for a LEASE_PERIODS_PER_SLOT_FALLBACK of 4. 
+		const emptyTuple = typeFactory.tupleOf([], []);
+		const emptyOption = typeFactory.optionOf(emptyTuple);
+		const mockWinningOptions = new Array(8).fill(
+			emptyOption
+		) as Option<Tuple>[];
+
+		// Total of 36 winning object, 2 for testing, 34 for mock
```
emostov, line 474:
```suggestion
		// Total of 10 winning object, 2 `Some(..)`, 8 `None`
```

PR #513, opened 2021-04-12
fix: Query lease periods per slot; use exact int types

emostov:
Changes:

- Dynamically `leasePeriodsPerSlot`, but use a fallback to ensure compatibility w/ current `rococo-v1` branch which does not expose the constant in metadata (https://github.com/paritytech/polkadot/pull/2862)
- Use exact int type annotations for relevant queries instead of `AbstractInt`

### DIFF
--- a/src/services/paras/ParasService.ts
+++ b/src/services/paras/ParasService.ts
@@ -1,3 +1,4 @@
+import { u32 } from '@polkadot/types';
 import { Option, Vec } from '@polkadot/types/codec';
 import { AbstractInt } from '@polkadot/types/codec/AbstractInt';
 import {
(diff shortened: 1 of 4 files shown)


src/services/paras/ParasService.ts
```diff
@@ -258,8 +262,11 @@ export class ParasService extends AbstractService {
 			winning = null;
 		}
 
+		const leasePeriodsPerSlot =
```
TarikGul, line 265: Assuming this will be `8` in the future, and there will be no need for `LEASE_PERIODS_PER_SLOT_FALLBACK`?

emostov: Yup exactly, check line 35

src/services/paras/ParasService.ts
```diff
@@ -228,7 +232,7 @@ export class ParasService extends AbstractService {
 				  );
 
 			if (winningOpt.isSome) {
-				const ranges = ParasService.enumerateLeaseSets(leasePeriodIndex);
+				const ranges = this.enumerateLeaseSets(leasePeriodIndex);
```
TarikGul, line 235: Good catch!

TarikGul (approved):
LGTM, after this is merged I will pull this into the test branch and adjust it appropriately.

PR #515, opened 2021-04-13
fix(leasesCurrent): refactor the filter and map for the keys, and adjust the tests

TarikGul:
I changed the way we filter the keys and retrieve the paraId. I also adjusted the tests as well.

### DIFF
--- a/src/services/paras/ParasService.ts
+++ b/src/services/paras/ParasService.ts
@@ -314,7 +314,7 @@ export class ParasService extends AbstractService {
 			blockNumber = number.unwrap();
 
 			currentLeaseHolders = leaseEntries
-				.filter(([_k, leases]) => leases[0].isSome)
+				.filter(([key, _l]) => !!key)
 				.map(([key, _l]) => key.args[0]);
 		}
 
--- a/src/services/test-helpers/mock/mockApi.ts
+++ b/src/services/test-helpers/mock/mockApi.ts
@@ -405,18 +405,17 @@ const leasesTupleTwo = typeFactory.tupleOf(
 );
 const parasOptionsOne = typeFactory.optionOf(leasesTupleOne);
 const parasOptionsTwo = typeFactory.optionOf(leasesTupleTwo);
-const vectorLeases = typeFactory.vecOf([parasOptionsOne, parasOptionsTwo]);
 
 export const slotsLeasesAt = (): Promise<Vec<Option<Tuple>>> =>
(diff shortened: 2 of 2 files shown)


src/services/paras/ParasService.ts
```diff
@@ -314,7 +314,7 @@ export class ParasService extends AbstractService {
 			blockNumber = number.unwrap();
 
 			currentLeaseHolders = leaseEntries
-				.filter(([_k, leases]) => leases[0].isSome)
+				.filter(([key, _l]) => !!key)
```
emostov, line 317: Can you elaborate on the issue and how this fix addresses it?

I don't understand - if we get the entries there should alway be a key value pair, so not sure why the key would ever be falsey, unless it was `0` (which should not happen since it is `Codec`)

Additionally since we are getting the _current_ lease holders we have to check that the paraId has a lease in the current lease index - which is the first index of each paraId's `leases`, hence why we check `leases[0].isSome`.

emostov (changes requested):

src/services/paras/ParasService.ts
TarikGul: Function: `this.api.query.slots.leases.entriesAt`

### Our Type: 
`<Vec<Option<Tuple<[AccountId, BalanceOf]>>>, [StorageKey<ParaId>]>`;

NOTE: I also missed this earlier, so this was a good thing to go over cause I also misunderstood one part of the `Vec` because we didn't actually use it. The api call returns an `Array or Vector` that's holding a bunch of sub arrays, that consist of a `StorageKey` in the first index, and the necessary Lease information is stored in the next indice. Every sub array represents one lease, so therefore I just grab the key from each first index, and retrieve the paraId.  

### Actual Return Type: 
`[ [StorageKey<ParaId>], Vec<Option<Tuple<[AccountId, BalanceOf]>>>][]`

### Ex: Returned data structure:

```javascript
[
    // Lease #One
    [
        StorageKey: [Type],
        Vec
    ],
    // Lease #Two
    [
        StorageKey: [Type],
        Vec 
    ]
]
```

> so not sure why the key would ever be falsey, unless it was 0 (which should not happen since it is Codec)

So the `!!k` check is to make sure that the ParaId exists. In fact we could also technically remove the filter completely. Its just more there as a sanity check.

With that though, this should also be updated to the correct type `as <Vec<Option<Tuple<[AccountId, BalanceOf]>>>, [StorageKey<ParaId>]>`

Issue #24, opened 2020-03-28
Run yarn start failed


JaniceLee629:
When I update the latest version of the program and run it on a mac, I get this error

yarn run v1.22.0
$ tsc && node ./build/main.js
Running on http://127.0.0.1:18080/
events.js:298
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use 127.0.0.1:18080
    at Server.setupListenHandle [as _listen2] (net.js:1309:16)
    at listenInCluster (net.js:1357:12)
    at doListen (net.js:1496:7)
    at processTicksAndRejections (internal/process/task_queues.js:85:21)
Emitted 'error' event on Server instance at:
    at emitErrorNT (net.js:1336:8)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '127.0.0.1',
  port: 18080
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

I checked, 18080 port is not occupied by other programs

joepetrowski:
There have been some updates since this was posted. I just ran on a macbook pro without problem. Is it a specific block or endpoint that gave a problem?

joepetrowski:
Closing as this is out of date

PR #518, opened 2021-04-21
build(yarn): Upgrade from yarn 1.x to yarn 2.x

TarikGul:
This PR aims to be a test PR that will facilitate upgrading Sidecar and hopefully the rest of our javascript repos to use yarn 2.x in the future. 

NOTE: I will reference `yarn 2.x` as `yarn berry` and `yarn 1.x` as `yarn classic` below. 

### Concerns with yarn 1.22.0:

* It is unmaintained other than security fixes. There is no foreseeable future for the `yarn classic`, especially now that `yarn berry` has backwards compatibility with node_modules, and isn't only featuring the PnP (Plug and Play) architecture.

* As Yarn 2.x grows migration from using a global package such as `yarn classic` to `yarn berry` might grow harder and more complicated. As `yarn berry` works on a per project basis it makes sense to get off of `yarn classic` sooner than later.

### Why yarn 2.x

* Stability: With the change from classic to berry a lot of core design issues were upgraded and are now maintained with the new package. 

* Features: This will allow for a more active Open Source community where bugs and features will be fixed and rolled out on a regular basis. 

    * `yarn dedupe` being one I am absolutely excited about. No more having to delete `node_modules`, and `yarn.lock` in order to fix issues with `@polkadot/util-crypto`
    
    * `.yarnrc.yml` allowing for more custom configuration and flexibility. 

* Efficiency: "Modern features new install strategies, leading projects to only be a fraction of their past self; as an example, under the default configuration the stock CRA artifacts now only take 45MB instead of 237MB. Performances were improved as well, with most installs now only taking a few seconds even on extremely large projects. We even made it possible to reach zero seconds!" (Taken directly from their docs)

More on the improvements [here](https://yarnpkg.com/getting-started/qa)

* Plugin's: One such as `upgrade-interactive` - "This command opens a fullscreen terminal interface where you can see any out of date packages used by your application, their status compared to the latest versions available on the remote registry, and select packages to upgrade."

### Overall 

I think this should be something we look into and seriously consider. As projects start to phase out of `yarn classic` its important to keep up and make sure if we can migrate and take advantage of the upcoming toolsets available to us we should. 

That being said there are still a few kinks I need to learn and understand. 

1. Why isn't `berry` able to launch `jest` straight from `node_modules` even though I have it linked, whereas classic `yarn` can read it. Same goes for `eslint`.

    * Once this is resolved I will remove the deps `jest` and `eslint`.

2. Inside of the `tsconfig` I need to path the `extends` to `./node_modules/@substrate/dev` not just `@substrate/dev` anymore.


dvdplm:
I don't have a lot to say about the changes per se other than I approve of the general thrust (and appreciate the thorough PR description, I learned a lot, ty!).

Random idea: should we open similar PRs to our other JS code and see if it works seamlessly there too before taking the plunge? Maybe it's better to do each project one by one?

TarikGul:
Yea I totally agree, I actually made a migration for `@substrate/dev`, I just need to send the PR up. 

I put this one up first to be kind of a source of truth for the other PR's and as an example, but moving forward trying `yarn berry` out on some less substantial/critical repo's will probably be our best bet to make sure we work through whatever kinks we can find or workflow issues. 

> Maybe it's better to do each project one by one?

Couldnt agree more!

I think the beauty of the way the migration works as well is it is very easy to go backwards to our original setup. For example, if we find that `yarn berry` doesnt fit within X project, we just delete the `.yarn*` files and continue to use yarn as we did earlier.

TarikGul:
Solved the `jest` and `eslint` issue. Once this PR [#24](https://github.com/paritytech/substrate-js-dev/pull/24) is merged than I will be able to configure the `package.json` to correctly call the supporting packages from inside of `@substrate/dev`

package.json
```diff
@@ -53,9 +53,10 @@
     "@types/http-errors": "^1.6.3",
     "@types/morgan": "^1.9.2",
     "@types/triple-beam": "^1.3.2",
+    "eslint": "^7.24.0",
+    "jest": "^26.6.3",
```
emostov, line 57: These can be removed now, right?

emostov (approved):

package.json
TarikGul: Yup, exactly!

TarikGul: Once we make a new release and update `@substrate/dev` I will also adjust the scripts for `lint` and `test` etc.

.yarnrc.yml
```diff
@@ -0,0 +1,3 @@
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-2.4.1.cjs
```
emostov, line 3: I can't see the file yarn-2.4.1.cjs

What is in there? Can it be stored in substrate-js-dev so we can just update it in one place?

TarikGul: I can see the file name in the files changed tab but, there is nothing as well under it for me. But if you pull down the branch the script will show up. Just checked it out. So it must be some kind of github characteristic. 

> Can it be stored in substrate-js-dev so we can just update it in one place?

I am not sure, but I would caution on the side of we probably shouldn't. Since yarn 2.x was built to be a per repo basis package manager, I am not sure how it would interact without having the script locally and what kind of side effects it could cause.

dvdplm (approved):

PR #531, opened 2021-05-02
fix: remove dependabot and add upgrade-interactive

TarikGul:
@emostov @dvdplm, this might be a good opportunity to add the plugin `upgrade-interactive`. Being that Dependabot doesn't have yarn 2.x support, so this plugin might be the next best thing. 

Per every monday's release a simple `yarn upgrade-interactive` shows all the packages that have available upgrades via a interactive CLI. Seeing that `@polkadot/*` packages are also using this it might be a good opportunity to try it. In this [PR #518](https://github.com/paritytech/substrate-api-sidecar/pull/518) we had decided to leave it out because we couldn't see any immediate purpose, but now with the move to remove dependabot this could be a great replacement. 

Let me know what you think, also happy to remove it I was looking for alternative's that could suffice and this seems like the best one.


emostov (approved):

niklasad1 (approved):
LGTM but should be possible to add a scheduled `github action` for this to run every monday.

For example we have one in [jsonrpsee](https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/benchmarks.yml)

EDIT: doesn't probably work "interactive" ^^

maciejhirsz (approved):

Issue #491, opened 2021-03-25
Assets API for Statemint


joepetrowski:
### `GET assets/{assetId}/asset-info`
```
{
  at: { ... },
  AssetInfo: https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L186
  AssetMetadata: https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L258
}
```

### `GET accounts/{accountId}/assets-balances?asset=assetId`
```
WITH param
{
  at: { ... },
  balance: Balance,
  isFrozen: bool
 // https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L248
}

WITHOUT param
{
  at: { ... },
  assets: [{ assetId: int, balance: Balance, isFrozen: bool}]
}
```

### `GET accounts/{accountId}/asset-approvals?asset=assetId&delegate=accountId
```
{
  at: { ... },
  amount: Balance, // amount approved for transfer
  deposit: Balance // amount owner has on deposit to store the approval
}
```

TarikGul:
@joepetrowski Picked this up again, and have now made it priority. Expect to see some heavy progress on this in the next few days.

PR #533, opened 2021-05-04
feat: assets endpoints

TarikGul:
# Assets Endpoints 

closes: [#491](https://github.com/paritytech/substrate-api-sidecar/issues/491)
closes: [#545](https://github.com/paritytech/substrate-api-sidecar/issues/545)

## `GET assets/{assetId}/asset-info`

Associated Controller: `AssetsController`
Associated Service: `AssetsService`
Response: 
```
{
    at: { ... }
    assetInfo: {
        owner: AccountId,
        issuer: AccountId,
        admin: AccountId,
        freezer: AccountId,
        supply: TAssetBalance, // U64
        deposit: TAssetDepositBalance, // BalanceOf
        minBalance: TAssetBalance,
        isSufficient: bool, // polkadot-js way of saying boolean
        accounts: U32,
        sufficients: U32,
        approvals: U32,
        isFrozen: bool
    }
    assetMetadata: {
        deposit: TAssetDepositBalance,
        name: Bytes, // Polkadot-js type
        symbol: Bytes,
        decimals: U8,
        isFrozen: bool
    }
}
```
AssetInfo(AssetDetails): [substrate ref](https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L186) | [polkadot-js ref](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/assets/types.ts#L34)
AssetMetaData: [substrate ref](https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L258) | [polkadot-js ref](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/assets/types.ts#L50)

## `GET accounts/{accountId}/asset-balances?assets=assetId`

Notes: 
1. Originally per the issue filed above, there were two different proposed responses depending on whether an `assetId` is provided as a parameter. I changed it slightly to always have one type of response. The only difference is when a assetId is provided the `assets` array that is returned will only have the values provided in the query param. This choice was made to simplify return values for the end user. Happy to change this or discuss suggestions.

2. `assetId` query parameter is treated as an array. It is formatted as such: `assetId=id1,id2,id3` 

Associated Controller: `AccountsAssetsController`
Associated Service: `AccountsAssetsService`
Response: 
```
{
    at: { ... },
    assets: [
        {
            assetId: AssetId,
            balance: TAssetBalance,
            isFrozen: bool, 
            isSufficient: bool
        }
        ...
    ]
}
```

AssetBalance: [substrate ref](https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L248) | [polkadot-js ref](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/assets/types.ts#L20)

## `GET accounts/{accountId}/asset-approvals?assetId=assetId&delegate=accountId`

Note: Both a `delegate` and `assetId` are required for this endpoint. The `owner` => `accoundId` and the `delegate` can be referred to as the `AssetApprovalKey` in polkadot-js or `ApprovalKey` in substrate.

Associated Controller: `AccountsAssetsController`
Associated Service: `AccountsAssetsService`
Response: 
```
{
    at: { ... },
    amount: TAssetBalance,
    deposit: TAssetDepositBalance
}
```

AssetApproval: [substrate ref](https://github.com/paritytech/substrate/blob/master/frame/assets/src/types.rs#L79) | [polkadot-js](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/assets/types.ts#L8)
AssetApprovalKey: [substrate ref](https://github.com/paritytech/substrate/blob/master/frame/assets/src/types.rs#L70) | [polkadot-js](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/assets/types.ts#L14)

## TODO's

- [X] Asset Endpoints
- [X] Unit tests
- [X] Manually test all cases against statemint
    - [X] `assets/:assetId/asset-info`
    - [X] `accounts/:address/asset-balances`
    - [X] `accounts/:address/asset-balances?assets[]=10&assets[]=1`
    - [X] `accounts/:address/asset-approvals` (Fails correctly, this is subject to change)(`delegate`, and `assetId` might be part of the path)
    - [X] `accounts/:address/asset-approvals?assetId=assetId&delegate=delegateId`
- [X] Docs 
    - [X] inline docs
    - [X] swagger ui docs
    
## Other

1. Important to note I added statemint's metadata to the metadata library. It was the only way for me to have access to the `assets` methods inside of testing.


src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+				return assetBalanceFactory[20];
+			case 30:
+				return assetBalanceFactory[30];
+			default:
+				return;
+		}
+	});
+};
+
+assetsAccount.extend = function (args: AssetsAccount) {
```
TarikGul, line 653: This is currently my implementation to have a function and methods point to the same key. I think for the context and merit it provides it's alright. But I think it can be done better.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+			assetAccountStorageKeyThree,
+			assetAccountStorageKeyThree,
+		];
+	});
+
+/**
+ * @param assetId options are 10, 20, 30
+ */
+interface AssetsAccount {
+	// eslint-disable-next-line  @typescript-eslint/no-explicit-any
```
TarikGul, line 629: Same thing as mentioned below with the .extend property to add args to a function call.

src/test-helpers/typeFactory.ts
```diff
@@ -73,10 +74,14 @@ export class TypeFactory {
 	storageKey(
 		index: number,
 		indexType: keyof InterfaceTypes,
-		storageEntry: StorageEntryBase<'promise', GenericStorageEntryFunction>
+		storageEntry: StorageEntryBase<'promise', GenericStorageEntryFunction>,
+		secondaryId?: AccountId
```
TarikGul, line 78: This needs to be changed to an `Options?` where it takes in any secondary id to be used in the StorageKey. The only issue is when I wrote this I wasnt sure how to access all polkadot-js types within one type check. The naive way, and way I might have to do is take in a `keyof interfaces` and perform the logic inside of the function itself to allocate a secondary key or not.

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,115 @@
...
+
+		let response;
+		if (assets.length === 0) {
+			/**
+			 * This will query all assets and return them in an array
+			 */
+			const keys = await api.query.assets.account.keysAt(hash);
+			const assetIds = this.extractAssetIds(keys);
+
+			const queryAllAssets = async () =>
```
TarikGul, line 33: Should this be a helper function. Looks like its used twice in very similar ways.

src/controllers/AbstractController.ts
```diff
@@ -164,6 +164,26 @@ export default abstract class AbstractController<T extends AbstractService> {
 		return num;
 	}
```
emostov, line 166: We can note in docs that this expects array elements to be ints.

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	private getAssetBalances: RequestHandler = async (
+		{ params: { address }, query: { at, assets } },
+		res
+	): Promise<void> => {
+		const hash = await this.getHashFromAt(at);
+
+		let assetsArray: number[] = [];
+
```
emostov, line 30:
```suggestion
```

src/services/accounts/AccountsAssetsService.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+						balance: '20000000',
+						isFrozen: true,
+						isSufficient: true,
+					},
+				],
+			};
+
+			const response = await accountsAssetsService.fetchAssetBalances(
+				blockHash789629,
+				'0xffff',
```
emostov, line 35: Any significance to this number, or could it be any string for the sake of the test and the mock return values?

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+	 *
+	 * @param hash `BlockHash` to make call at
+	 * @param address `AccountId` associated with the balances
+	 * @param assets An array of `assetId`'s to be queried. If the length is zero
+	 * all assetId's associated to the account will be queried
+	 */
+	async fetchAssetBalances(
+		hash: BlockHash,
+		address: string,
+		assets: number[] | []
```
emostov, line 29:
```suggestion
		assets: number[]
```

src/services/accounts/AccountsAssetsService.ts
emostov: Yeah looks like both branches of the if statement have the same logic, just differ in mapping over assetIds vs assets.

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+
+		return {
+			at,
+			assets: response,
+		};
+	}
+
+	/**
+	 * Fetch all `AccountApproval`'s with a given `AssetId` and a `AssetApprovalKey`
+	 * which consists of a `delegate` and an `owner(AccoundId)`
```
emostov, line 91:
```suggestion
	 * which consists of a `delegate` and an `owner<AccoundId>`
```
? Not sure if you meant that as a type arg

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+			amount,
+			deposit,
+		};
+	}
+
+	/**
+	 * @param keys An Array of storage keys used to extract `AssetId`'s
+	 */
+	extractAssetIds(keys: StorageKey<[AssetId, AccountId]>[]): AssetId[] {
+		return keys.map(({ args: [assetId] }) => assetId).sort((a, b) => a.cmp(b));
```
emostov, line 135: Why do we need to sort?

src/services/assets/AssetsService.ts
```diff
@@ -0,0 +1,38 @@
...
+import { IAssetInfo } from '../../types/responses';
+import { AbstractService } from '../AbstractService';
+
+export class AssetsService extends AbstractService {
+	constructor(api: ApiPromise) {
+		super(api);
+	}
+
+	/**
+	 * Fetch an asset for its `AssetDetails` and `AssetMetadata` based on their `AssetId`.
```
emostov, line 13:
```suggestion
	 * Fetch an asset's `AssetDetails` and `AssetMetadata` with its `AssetId`.
```

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+	[key: string]: any;
+}
+
+const assetsAccount: AssetsAccount = (
+	assetId: number | AssetId,
+	_address: string
+) => {
+	const id = typeof assetId === 'number' ? assetId : assetId.toNumber();
```
emostov, line 637: Do we need to accept both types?

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+			case 30:
+				return assetBalanceFactory[30];
+			default:
+				return;
+		}
+	});
+};
+
+assetsAccount.extend = function (args: AssetsAccount) {
+	for (const i in args) {
```
emostov, line 654:
```suggestion
	for (const key in args) {
```
Just to make it more obvious that this is the key of the object because of `in`

src/services/test-helpers/mock/mockApi.ts
emostov: Whats preventing use from just hardcoding these fields?

src/test-helpers/typeFactory.ts
emostov: You could also create a second function for double map storage keys

emostov (commented):
Looks solid. Need to come back for a more thorough review.

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,66 @@
...
+		);
+	};
+
+	private getAssetApprovals: RequestHandler = async (
+		{ params: { address }, query: { at, delegate, assetId } },
+		res
+	): Promise<void> => {
+		const hash = await this.getHashFromAt(at);
+
+		if (!(typeof delegate === 'string') || !(typeof assetId === 'string')) {
```
maciejhirsz, line 50:
```suggestion
		if (typeof delegate !== 'string' || typeof assetId !== 'string') {
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+		let response;
+		if (assets.length === 0) {
+			/**
+			 * This will query all assets and return them in an array
+			 */
+			const keys = await api.query.assets.account.keysAt(hash);
+			const assetIds = this.extractAssetIds(keys);
+
+			const queryAllAssets = async () =>
+				await Promise.all(
```
maciejhirsz, line 44:
```suggestion
			const queryAllAssets = () =>
				Promise.all(
```

Equivalent but should be slightly cheaper one execution (assuming JIT wasn't compiling it away to begin with).

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+					isFrozen: asset.isFrozen,
+					isSufficient: asset.isSufficient,
+				};
+			});
+		} else {
+			/**
+			 * This will query all assets by the requested AssetIds
+			 */
+			const queryAllAssets = async () =>
+				await Promise.all(
```
maciejhirsz, line 62:
```suggestion
			const queryAllAssets = () =>
				Promise.all(
```

Ditto.

src/services/assets/AssetsService.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AssetsService } from './AssetsService';
+
+const assetsService = new AssetsService(mockApi);
+
+describe('AssetsService', () => {
+	describe('AssetsService.fetchAssetById', () => {
+		it('Should return the correct response for an AssetId', async () => {
+			const expectedResponse = {
+				at: {
+					hash:
```
maciejhirsz, line 12: That formatting looks wonky, is this done auto?

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+			case 10:
+				return assetBalanceFactory[10];
+			case 20:
+				return assetBalanceFactory[20];
+			case 30:
+				return assetBalanceFactory[30];
+			default:
+				return;
+		}
+	});
```
maciejhirsz, line 650: This doesn't look like it needs to be async at all?

```suggestion
	switch (id) {
		case 10:
			return Promise.resolve(assetBalanceFactory[10]);
		case 20:
			return Promise.resolve(assetBalanceFactory[20]);
		case 30:
			return Promise.resolve(assetBalanceFactory[30]);
		default:
			return Promise.resolve();
	}
```

Or just

```suggestion
	switch (id) {
		case 10:
			return assetBalanceFactory[10];
		case 20:
			return assetBalanceFactory[20];
		case 30:
			return assetBalanceFactory[30];
		default:
			return;
	}
```

Should be fine.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+};
+
+assetsAccount.extend = function (args: AssetsAccount) {
+	for (const i in args) {
+		// eslint-disable-next-line  @typescript-eslint/no-unsafe-assignment
+		this[i] = args[i];
+	}
+	return this;
```
maciejhirsz, line 658:
```suggestion
        return Object.assign(this, args);
```

Though that alone looks like code smell, would just use `Object.assign` over `x.extend` and drop the method entirely.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+};
+
+// eslint-disable-next-line  @typescript-eslint/no-unsafe-call
+assetsAccount.extend({
+	keysAt: assetsAccountKeysAt,
+});
```
maciejhirsz, line 664:
```suggestion
Object.assign(assetsAccount, {
	keysAt: assetsAccountKeysAt,
});
```

Comment above for context.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,159 @@ const auctionsWinningsAt = () =>
...
+
+const assetApprovals = () =>
+	Promise.resolve().then(() => {
+		const assetObj = {
+			amount: assetTBalanceOne,
+			deposit: balanceOfTwo,
+		};
+		return rococoRegistry.createType('Option<AssetApproval>', assetObj);
+	});
```
maciejhirsz, line 673:
```suggestion
const assetApprovals = async () => {
	const assetObj = {
		amount: assetTBalanceOne,
		deposit: balanceOfTwo,
	};
	return rococoRegistry.createType('Option<AssetApproval>', assetObj);
};
```

Does this need to return a promise?

maciejhirsz (commented):
Looks good! I've left some comments with grumbles, suggestions and questions, but I think all are just syntactic quirks.

src/services/test-helpers/mock/mockApi.ts
TarikGul: I agree this is an excellent solution instead. `Object.assign` escaped my mind when considering attaching keys to a Function Object. Thanks! Fixed with your solution.

src/services/test-helpers/mock/mockApi.ts
TarikGul: Fixed per comment below.

src/services/test-helpers/mock/mockApi.ts
TarikGul: Yup completely correct, it doesnt need to be async at all. Fixed

src/services/test-helpers/mock/mockApi.ts
TarikGul: Per below Maciej's suggesetion, it should just be a `Object.assign(assetsAccount, { keysAt: assetsAccountKeysAt})`

But yea we cant hardcode it as a key without the Function object first being created.

src/test-helpers/typeFactory.ts
TarikGul: Awesome, took your suggestion and made a `storageKeyDoubleMap` thanks!

src/services/test-helpers/mock/mockApi.ts
TarikGul: Removed this function and replaced with an `Object.assign`

src/services/test-helpers/mock/mockApi.ts
TarikGul: Yea because the function `fetchAssetBalances` uses `number` when the `assetId` is provided by the query parameter but then uses `AssetId` type when it's extracted from the keys.

src/services/accounts/AccountsAssetsService.ts
TarikGul: Yes and no, and now that I look back at it, it should probably more clear. Something like 'which consists of a `delegate` and an `owner`, and the owner is the paths accountId'

src/services/accounts/AccountsAssetsService.spec.ts
TarikGul: No significance, just a placeholder.

src/controllers/AbstractController.ts
```diff
@@ -164,6 +164,26 @@ export default abstract class AbstractController<T extends AbstractService> {
...
 
+	protected parseQueryParamArrayOrThrow(
+		n: string,
+		errorMessage: string
+	): number[] {
+		const isValidParam = /[^,]+/.test(n);
```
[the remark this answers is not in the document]
TarikGul: Yea I also realized this regex is right but wrong. It will pass for all cases basically which is not what I want. Ex: `aaa,aaa` => true `aaa,,,,a`  => true (should be false), and there are many more cases to, Going to rewrite this one.

Which is also a good reminder this method needs a test to confirm that.

src/services/assets/AssetsService.spec.ts
TarikGul: No its not auto-generated. It's similar structure I wrote out for `ParasService` tests [here](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/paras/ParasService.spec.ts).

Curious how do you think i could make the structure less wonky?

src/controllers/AbstractController.ts
```diff
@@ -164,6 +164,26 @@ export default abstract class AbstractController<T extends AbstractService> {
...
+
+		if (!isValidParam) {
+			throw new BadRequest(errorMessage);
+		}
+
+		return n
+			.split(',')
```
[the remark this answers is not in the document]
TarikGul: Yea so I noticed above as well with Maciej, but this regex works in practice but not what we want exactly. It will basically check for `,` and then thats it. It basically will always pass. Basically the idea I need to get with this regex it to make sure and confirm we can only have two possibilities. 

`12341,123412,8545` or `12313` and then if the Regex passes split it by the commas and check for numbers.

src/services/accounts/AccountsAssetsService.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+						balance: '20000000',
+						isFrozen: false,
+						isSufficient: false,
+					},
+					{
+						assetId: '30',
+						balance: '20000000',
+						isFrozen: false,
+						isSufficient: false,
+					},
```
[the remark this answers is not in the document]
TarikGul: This is a good point. In this case yes, but for practicality sake I am going to remove it so it doesnt confuse anyone because a real world scenario would have this. Instead I will make it just one.

src/services/accounts/AccountsAssetsService.ts
TarikGul: Fixed!

src/services/accounts/AccountsAssetsService.ts
TarikGul: The idea behind sorting is so that every time one queries all assets they can expect the same order of assets back. If you guys think this is unnecessary I can totally remove it. It was more for consistency's sake.

src/controllers/AbstractController.ts
```diff
@@ -164,6 +164,26 @@ export default abstract class AbstractController<T extends AbstractService> {
...
+			.map((str) =>
+				this.parseNumberOrThrow(
+					str,
+					`Incorrect AssetId format: ${str} is not a U32.`
```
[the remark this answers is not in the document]
TarikGul: Yea it should be fine. I decided to put the primitive type of the `assetId` in the error message for more accurate messaging instead of just saying `number`, I don't think an assetId above 2,147,483,647 will ever exist though. Do you think it would be better to replace the `U32` with number?

src/services/accounts/AccountsAssetsService.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+	describe('AccountsAssetsService.fetchAssetBalances', () => {
+		it('Should return the correct response with the assets param', async () => {
+			const expectedResponse = {
+				at,
+				assets: [
+					{
+						assetId: '10',
+						balance: '10000000',
+						isFrozen: false,
+						isSufficient: true,
```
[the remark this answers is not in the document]
TarikGul: isSufficient (via substrate docs): 

```rust
/// - `is_sufficient`: Whether a non-zero balance of this asset is deposit of sufficient
/// value to account for the state bloat associated with its balance storage. If set to
/// `true`, then non-zero balances may be stored without a `consumer` reference (and thus
/// an ED in the Balances pallet or whatever else is used to control user-account state
/// growth).
```

isFrozen: 

```rust
/// - `is_frozen`: Whether this asset class is frozen except for permissioned/admin
/// instructions.
```

src/types/responses/AccountAssets.ts
```diff
@@ -0,0 +1,29 @@
...
+	TAssetBalance,
+	TAssetDepositBalance,
+} from '@polkadot/types/interfaces';
+import { AssetId } from '@polkadot/types/interfaces/runtime';
+import { bool } from '@polkadot/types/primitive';
+
+import { IAt } from '.';
+
+/**
+ *
```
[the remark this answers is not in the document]
TarikGul: Yup per the issue above I added docs to all the types.

src/types/responses/AccountAssets.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export interface IAssetBalance {
+	assetId: AssetId | number;
+	balance: TAssetBalance;
+	isFrozen: bool;
+	isSufficient: bool;
+}
+
+export interface IAccountAssetsBalance {
```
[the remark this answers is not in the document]
TarikGul: fixed!

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,137 @@
...
+					assetIds.map((assetId: AssetId) =>
+						api.query.assets.account(assetId, address)
+					)
+				);
+			response = (await queryAllAssets()).map((asset, i) => {
+				return {
+					assetId: assetIds[i],
+					balance: asset.balance,
+					isFrozen: asset.isFrozen,
+					isSufficient: asset.isSufficient,
```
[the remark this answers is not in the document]
TarikGul: Most likely outdated dependencies. Both substrate and polkadot-js still have it.

src/services/accounts/AccountsAssetsService.spec.ts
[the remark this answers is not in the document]
TarikGul: Added the docs for all the types in this file.

src/controllers/AbstractController.ts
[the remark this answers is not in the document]
emostov: I would just say "is not a positive integer" since that seems the "most" correct in this case.

src/services/accounts/AccountsAssetsService.spec.ts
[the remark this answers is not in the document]
emostov: Probably over the top, but if in real life this shouldn't happen should we put an error check on the data from the node to make sure this is never the case?

src/services/accounts/AccountsAssetsService.ts
emostov: I think we should remove it. I don't see the utility and it sounds unnecessary to have in our user api contract (maybe even confusing at the conceptual level since it might lead people to think assetId is linked to some deeper meaning of the assets ordering)

src/services/test-helpers/mock/mockApi.ts
emostov: Yeah, since its mocking a response to a storage query the code expects a `Promise`

TarikGul: No it technically doesn't, but I do it here to keep the format the same as the rest of the functions affiliated with the mockApi in the file.

src/controllers/AbstractController.ts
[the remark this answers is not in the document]
TarikGul: Fixed!

src/services/accounts/AccountsAssetsService.ts
TarikGul: Awesome removed :)

src/services/accounts/AccountsAssetsService.spec.ts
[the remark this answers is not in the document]
TarikGul: Yea we could do that, not sure it's totally necessary, If that were to happen, I would assume that would be a serious bug on the substrate or polkadot-js end. 

One thing to potentially filter as well though could be Assets that have a balance of zero, but that should also be up to the end user.

src/controllers/AbstractController.ts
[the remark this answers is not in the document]
emostov: Its of trying to hand roll our own regex what if we used the express syntax for query param arrays? See http://expressjs.com/en/api.html#req.query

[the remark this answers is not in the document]
TarikGul: Yea i considered it at first, and now that you mention it again I think it's probably the best route we have considering my lack of regex knowledge. Will change it to reflect the built in query support

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,65 @@
...
+		{ params: { address }, query: { at, assets } },
+		res
+	): Promise<void> => {
+		const hash = await this.getHashFromAt(at);
+
+		let assetsArray: number[] = [];
+		if (typeof assets === 'string') {
+			assetsArray = this.parseQueryParamArrayOrThrow(
+				assets,
+				'assets query parameter is not in the correct format.'
```
emostov, line 33:
```suggestion
				'`assets` query parameter is not in the correct format. Should be of the form `?assets[]=42&assets[]=3`'
```
Maybe? It at least gives a clue to the person reading the code, but maybe we should just say to reference the docs.

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,65 @@
...
+			);
+		}
```
emostov, line 35:
```suggestion
		const assetsArray = Array.isArray(assets) 
		     ? this.parseQueryParamArrayOrThrow(
				assets,
				'assets query parameter is not in the correct format.'
			)
		    : [ ];	
```
1) Assumes when we update the query param syntax it gives us an array
2) Ternaries are fun! But I don't think its a must :)

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	private getAssetApprovals: RequestHandler = async (
+		{ params: { address }, query: { at, delegate, assetId } },
+		res
+	): Promise<void> => {
+		const hash = await this.getHashFromAt(at);
+
+		if (typeof delegate !== 'string' || typeof assetId !== 'string') {
+			throw new BadRequest(
+				'Must include a delegate accountId as a query param'
```
emostov, line 51:
```suggestion
				'Must include a `delegate` and `assetId` query param'
```

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,65 @@
...
+		);
+	};
+
+	private getAssetApprovals: RequestHandler = async (
+		{ params: { address }, query: { at, delegate, assetId } },
+		res
+	): Promise<void> => {
+		const hash = await this.getHashFromAt(at);
+
+		if (typeof delegate !== 'string' || typeof assetId !== 'string') {
```
emostov, line 49: If both are mandatory, should they be in the path? @joepetrowski 

`/accounts/:address/asset-approvals/{assetId}/delegate/{delegateId}`
or
`/accounts/:address/asset-approvals/{assetId}-{delegateId}`

src/controllers/assets/AssetsController.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default class AssetsController extends AbstractController<AssetsService> {
+	constructor(api: ApiPromise) {
+		super(api, '/assets', new AssetsService(api));
+		this.initRoutes();
+	}
+
+	protected initRoutes(): void {
+		this.safeMountAsyncGetHandlers([
+			['/:assetsId/assetInfo', this.getAssetById],
```
emostov, line 15:
```suggestion
			['/:assetsId/asset-info', this.getAssetById],
```
kebab-case is consistent with all out other URIs

src/controllers/assets/AssetsController.ts
```diff
@@ -0,0 +1,38 @@
+import { ApiPromise } from '@polkadot/api';
+import { RequestHandler } from 'express';
+
+import { AssetsService } from '../../services';
+import AbstractController from '../AbstractController';
+
+export default class AssetsController extends AbstractController<AssetsService> {
+	constructor(api: ApiPromise) {
+		super(api, '/assets', new AssetsService(api));
```
emostov, line 9: `/pallets/assets/:assetId/asset-info` @joepetrowski ?

src/services/accounts/AccountsAssetsService.spec.ts
emostov, line 35:
```suggestion
				'0xffff', // AccountId arg here does not affect the test results
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,145 @@
...
+	IAccountAssetApproval,
+	IAccountAssetsBalances,
+	IAssetBalance,
+} from '../../types/responses';
+import { AbstractService } from '../AbstractService';
+
+export class AccountsAssetsService extends AbstractService {
+	constructor(api: ApiPromise) {
+		super(api);
+	}
```
emostov, line 17:
```suggestion
```
Implicit if its just a straight `super`

src/test-helpers/typeFactory.ts
```diff
@@ -81,6 +81,34 @@ export class TypeFactory {
 		return key.setMeta(storageEntry.creator.meta);
 	}
 
+	/**
+	 * Similar to the storageKey method above, this is specific to storageKeys
+	 * that take in double map for keys.
+	 *
+	 * @param index The id to assign the key to.
```
emostov, line 88: What do you mean by `id` here? As in what is `id` to the keys of a double map?

src/test-helpers/typeFactory.ts
```diff
@@ -81,6 +81,34 @@ export class TypeFactory {
...
+	 * @param indexTypeOne The First InterfaceType that will be used to create the index into its new appropriate index type
```
emostov, line 89: Does index === map key?

src/test-helpers/typeFactory.ts
```diff
@@ -81,6 +81,34 @@ export class TypeFactory {
...
+	storageKeyDoubleMap(
+		index: number,
+		indexTypeOne: keyof InterfaceTypes,
+		indexTypeTwo: keyof InterfaceTypes,
+		storageEntry: StorageEntryBase<'promise', GenericStorageEntryFunction>
+	): StorageKey {
+		const idOne = this.#registry.createType(indexTypeOne, index);
+		const idTwo = this.#registry.createType(indexTypeTwo);
+
+		const key = new StorageKey(this.#registry, storageEntry.key(idOne, idTwo));
```
emostov, line 107:
```suggestion
		const keys = new StorageKey(this.#registry, storageEntry.key(idOne, idTwo));
```
Maybe? Confusing though since its `StorageKey` singular

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -517,6 +519,151 @@ const auctionsWinningsAt = () =>
 		return optionWinnings;
 	});
 
+/**
+ * Asset specific constants.
+ * Note: It borrows some variables used in the parachains constant section
+ *
+ * Used in `/assets` and `/accounts` endpoints
+ */
```
emostov, line 527: Anyone have an opinion on if we should break some fo the below out into a new file for organization?

TarikGul: I was thinking the same thing. The file itself is starting to get bloated with endpoint specific functions. I am onboard for moving them out.

src/controllers/accounts/AccountsAssetsController.ts
TarikGul: I like this as well, I think what you have suggested is good, then we can tag on at the and 'reference the docs for more information.'

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,141 @@
...
+		const { api } = this;
+
+		const { number } = await api.rpc.chain.getHeader(hash);
+
+		let response;
+		if (assets.length === 0) {
+			/**
+			 * This will query all assets and return them in an array
+			 */
+			const keys = await api.query.assets.account.keysAt(hash);
```
emostov, line 39: I don't fully know how this works but I am worried that this 1) be a huge query and 2) a lot to loop over. If we just want all the available assetIds why not just get the keys of [Asset](https://github.com/paritytech/substrate/blob/462653636821f542fd99beed25e817b7f14c38c9/frame/assets/src/lib.rs#L218-L222) or even [Metadata](https://github.com/paritytech/substrate/blob/462653636821f542fd99beed25e817b7f14c38c9/frame/assets/src/lib.rs#L252-L257)

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	async queryAssets(
+		api: ApiPromise,
+		assets: AssetId[] | number[],
+		address: string
+	): Promise<IAssetBalance[]> {
+		const queryInputAssets = async (): Promise<AssetBalance[]> =>
+			await Promise.all(
+				assets.map((assetId: number | AnyNumber) =>
+					api.query.assets.account(assetId, address)
```
emostov, line 119: Instead of the extra loop below could we make this an `async` call back that like:

```javascript
Promise.all(
   assets.map(async (assetId) => {
      const assetBalance = api.query.assets.account(assetId, address);
     // could we just have `return assetBalance` - unclear to me what the below adds and if it misses `extra`
      return { 
   	            assetId,
		    balance: assetBalance.balance,
		    isFrozen: assetBalance.isFrozen,
		    isSufficient: assetBalance.isSufficient,    
     }
   }

File: src/services/accounts/AccountsAssetsService.ts [hunk 43]
```diff
@@ -0,0 +1,141 @@
...
+				)
+			);
+
+		const response = (await queryInputAssets()).map((asset, i) => {
+			return {
```
emostov, line 124: Why do we need to reformat `AssetBalance` like this? Should we just return the struct as is?

emostov (commented):
After a more thorough review and reading through the assets pallet I think we should have some discussion on 
1) should `/assets` go under `/pallets`?
2) `/accounts/:address/asset-approvals/{assetId}/delegate/{delegateId}`?
3) should we include every field of `AssetBalance` in the response in `/accounts/:address/asset-balances`?

src/controllers/AbstractController.ts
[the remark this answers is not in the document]
TarikGul: removed the regex, and replaced it with express's query parameter for formatting arrays

src/test-helpers/typeFactory.ts
TarikGul: Yea I think a singluar `key` here makes sense because we are instantiating a single StorageKey object.

src/services/accounts/AccountsAssetsService.ts
TarikGul: If we return the `Struct` as is there is no AssetId attached to it.

src/test-helpers/typeFactory.ts
TarikGul: I changed this up to reflect its actual meaning. You're correct that it is confusing. The `index` is just the value assigned to `indexTypeOne`.

TarikGul: So I changed the name to value, and referenced what it is in the docs

src/test-helpers/typeFactory.ts
TarikGul: `index` is the value of `indexTypeOne` changed the name to `value`

src/test-helpers/typeFactory.ts
```diff
@@ -81,6 +81,34 @@ export class TypeFactory {
...
+	 * Some Parameter Examples:
+	 * 1. apiPromise.query.assets.account
+	 */
+	storageKeyDoubleMap(
+		index: number,
+		indexTypeOne: keyof InterfaceTypes,
+		indexTypeTwo: keyof InterfaceTypes,
+		storageEntry: StorageEntryBase<'promise', GenericStorageEntryFunction>
+	): StorageKey {
+		const idOne = this.#registry.createType(indexTypeOne, index);
```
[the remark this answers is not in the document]
TarikGul: The here is wrong, I had just named it because the doubleMap i was directly building this for was `AssetId`, and `AccountId` so i naturally wrote `id` switching it to `mapValueOne` and `mapValueTwo`

src/services/accounts/AccountsAssetsService.ts
TarikGul: If we did it this way, I dont think we would need to wrap it in the `Promise.all`. The purpose for the map here is to create all the calls for the `Promise.all`. My whole idea was trying to avoid making individual queries to the chain, which is why that logic below had to be added as well.

src/services/accounts/AccountsAssetsService.ts
TarikGul: This is actually the same way that polkadot-js-ui queries all assets, thats where I got the idea from. I think it could definitely be costly. But I am not sure the above example with `Asset` and `Metadata` are correct. I might be wrong but that would only be for a single asset. 

Here is where and how its implemented in Polkadot-js apps. [useMapKeys](https://github.com/polkadot-js/apps/blob/master/packages/react-hooks/src/useMapKeys.ts#L16) is the hook that calls the `keysAt`, and then [useMapKeys is called here](https://github.com/polkadot-js/apps/blob/master/packages/page-assets/src/useAssetIds.ts#L19) inside of `useAssetIds`.

src/services/assets/AssetsService.ts
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	/**
+	 * Fetch an asset's `AssetDetails` and `AssetMetadata` with its `AssetId`.
+	 *
+	 * @param hash `BlockHash` to make call at
+	 * @param assetId `AssetId` used to get info and metadata for an asset
+	 */
+	async fetchAssetById(hash: BlockHash, assetId: number): Promise<IAssetInfo> {
+		const { api } = this;
```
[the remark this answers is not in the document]
TarikGul: Yea I have actually grown to like it. Im indifferent though too.

src/services/assets/AssetsService.spec.ts
maciejhirsz: Put the value on the same line as the key? :)

TarikGul: Ahh, i see what you are saying, yea definitely a little wonky. That's just the linter in this case :).

src/services/accounts/AccountsAssetsService.ts
emostov: I see what you are saying, your original logic mimics the app exactly afaict. I just don't understand why thats correct.

From my understanding we need to:
1) Get all `assetIds`.
2) Check every single `assetId` and see if the user has them.

Every `assetId` is stored in `Asset`. If you look at the `create` call you can [see the `assetId` is always inserted into `Assets`](https://github.com/paritytech/substrate/blob/462653636821f542fd99beed25e817b7f14c38c9/frame/assets/src/lib.rs#L381-L396).

Operating under the assumption that `keys()` will return all the combination of keys on a double map and not just the first key: in a single map I would expect that number of keys to be way less (#keys == #assets). On the other hand `Account` is a double map with 2 keys per entry; so the total of amount of keys could be much more and is always guaranteed to be a super set of the keys in `Account` (#keys == #assets * #accountIds)

src/services/accounts/AccountsAssetsService.ts
emostov: Makes sense, then for `extra` we can do something like `extra: asset.extra ?? undefined`, but need to figure out the appropriate way to check what `extra` is

emostov: I would prefer if we treated the pallet as generically as possible and not make too many assumptions about the runtime it lives in.

src/services/accounts/AccountsAssetsService.ts
emostov: It would still need to be wrapped in a `Promise.all` bc the async callback itself returns a `Promise` and map will not wait for the promise to resolve. Query complexity should not increase since they will still be parallel-ized.

src/services/accounts/AccountsAssetsService.ts
TarikGul: Yup you are 100% right. I changed the call to be `api.query.assets.asset.keysAt(hash)` which will only go through a Single Map StorageKey. Good catch and thanks for pointing that out.

TarikGul: Resolved thanks!

src/services/accounts/AccountsAssetsService.ts
TarikGul: Awesome, I see what you are saying. Fixed a refactored, thanks.

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,141 @@
...
+		 * delegate `accountId`. Both are required for a successfull query
+		 */
+		const approvalKey = { owner: address, delegate };
+
+		const [{ number }, assetApproval] = await Promise.all([
+			api.rpc.chain.getHeader(hash),
+			api.query.assets.approvals(assetId, approvalKey),
+		]);
+
+		if (assetApproval.isNone) {
```
TarikGul, line 88: @emostov for an asset-approval that doesnt exist, do you think we should throw an error, or just give null fields?

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
 - name: transaction
 - name: paras
 paths:
+  /accounts/{accountId}/asset-balances:
+    get:
+      tags:
+      - accounts
+      summary: Get an array of asset-balances for an account.
+      description: Returns information about an accounts asset-balances. This is 
+        specific to the assets-pallet for parachains. If no `assets` query parameter
```
emostov, line 29:
```suggestion
        specific to the assets pallet for parachains. If no `assets` query parameter
```

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+        schema:
+          type: string
+          description: Block height (as a non-negative integer) or hash
+            (as a hex string).
+          format: unsignedInteger or $hex
+      - name: assets
+        in: query
+        description: An array of AssetId's to be queried. If not inputed by default
+          all asset-balances associated with the `accountId` will be returned. The array
+          format follows Express 4.x API recommendations. ex:`?assets[]=1&assets[]=2&assets[]=3`.
```
emostov, line 54:
```suggestion
          format follows Express 4.x API. ex:`?assets[]=1&assets[]=2&assets[]=3`.
```

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+          specified account.
+        required: false
+        schema:
+          type: string
+          description: Block height (as a non-negative integer) or hash
+            (as a hex string).
+          format: unsignedInteger or $hex
+      - name: assets
+        in: query
+        description: An array of AssetId's to be queried. If not inputed by default
```
emostov, line 52:
```suggestion
        description: An array of AssetId's to be queried. If not supplied, defaults to providing
```

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+          all asset-balances associated with the `accountId` will be returned. The array
```
emostov, line 53:
```suggestion
          all asset balances associated with the `accountId`. The array query param
```
?

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+          description: successfull operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccountAssetsBalances'
+  /accounts/{accountId}/asset-approvals:
+    get:
+      tags:
+      - accounts
+      summary: Get an asset-approval for an account.
```
emostov, line 71:
```suggestion
      summary: Get an asset approval for an account.
```

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+      description: Returns information about an account's asset-approval transaction.
```
emostov, line 72:
```suggestion
      description: Returns information about an account's asset approval transaction.
```

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+        It is required to pass in a delegate and an assetId as query parameters.
```
emostov, line 73: Are these line breaks working when displaying the UI? I forget exactly but sometimes these line breaks don't play way with openapi yaml.
If not you can do:
```
description: |
    80 chars of text
    more text

File: docs/src/openapi-v1.yaml [hunk 54]
```diff
@@ -20,6 +20,98 @@ tags:
...
+          type: string
+          description: An assetId represented as an unsignedInteger.
+          format: unsignedInteger
+      - name: delegate
+        in: query
+        description: The delegate's `accountId` associated with an asset-approval.
+        required: true
+        schema:
+          type: string
+          pattern: '^[1-9A-HJ-NP-Za-km-z]{8,64}$'
```
emostov, line 107: @dvdplm does this look like an ok regex?

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,108 @@
+import { ApiPromise } from '@polkadot/api';
+import { RequestHandler } from 'express';
+import { BadRequest } from 'http-errors';
+
+import { validateAddress } from '../../middleware';
+import { AccountsAssetsService } from '../../services/accounts';
+import AbstractController from '../AbstractController';
+
+/**
+ * Get asset information for an address.
```
emostov, line 10: These are nice but in the future you don't have to do them. They are legacy from before we had the openapi docs.

src/services/accounts/AccountsAssetsService.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+		it('Should return the correct response', async () => {
+			const expectedResponse = {
+				at,
+				amount: '10000000',
+				deposit: '2000000',
+			};
+
+			const response = await accountsAssetsService.fetchAssetApproval(
+				blockHash789629,
+				'', // AccountId arg here does not affect the test results
```
emostov, line 87: 👍

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+
+		const [{ number }, assetApproval] = await Promise.all([
+			api.rpc.chain.getHeader(hash),
+			api.query.assets.approvals(assetId, approvalKey),
+		]);
+
+		let amount = null,
+			deposit = null;
+		if (assetApproval.isSome) {
+			({ amount, deposit } = assetApproval.unwrap());
```
emostov, line 90: Did the linter put these extra parenthesis are this line?

src/types/responses/AccountAssets.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Response from `/accounts/{accountId}/assets-balances?asset=assetId`
+ */
+export interface IAccountAssetsBalances {
+	at: IAt;
+	assets: IAssetBalance[];
+}
+
+/**
+ * Response from `/accounts/{accountId}/asset-approvals?asset=assetId&delegate=accountId`
```
emostov, line 42:
```suggestion
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+
+		const at = {
+			hash,
+			height: number.unwrap().toString(10),
+		};
+
+		return {
+			at,
+			amount,
+			deposit,
```
emostov, line 101:
```suggestion
			assetApproval: assetApproval.isSome
			    ? assetApproval.unwrap() // assuming this gets serialized to `{ amount, deposit }`
			    : null 
```
Right now if `assetApproval.isNone` it looks like `amount` and `deposit` will be undefined. I am thinking if we put in under this key we can better communicate "no approval exists"

emostov (approved):
Looks good! I just have one suggestion for how to change the api for `fetchAssetApproval` that I think will improve readability of the `None` response. (And tiny recs in docs)

src/chains-config/kusamaControllers.ts
```diff
@@ -8,6 +8,7 @@ export const kusamaControllers: ControllerConfig = {
 	controllers: {
 		Blocks: true,
 		BlocksExtrinsics: true,
+		AccountsAssets: true,
```
dvdplm, line 11: Ideally we should just state which controllers we want and assume that all controllers not explicitly mentioned are not available. So it'd be `controllers = [Blocks, BlocksExtrinsics, AccountAssets, …, …];`. The controllers for kilt, dock, kulupu etc are getting pretty boilerplatey.

src/controllers/AbstractController.ts
```diff
@@ -164,6 +164,24 @@ export default abstract class AbstractController<T extends AbstractService> {
...
 
+	protected parseQueryParamArrayOrThrow(
+		n: string[],
+		errorMessage: string
+	): number[] {
+		const isValidArray = Array.isArray(n);
+
+		if (!isValidArray) {
+			throw new BadRequest(errorMessage);
+		}
```
dvdplm, line 175:
```suggestion
		if (!Array.isArray(n)) {
			throw new BadRequest(errorMessage);
		}
```

…but really I would have expected TS to handle this. You sure you can call this with a non-array of strings?

src/controllers/accounts/AccountsAssetsController.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Paths:
+ * - `address`: The address to query
+ *
+ * Query:
+ * - (Optional)`at`: Block at which to retrieve runtime version information at. Block
+ *  	identifier, as the block height or block hash. Defaults to most recent block.
+ * - (Optional for `/accounts/:address/asset-balances`)`assets`
+ * - (Not-Optional for `/accounts/:address/asset-approvals)`assetId` The assetId associated
```
dvdplm, line 19: s/Non-Optional/Required/ ?

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+import {
+	IAccountAssetApproval,
+	IAccountAssetsBalances,
+	IAssetBalance,
+} from '../../types/responses';
+import { AbstractService } from '../AbstractService';
+
+export class AccountsAssetsService extends AbstractService {
+	/**
+	 * Fetch all the `AssetBalance`s along side their `AssetId`'s for a given array of queried AssetId`'s.
```
dvdplm, line 14:
```suggestion
	 * Fetch all the `AssetBalance`s alongside their `AssetId`'s for a given array of queried `AssetId`'s.
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+	 * If none are queried the function will get all AssetId's associated with the
```
dvdplm, line 15:
```suggestion
	 * If none are queried the function will get all `AssetId`'s associated with the
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+		return {
+			at,
+			amount,
+			deposit,
+		};
+	}
+
+	/**
+	 * Takes in an array of AssetId's, and an AccountId and returns
+	 * all balances tied to those AssetId's.
```
dvdplm, line 107:
```suggestion
	 * Takes in an array of `AssetId`s, and an `AccountId` and returns
	 * all balances tied to those `AssetId`s.
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+	 *
+	 * @param api ApiPromise
+	 * @param assets An Array of `AssetId`'s or numbers representing assetId's
```
dvdplm, line 110:
```suggestion
	 * @param assets An Array of `AssetId`s or numbers representing `assetId`s
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -0,0 +1,138 @@
...
+					balance: assetBalance.balance,
+					isFrozen: assetBalance.isFrozen,
+					isSufficient: assetBalance.isSufficient,
+				};
+			})
+		);
+	}
+
+	/**
+	 * @param keys An Array of storage keys used to extract `AssetId`'s
```
dvdplm, line 133:
```suggestion
	 * @param keys Extract `assetId`s from an array of storage keys
```

src/services/test-helpers/mock/mockApi.ts
dvdplm: I'm fine either way but I don't mind long test files generally. They are typically not read top to bottom anyway?

src/types/responses/AccountAssets.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { IAt } from '.';
+
+export interface IAssetBalance {
+	/**
+	 * Identifier for the class of asset.
+	 */
+	assetId: AssetId | number;
+	/**
+	 * The units in which substrate record's balances.
```
dvdplm, line 16:
```suggestion
	 * The units in which substrate records balances.
```

src/services/accounts/AccountsAssetsService.ts
TarikGul: So actually in order to destruct an object and assign it to let variables the destructuring needs to be wrapped in paranthesis. 

```
Destructuring needs to be either after a let, const or var declaration or it needs to be in an expression context to distinguish it from a block statement.
```

Per an issue I found on the topic.

src/controllers/accounts/AccountsAssetsController.ts
TarikGul: Good to know :). Yea i noticed them on the other controllers, and thought it wouldn't hurt.

src/services/accounts/AccountsAssetsService.ts
TarikGul: So I actually assign both `amount` and `deposit` to null above, before doing the `isSome` check.

src/controllers/AbstractController.ts
dvdplm: No, what I was referring to is that that the compiler should be able to infer any calls that do not comply with the signature, `string[]`? If it can do that then we can know that it's an array and we don't need to check at all.

docs/src/openapi-v1.yaml
```diff
@@ -20,6 +20,98 @@ tags:
...
+      - name: assets
+        in: query
+        description: An array of AssetId's to be queried. If not inputed by default
+          all asset-balances associated with the `accountId` will be returned. The array
+          format follows Express 4.x API recommendations. ex:`?assets[]=1&assets[]=2&assets[]=3`.
+        required: false
+        schema:
+          type: string
+          description: An array of assetId numbers represented as strings
+          format: Array of unsignedInteger's
```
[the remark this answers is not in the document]
TarikGul: Yup this works. Just changed it!

src/controllers/AbstractController.ts
emostov: Since we are taking use input via query param I think it makes sense to do a runtime verification

TarikGul: Ahh yes, I see what you are saying. That was a mis-judgement on my part. Yea this is definitely an unnecessary check, and we can remove it. Now that i look back the reason I had this was because the `res.query` has 4 types that it can be returned as. 

`QueryString.ParsedQs | string | string[] | QueryString.ParsedQs[]`

therefore before I was checking to make sure the input was an array, where now its redundant because we check that before we even plug it in. Great catch!

docs/src/openapi-v1.yaml
TarikGul: Awesome in this case, i switched from `pattern: <some-regex>` to `format: SS58`. Per my conversation with @emostov it seems as if regex is not the best way to verify SS58's pattern via the docs but instead to just specifiy it's format.

TarikGul:
This PR is ready to merge. Just manually test against Statemint today again. In return we found out the version of polkadot needed to be updated in order to resolve a few more types. I also changed all the `pattern` fields for `accountId` in the docs to `format: SS58` in order to close [#545](https://github.com/paritytech/substrate-api-sidecar/issues/545).

PR #535, opened 2021-05-06
fix: Bump polkadot-js/api to get parachain types: Document user contract regressions

emostov:
### NOTE 
~~need to hold off on merging until I figure out how to get a valid type registry. The one of the derive block is not working~~ Using the registry from `Vec<EventRecord>` now, which seems to work.

### Changes

- Bump polkadot-js/api => 4.9.2
- Document changes to serialization of `ConsensusEngineId` (see specs)
- Use `derive.chain.getBlock` to get both authorId and block. We remove the query to `session.validators` and the `authorId` extraction function because it looks like something about the new `ConsensusEngineId` serialization broke that.

We should also check to see if there is any significant performance regression from the new derive.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -323,7 +323,7 @@ export class BlocksService extends AbstractService {
 		block: Block,
 		events: Vec<EventRecord> | string,
 		extrinsicDocs: boolean
-	) {g
+	) {
 		const defaultSuccess = typeof events === 'string' ? events : false;
 		// Note, if events is a string then there was an issue getting them from the node.
 		// In this case we try and create the calls with the registry on `block`
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -3,12 +3,10 @@ import { expandMetadata } from '@polkadot/metadata/decorate';
 import { Compact, GenericCall, Struct } from '@polkadot/types';
 import { AbstractInt } from '@polkadot/types/codec/AbstractInt';
 import {
-	AccountId,
 	Block,
 	BlockHash,
 	BlockNumber,
 	BlockWeights,
-	Digest,
 	DispatchInfo,
 	EventRecord,
 	Hash,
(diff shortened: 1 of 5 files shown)


niklasad1 (approved):
LGTM

src/services/blocks/BlocksService.ts
```diff
@@ -339,15 +321,23 @@ export class BlocksService extends AbstractService {
...
+		events: Vec<EventRecord> | string,
 		extrinsicDocs: boolean
 	) {
 		const defaultSuccess = typeof events === 'string' ? events : false;
+		// Note, if events is a string then there was an issue getting them from the node.
+		// In this case we try and create the calls with the registry on `block`
+		// The block from `api.derive.chain.getBlock` has the most recent registry,
+		// which coudl cause issues with historical querries.
+		// On the other hand, we know `events` will have the correctly dated query
+		// since it is a storage query..
```
emostov, line 333:
```suggestion
		// since it is a storage query.
```

src/services/blocks/BlocksService.ts
```diff
@@ -339,15 +321,23 @@ export class BlocksService extends AbstractService {
 	 */
 	private extractExtrinsics(
 		block: Block,
-		events: EventRecord[] | string,
+		events: Vec<EventRecord> | string,
 		extrinsicDocs: boolean
 	) {
 		const defaultSuccess = typeof events === 'string' ? events : false;
+		// Note, if events is a string then there was an issue getting them from the node.
+		// In this case we try and create the calls with the registry on `block`
```
emostov, line 329:
```suggestion
		// In this case we try and create the calls with the registry on `block`.
```

src/services/blocks/BlocksService.ts
```diff
@@ -339,15 +321,23 @@ export class BlocksService extends AbstractService {
...
+		// The block from `api.derive.chain.getBlock` has the most recent registry,
+		// which coudl cause issues with historical querries.
```
emostov, line 331:
```suggestion
		// which could cause issues with historical querries.
```

TarikGul (approved):
LGTM, benchmarked and tested manually

Issue #495, opened 2021-03-29
Script to test queries across multiple runtimes, and different blocks


TarikGul:
**Proposed Change or Idea**

I wrote an example as to what it would look like [here](https://github.com/TarikGul/sidecar-runtime-test). Every time we push out a new release we have to test sidecar manually against chains such as 'Polkadot', 'Kusama', 'westend' and make sure there are no errors across many types of transactions and runtimes. We might find something like this useful, whether we want to integrate it in Sidecar or use it as an external helper script. 

Things that could be updated with it. A arg parser to determine which chain we want to test against, and a bash script to automate it all. In the example I have it working with Polkadot.

dvdplm:
Maybe something we can have CI do for us?

TarikGul:
Yea I will open up an issue with the CI team, and see what kind of thoughts they have on this.

dvdplm:
Maybe we could leverage the WASM runtime override feature that @insipx added for `substrate-archive`, and maybe even the library of wasm blobs maintained there?

TarikGul:
Now that the [sidecar-runtime-test](https://github.com/TarikGul/sidecar-runtime-test) is working well, and is used frequently per release, I will clean it up and work on making it a `paritytech` repo. For now I think it's better to have it as a helper library instead of integrated into sidecar because it requires Sidecar to be in either dev mode, or built and then ran.

PR #549, opened 2021-05-14
test: integrate runtime-tests as a helper library

TarikGul:
closes: [#495](https://github.com/paritytech/substrate-api-sidecar/issues/495)

This integrates [sidecar-runtime-test](https://github.com/TarikGul/sidecar-runtime-test) into Substrate-api-sidecar. I also used [this script](https://github.com/paritytech/substrate-api-sidecar/blob/na-script-for-runtime-tests/scripts/run_chain_tests.py) written by @niklasad1 to help automate the tests for each chain.

TODO:
- [X] Fix and organize types inside of runtime-tests
- [X] Setup package.json 
    - [X] Lint, test, and build should ignore runtime-tests for sidecar
    - [X] Lint, test, and build should ignore sidecar for runtime-tests
- [X] Adjust readme for publishing and releasing
- [X] Setup a readme for  runtime-tests (Should explain Python env, and dependencies)


scripts/run_chain_tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+def run_process(args):
+    return subprocess.check_output(args, stderr=subprocess.STDOUT,
+                                   encoding="utf-8")
+
+def run_chain_test(chain):
+    run_process(["yarn"])
+
+    if chain == "polkadot":
+        url = "wss://rpc.polkadot.io"
```
emostov, line 18: Would be cool if we had our own nodes - but I supposed you could not check the addresses into source control.

scripts/run_chain_tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+    print(res)
+    return res.find("PASS")
+
+
+def main():
+    polka_test = run_chain_test("polkadot")
+    kusama_test = run_chain_test("kusama")
+    westend_test = run_chain_test("westend")
+    # shutil.rmtree(sidecar_test_path)
```
niklasad1, line 55:
```suggestion
```

scripts/run_chain_tests.py
niklasad1: I think we could input the URLs as either `CLI args` or set `env variables` for them instead.

scripts/run_chain_tests.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import os
+import psutil
+import shutil
```
niklasad1, line 5: unused now?!

```suggestion
```

niklasad1 (approved):
Great,

In another PR: would be super to have a job that runs this on `chore(release)` PRs in the CI.

runtime-tests/index.ts
```diff
@@ -0,0 +1,39 @@
...
+	chain: 'polkadot',
+};
+const argv = process.argv.slice(0, 2);
+
+/**
+ * We return empty string's as an indication of a false value. JS recognizes "" as falsey.
+ * NodeJS.Process.argv.reduce expects the return value of the accumulator to
+ * be of type string, this is a way of satisfying the compiler, but also
+ * returning the correct behavior.
+ */
```
dvdplm, line 11: Not sure I understand what this does. What is the expected input here? It walks all cli arguments, filters out anything that starts with "/", sets the `config[cmd]` as it goes? How is it used?

runtime-tests/index.ts
```diff
@@ -0,0 +1,39 @@
...
+	if (arg.startsWith('--')) {
+		const sub = arg.substring('--'.length);
+		// Will only accept the flag if it's within the config
+		if (Object.keys(config).includes(sub)) {
+			// Set the sub to the next cmd
+			return sub;
+		}
+	}
+
+	argv.push(arg);
```
dvdplm, line 28: I'm used to seeing `reduce()` return the accumulator at every iteration (including the last) and assign the return value to a variable, e.g. `let sum = [1,2,3].reduce((acc, x) => x*x)`. Instead here it seems like you're not using the return value of the whole `reduce` and relying instead on side-effects like this one. Should it be a for-loop then?

runtime-tests/index.ts
TarikGul: Yea this can be a bit confusing. 

1. Parses out all inputs that start with `/`. (This is so we dont save things such as the path of the node.js binary into the config)

2. For everything else If the input doesn't start with `--`, then we return `""` (This sets the accumulator in this case `cmd` to false). This will resolve to a falsey value which will stop the reducer from saving it into the config. 

3. `argv.push(arg)` is used to save any local config options in order to pass it into the jest binary such as `--config=./runtime-tests/jest.config.js`. 

4. If the `arg` is equal to `--` and has the name of `chain` then it will set it equal to the accumulator or `cmd` and then the arg the follows will be the argument for the command itself.

runtime-tests/index.ts
TarikGul: So in this case I am using the reducer a little differently than normal, the accumulator in this case is being used to identify falsey values, vs truthy commands. I stayed away from the `for loop` because I didnt want to introduce messy iterator manipulation in order to parse the args, and I think the reducer itself allows for more customization for the future in case we wanted to add arguments to the script.

scripts/run_chain_tests.py
TarikGul: Yea I agree having our own nodes would be very nice. Would be a nice addition for the tools team in general.

scripts/run_chain_tests.py
```diff
@@ -0,0 +1,62 @@
...
+        url = "wss://westend-rpc.polkadot.io"
+        chain = "westend"
+    else:
+        return -1
+
+    os.environ["SAS_SUBSTRATE_WS_URL"] = url
+    proc = subprocess.Popen(["yarn", "dev"], stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT)
+
+    print('Sidecar is loading in developement mode...')
```
niklasad1, line 32:
```suggestion
    print('Sidecar is loading in development mode...')
```

TarikGul:
@emostov Before we merge this, can you pull this branch down and test two things as a sanity check as well. 

1. `yarn test:init-runtime-tests` - Lets make sure you like the output as well, and the stdout output. 

2. `yarn lint` - Sanity check to make sure the linting issue was something local to my machine and not the setup.

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
+## Summary
+
+This is a helper library for Sidecar to run runtime tests against specific chains, at certain blocks.
```
emostov, line 3: What does "runtime tests" mean here?

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
+## Summary
+
+This is a helper library for Sidecar to run runtime tests against specific chains, at certain blocks. 
+
+## Testing
+
+Below I will go over testing for Polkadot. These instructions are specific to runtime-tests against one chain.
```
emostov, line 7:
```suggestion
The below instructions are specific to runtime-tests against one chain. 
```

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
...
+the root directory of sidecar.
+
+### Polkadot 
+
+Lets first get sidecar ready in a seperate terminal.
+
+```
+$ cd substrate-api-sidecar
+$ git checkout <your_branch>
+$ export SAS_SUBSTRATE_WS_URL=wss://rpc.polkadot.io
```
emostov, line 18:
```suggestion
$ export SAS_SUBSTRATE_WS_URL=<network archive node>
```

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
...
+$ yarn
+$ yarn dev
```
emostov, line 20:
```suggestion
$ yarn build && yarn start
```

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+Sidecar should be now connected to the node, and running successfully. If you are having issues running sidecar and are receiving some odd behavior, contact the maintainers, or file an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
```
emostov, line 23:
```suggestion
Sidecar should now connected to the node and running successfully. If you are having issues running sidecar and are receiving some odd behavior, contact the maintainers, or file an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
```

emostov, line 23:
```suggestion
Sidecar should be now connected to the node, and running successfully. If you find a bug file an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
```

runtime-tests/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Now lets run our runtime tests against polkadot. Go to the seperate terminal and run:
```
emostov, line 25:
```suggestion
Now lets run our runtime tests against polkadot. Go to a separate terminal and run:
```
?

runtime-tests/endpoints/polkadotEndpoints.ts
```diff
@@ -0,0 +1,12 @@
+export const polkadotEndpoints = [
+	['/blocks/943438', '943438'], //v17
```
emostov, line 2: Is it redundant to have the same number as next to the URL?

runtime-tests/index.ts
dvdplm: Hmm, I wonder if we're trying to be a bit too clever for our own good here; feels like this should be easier to do? ("no it's just you being thick" is a valid answer here!).

I was actually wondering something more basic: what is this args processing for? Is this right (or close to right)? 

"Process the command line arguments used to launch the test session. Merge any config options provided at the command line with the on disk config."

runtime-tests/endpoints/polkadotEndpoints.ts
TarikGul: So this is actually to stop us from having to parse the endpoint for the block number. I use the block number to confirm the request was successful. Each of the values in each array are passed as arguments into the callback that `test.each` returns.

runtime-tests/index.ts
TarikGul: Yea so you actually make a great point and I was able to incredibly simplify it. This method is way too clever for what it is suppose to be. I basically just wrote out a custom parser which doesn't need to be done either. 

I chose to do it this way because there is no straight forward way to pass custom CLI commands to jest other than setting an entry file, and calling the binary after adding a custom configuration to the disk. But that being said I also could have just used a arg-parser and then directly pass it in same as I did. Which is exactly what I changed it too. I added `argparse` now and have it just be the same as any other node argument parsing tool. No loops, or complicated logic anymore.

Good catch. Thanks

runtime-tests/helpers/consts.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Port to Substrate-api-sidecar
+ */
+export const PORT = 8080;
+
+/**
+ * Path to Substrate-api-sidecar
+ */
+export const URL_PATH = '127.0.0.1';
```
emostov, line 9:
```suggestion
export const HOST = '127.0.0.1';
```
maybe?

runtime-tests/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * We return empty string's as an indication of a false value. JS recognizes "" as falsey.
+ * NodeJS.Process.argv.reduce expects the return value of the accumulator to
+ * be of type string, this is a way of satisfying the compiler, but also
+ * returning the correct behavior.
+ */
+process.argv.reduce((cmd, arg) => {
+	if (cmd) {
+		if (cmd.startsWith('/')) return '';
```
emostov, line 14:
```suggestion
		if (cmd.startsWith('/')) {
		    return ''
		};
```

runtime-tests/runtimeTests.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+	/**
+	 * Test runtimes for `/blocks`
+	 */
+	test.each(polkadotEndpoints)(
+		'Given path %p, it should return block height %p',
+		async (blockPath, blockHeight) => {
+			const res = await request(blockPath, URL_PATH, PORT);
+			const responseJson = JSON.parse(res) as BlockResponse;
+
+			expect(responseJson['number']).toBe(blockHeight);
```
emostov, line 23: In the future we should just take a snapshot of the response, save them and then compare the responses those save snapshots. That way we would essentially have an e2e user api contract testing

TarikGul: I very much like this idea, do you think I should just add it in this PR?

runtime-tests/index.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+const argv = process.argv.slice(0, 2);
+
+/**
+ * We return empty string's as an indication of a false value. JS recognizes "" as falsey.
+ * NodeJS.Process.argv.reduce expects the return value of the accumulator to
+ * be of type string, this is a way of satisfying the compiler, but also
+ * returning the correct behavior.
+ */
+process.argv.reduce((cmd, arg) => {
```
emostov, line 12:
```suggestion
process.argv.slice(2).reduce((cmd, arg) => {
```
Shouldn't we not be checking the first two elements of argv since they are the program?

runtime-tests/runtimeTests.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import { request } from './helpers/request';
+import { BlockResponse, ChainConfig, ChainSpec } from './types';
+
+const config = JSON.parse(process.env.__CONFIGURATION as string) as ChainConfig;
+const chain = config.chain as ChainSpec;
+
+const polkadotEndpoints: string[][] = endpoints[chain];
+
+describe('Runtime Tests for blocks', () => {
+	jest.setTimeout(15000);
```
emostov, line 12: What does this timeout do for us?

runtime-tests/index.ts
TarikGul: Yea that totally makes sense, I just completely changed this whole file, and made it way simpler with an arg parser.

runtime-tests/runtimeTests.spec.ts
TarikGul: The default timeout in jest is 5000 ms, which sometimes with all the requests can take up to or more than 10000 or more ms. This sets the default timeout to 15000 so the test doesn't fail for a slow response.

runtime-tests/index.ts
```diff
@@ -0,0 +1,35 @@
...
+config['chain'] = args.chain
+
+// Pass in the Jest CLI path option to the correct config
+argv.push('--config=' + args.config)
+
+// Store configuration on env
+process.env.__CONFIGURATION = JSON.stringify(config);
+
+// Setting real ARGV
+process.argv = argv;
```
emostov, line 32: Is this a normal pattern? Makes sense but I haven't seen it before

runtime-tests/runtimeTests.spec.ts
emostov: Either sounds good to me! Will be a super valuable feature to be able to show how responses change with their diffs

runtime-tests/index.ts
```diff
@@ -0,0 +1,35 @@
+import { ArgumentParser } from 'argparse';
+
+const config = {
+	chain: 'polkadot',
```
emostov, line 4: Why do we start with polkadot?

emostov (approved):
LGTM

For follow up based on convos 

- Scan sidecar process stdout on startup to tell if sidecar is connected to the archive node before starting tests (right now we have a 30sec timeout)
- Save response from each endpoint and compare those to the responses received while the test is running. We can use these to document changes to the user api contract

runtime-tests/index.ts
TarikGul: 'polkadot' currently just sits as a placeholder for the key value pair.

runtime-tests/index.ts
TarikGul: I believe for passing in custom cli args into a jest instance its been used by many, but I am not sure it's a common pattern.

runtime-tests/runtimeTests.spec.ts
TarikGul: 2 part PR it is! :)

emostov:
> @emostov Before we merge this, can you pull this branch down and test two things as a sanity check as well.
> 
> 1. `yarn test:init-runtime-tests` - Lets make sure you like messaging as well as the stdout output.
> 2. `yarn lint` - Sanity check to make sure the linting issue was something local to my machine and not the setup.

@TarikGul Ran it locally and it all checks out ✅

TarikGul:
Ready to merge.

Per my conversation with @emostov, we think it could be a great idea to turn this runtime-tests into a full e2e test for the `/blocks` endpoint. This would be done in a Part 2 extension of this PR.

### Part 2

Currently the runtime tests only checks to make sure we receive a successful response back from the node. I do this by checking the block height of the response against the expected block height. But this could be done better by checking against a full response. Therefore simulating a full user request-response cycle. This would mean efficiently storing the endpoints we currently have logged for each runtime per chain, and then matching the JSON responses from the node against our stored responses.

scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+## Script `run_chain_tests.py`
+
+### Summary
+
+This script calls the local runtime-tests helper library in order to test the current branch or development enviornment against 
+a collection of different blocks, across different runtimes. It does this for Polkadot, Kusama, and Westend. 
+
+### Requirements
```
niklasad1, line 17: It would be better to do:

To save the file `requirements.txt` by `$ pip freeze > requirements.txt`
Then just state `$ pip install -r requirements.txt` in the README instead

runtime-tests/index.ts
```diff
@@ -1,36 +1,37 @@
...
-		// Will only accept the flag if it's within the config
-		if (Object.keys(config).includes(sub)) {
-			// Set the sub to the next cmd
-			return sub;
-		}
-	}
-
-	argv.push(arg);
-	return '';
+const parser = new ArgumentParser();
```
dvdplm, line 17: ❤️

dvdplm (approved):

scripts/README.md
TarikGul: Awesome! I was hesitant at first to put a `requirements.txt` for one dependency, but this convinces me, Thanks!

TarikGul: Awesome fixed.

Issue #544, opened 2021-05-11
Reduce controller config boilerplate


emostov:
Ideally we should just state which controllers we want and assume that all controllers not explicitly mentioned are not available. So it'd be `controllers = [Blocks, BlocksExtrinsics, AccountAssets, …, …];`. The controllers for kilt, dock, kulupu etc are getting pretty boilerplatey.

_Originally posted by @dvdplm in https://github.com/paritytech/substrate-api-sidecar/pull/533#discussion_r629967716_

TarikGul:
Tackling this today.

PR #555, opened 2021-05-17
fix: reduce controller config boilerplate

TarikGul:
closes: [#544](https://github.com/paritytech/substrate-api-sidecar/issues/544)

Per the issue, I have reduced the boilerplate for controllers to the following: `controllers: ['Blocks', 'BlocksExtrinsics', ....]`

Moving forward chain-configs should only contain the controllers they wish to use, and any other controller that is absent wont be mounted for the instance.


emostov (approved):
Code looks good but need to update docs here: https://github.com/paritytech/substrate-api-sidecar/blob/master/CHAIN_INTEGRATION.md#1-create-a-controller-config

dvdplm (approved):
ty!

PR #561, opened 2021-05-25
test: e2e-tests

TarikGul:
## Part 2

This is an extension of [PR #549](https://github.com/paritytech/substrate-api-sidecar/pull/549).

Turns the current runtime tests into e2e tests.

With `test.jest` all values in the 2D array have to be strings. This means we have to `JSON.stringify` the expected responses from our JSON files when sending it over to Jest, then `JSON.parse` that string value and do a `.toStrictEqual` to make sure the e2e response is correct. 

With the given description above I decided to keep each block response files as `JSON` instead of `.ts` for 3 reasons. 

1. Having each blockResponse as a string version of the JSON in a `.ts` file can cause undesired behavior with the typescript compiler when importing them all into their appropriate `index` files. (Specifically with very large strings)

2. Readability of the JSON files are way better.

- [X] Confirm e2e tests work
- [X] Import all JSON files for each block
- [X] Test each chain separately and change the tests to strict equal.
- [X] Change Filename `runtime-tests` to `e2e-tests`


TarikGul:
When reviewing be aware that most of the additions are JSON files that represent the responses from the blocks. There is no need to read through the actual JSON file as they are actual responses from the respective chains.

e2e-tests/endpoints/westend/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import block3889918 from './3889918.json';
+import block4781573 from './4781573.json';
+import block5277929 from './5277929.json';
+import block5480769 from './5480769.json';
+import block5493461 from './5493461.json';
+import block5495855 from './5495855.json';
+import block5657482 from './5657482.json';
+
+export const westendEndpoints = [
+	['/blocks/3032259', JSON.stringify(block3032259)],
```
insipx, line 12:
```suggestion
	['/blocks/3032259', JSON.stringify(block3032259)], //v45
```

e2e-tests/endpoints/westend/index.ts
```diff
@@ -0,0 +1,21 @@
...
+	['/blocks/3813629', JSON.stringify(block3813629)],
+	['/blocks/3889918', JSON.stringify(block3889918)],
+	['/blocks/4781573', JSON.stringify(block4781573)],
```
insipx, line 15:
```suggestion
	['/blocks/4781573', JSON.stringify(block4781573)], //v49
```

e2e-tests/endpoints/westend/index.ts
```diff
@@ -0,0 +1,21 @@
...
+	['/blocks/5277929', JSON.stringify(block5277929)],
```
insipx, line 16:
```suggestion
	['/blocks/5277929', JSON.stringify(block5277929)], //v50
```

e2e-tests/endpoints/westend/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import block5277929 from './5277929.json';
+import block5480769 from './5480769.json';
+import block5493461 from './5493461.json';
+import block5495855 from './5495855.json';
+import block5657482 from './5657482.json';
+
+export const westendEndpoints = [
+	['/blocks/3032259', JSON.stringify(block3032259)],
+	['/blocks/3813629', JSON.stringify(block3813629)],
+	['/blocks/3889918', JSON.stringify(block3889918)],
```
insipx, line 14:
```suggestion
	['/blocks/3889918', JSON.stringify(block3889918)], //v45
```

e2e-tests/endpoints/westend/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import block4781573 from './4781573.json';
+import block5277929 from './5277929.json';
+import block5480769 from './5480769.json';
+import block5493461 from './5493461.json';
+import block5495855 from './5495855.json';
+import block5657482 from './5657482.json';
+
+export const westendEndpoints = [
+	['/blocks/3032259', JSON.stringify(block3032259)],
+	['/blocks/3813629', JSON.stringify(block3813629)],
```
insipx, line 13:
```suggestion
	['/blocks/3813629', JSON.stringify(block3813629)], //v45
```

insipx (approved):
LGTM! just small nits because I saw westend was the only file missing runtime versions for its blocks.

emostov (approved):
Looks good 

Perhaps over engineering, but one step further would be to have a script to auto update those JSON responses when they broke. But I think we can wait until we need it to make that

TarikGul:
> script to auto update those JSON responses when they broke. But I think we can wait until we need it to make that

It could be very useful for us to have our own `--fix` flag which would allow for that. That way you can choose if you want to run that each time or not. 

The one thing that I would find tricky about this is staying away from touching the file system in a test, meaning the rewrite of the JSON files would have to be a post jest test script which would re fetch each JSON response and rewrite them accordingly.

emostov:
> meaning the rewrite of the JSON files would have to be a post jest test script which would re fetch each JSON response and rewrite them accordingly.

We could do an entirely separate script.

TarikGul:
Do not merge yet, as I want to do one more sanity check.

TarikGul:
Good to merge

PR #570, opened 2021-05-31
fix!: fundInfo with updated polkadot/api

TarikGul:
Before publishing the next release this PR updates the fields `firstSlot` and `lastSlot` within type `FundInfo` to correctly match the type generated from `@polkadot/api`.

I updated the docs, tests, and service, as well as updated the dependencies.

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -2386,10 +2386,10 @@ components:
           enum:
           - preEnding
           - ending
-        firstSlot:
+        firstPeriod:
           type: string
           format: unsignedInteger
-        lastSlot:
+        lastPeriod:
           type: string
           format: unsignedInteger
         trieIndex:
--- a/package.json
+++ b/package.json
@@ -43,9 +43,9 @@
     "test:init-runtime-tests": "python3 ./scripts/run_chain_tests.py"
   },
   "dependencies": {
-    "@polkadot/api": "^4.11.2",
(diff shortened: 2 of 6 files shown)


insipx (approved):

docs/src/openapi-v1.yaml
```diff
@@ -2386,10 +2386,10 @@ components:
           enum:
           - preEnding
           - ending
-        firstSlot:
+        firstPeriod:
```
niklasad1, line 2389: I don't understand this, the variable is just renamed and the type is not changed?

TarikGul: Yea so the fields `firstSlot`, and `lastSlot` within the `FundInfo` type has been renamed to `firstPeriod` and `lastPeriod`. I worded the description wrong it should say: 

"updates the fields 'firstSlot' and 'lastSlot' within the type `FundInfo`"

niklasad1 (approved):

docs/src/openapi-v1.yaml
dvdplm: Given `FundInfo` is a public type, changing it's member names like this is a breaking change right? So the next release is major?

dvdplm (approved):

docs/src/openapi-v1.yaml
TarikGul: Yes indeed. This would be a breaking change, fixed the commit name.

Issue #569, opened 2021-05-31
GET on /accounts/accountId/asset-balances with connected Westend node returns an error 500


qinheqing:
env：
Polkadot v0.9.3
substrate-api-sidecar 6.2.3
Westend

`http://127.0.0.1:8080/accounts/5CMFCMU2wEbs6hCX7B2chU9yQMYxuiJiJxbQg6Nx97VAaYtX/asset-balances`

`{
    "code": 500,
    "message": "Cannot read property 'asset' of undefined",
    "stack": "TypeError: Cannot read property 'asset' of undefined\n    at AccountsAssetsService.fetchAssetBalances (/usr/lib/node_modules/@substrate/api-sidecar/build/src/services/accounts/AccountsAssetsService.js:24:49)\n    at processTicksAndRejections (internal/process/task_queues.js:93:5)\n    at async AccountsAssetsController.getAssetBalances (/usr/lib/node_modules/@substrate/api-sidecar/build/src/controllers/accounts/AccountsAssetsController.js:61:57)\n    at async /usr/lib/node_modules/@substrate/api-sidecar/build/src/controllers/AbstractController.js:159:9",
    "level": "error"
}`

emostov:
Westend doesn't have the assets pallet afaik. @TarikGul we should take the endpoint out of the controller config?

TarikGul:
Yea lets take it off of westend, polkadot, and kusama, and the default controller. While we are at it should we add a statetmint config as well?

emostov:
Yup, makes sense to me

PR #573, opened 2021-06-01
fix: remove assets endpoint from chains config, add statemint, and statemine

TarikGul:
closes: [#569](https://github.com/paritytech/substrate-api-sidecar/issues/569)

This PR removes the assets endpoint from `Polkadot`, `Kusama`, and `Westend` while adding new chain configs for both `statemint` and `statemine`.


insipx (approved):

emostov (approved):

src/chains-config/statemineControllers.ts
```diff
@@ -0,0 +1,28 @@
+import { ControllerConfig } from '../types/chains-config';
+
+/**
+ * Statemine configuration for Sidecar.
+ */
+export const statemineControllers: ControllerConfig = {
+	controllers: [
+		'Blocks',
+		'BlocksExtrinsics',
+		'AccountsAssets',
```
dvdplm, line 10: This should be the first in the list I think? My alphabet-OCD gets all tingly.

Issue #574, opened 2021-06-01
[ERROR]: Unable decode storage assets.asset:: createType(AssetDetails)


TarikGul:
**Description**

local Path: `http://127.0.0.1:8080/pallets/assets/8/asset-info`
rpc-endpoint: `wss://westend-shell-rpc.parity.io`
Sidecar-version: 7.0.0

This happens when querying the most recently launched `westmine`.

### Log
```
[...]

PR #575, opened 2021-06-02
fix: update deps, add westmint, and westmine

TarikGul:
I updated the polkadot deps to `beta` as it fixes a `set_storage` decoding error for assets. 

Added `westmint` and `westmine`

Changed, all controllers to alphabetical order.

closes: [#574](https://github.com/paritytech/substrate-api-sidecar/issues/574)


package.json
```diff
@@ -43,7 +43,7 @@
     "test:init-e2e-tests": "python3 ./scripts/run_chain_tests.py"
   },
   "dependencies": {
-    "@polkadot/api": "^4.12.1",
+    "@polkadot/api": "beta",
```
insipx, line 46: dq: any implications to using the beta channel / this mean we wouldn't have to update @polkadot/api ?

insipx (approved):

package.json
niklasad1: I have the same question otherwise it looks good.

TarikGul: > any implications

Probably best if we didn't rely on the beta for each weeks release. As of now I think we should just update `@polkadot/api` each monday to its main release, and if we have to do a beta update mid week to master to fix a bug then we do that. This way we could find potential issues per each main release. 

So i would say one major implication of beta would be making it harder to track changes and apply them per release.

niklasad1: Ok, are you saying that this enabled this for testing and change to a specific release on Monday?

TarikGul: > Ok, are you saying that this enabled this for testing/for it work and change to a specific release on Monday (when it's released)?

Yea exactly. This beta-version is just fixing a bug in polkadot-js for our master branch in regards to assets, but once monday comes we would switch back to the latest(official) release.

src/chains-config/index.ts
```diff
@@ -29,6 +31,8 @@ const specToControllerMap = {
 	'dock-main-runtime': dockMainnetControllers,
 	statemint: statemintControllers,
 	statemine: statemineControllers,
+	westmine: westmineControllers,
```
emostov, line 34: Might DRY things up a bit if we had all these keys point to the same values - just means if those parachains ever diverged we would need to re-evaluate controller configs; but we will need to do that anyways.

emostov (approved):
Looks good. See my one comment for a little idea to DRY up controller configs in the future (not necessary in this PR tho)

src/chains-config/index.ts
TarikGul: Ahhh, I see what you are saying. Good point. Yea Ill fix that up and make them one value. Also will definitely lower the amount of files we have which are preferred.

TarikGul:
Before merging. Two things need to get done.

1. Reorder controllers so they are in alphabetical order.

2. Implement Zekes suggestion about DRY'ing the values for the parachain controllers.

TarikGul:
Updated the `@polkadot/api` dep off of beta to 4.13.1. 

@dvdplm Fixed all the controllers to alphabetical order.

PR #584, opened 2021-06-14
fix: update deps, reconfigure tests to reflect most recent polkadot-js changes

TarikGul:
This PR updates the deps for polkadot-js, and substrate/dev

I also reconfigured tests in order to reflect the most recent internal changes to polkadot-js [here](https://github.com/polkadot-js/api/commit/400e14d6c94613e3b013f207607e4c16c61d9b4b#diff-4057209e51321417a7ba8384eeaa7221603262b495ae1f975753c129de848f4eR24)

### DIFF
--- a/src/sanitize/mockData.ts
+++ b/src/sanitize/mockData.ts
@@ -30,7 +30,7 @@ export const PRE_SANITIZED_BALANCE_LOCK = kusamaRegistry.createType(
 	'Vec<BalanceLock>',
 	[
 		{
-			id: 'LockId',
+			id: '00000000',
 			amount: kusamaRegistry.createType(
 				'Balance',
 				'0x0000000000000000ff49f24a6a9c00'
--- a/src/test-helpers/constants.ts
+++ b/src/test-helpers/constants.ts
@@ -23,8 +23,8 @@ export const MAX_U8 = '255';
 export const MAX_I8 = '127';
 export const MIN_I8 = '-128';
 
-export const H512_U64 =
+export const MAX_H512 =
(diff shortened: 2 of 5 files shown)


src/test-helpers/constants.ts
```diff
@@ -22,3 +22,9 @@ export const MAX_U8 = '255';
 
 export const MAX_I8 = '127';
 export const MIN_I8 = '-128';
+
+export const H512_U64 =
```
niklasad1, line 26:
```suggestion
export const MAX_H512 =
```

TarikGul: I like the name suggestion!
Just changed them all to reflect this.

niklasad1 (approved):

src/sanitize/mockData.ts
```diff
@@ -30,7 +30,7 @@ export const PRE_SANITIZED_BALANCE_LOCK = kusamaRegistry.createType(
 	'Vec<BalanceLock>',
 	[
 		{
-			id: 'LockId',
+			id: '00000000',
```
dvdplm, line 33: What happened here?

dvdplm (approved):
lgtm.

I think in general that we should try to be explicit about any numbers that appear in tests. If they are just random numbers or have a meaning, it'd be helpful to see a short comment as a hint of "don't bother wondering where this came from".

src/sanitize/mockData.ts
TarikGul: It expected the id to be of 8 bytes but instead `lockId` is 6 bytes

PR #587, opened 2021-06-22
fix: createCalcFee bug

TarikGul:
[BUG] Urgent

Using Kusama as an example querying any block with a runtime less than 1062 (the minimum runtime version used to calc fees for kusama) will throw an error because `api.query.transactionPayment?.nextFeeMultiplier?.at` `Multiplier` type isnt supported.  

Currently there is no check before `api.query.transactionPayment?.nextFeeMultiplier?.at` to make sure we don't calculate any fees for runtime's less than the least known runtime for fee's.

This PR fixes that by adding a check.

### DIFF
--- a/e2e-tests/endpoints/kusama/index.ts
+++ b/e2e-tests/endpoints/kusama/index.ts
@@ -1,3 +1,4 @@
+import block9253 from './9253.json';
 import block2350438 from './2350438.json';
 import block2684767 from './2684767.json';
 import block2713513 from './2713513.json';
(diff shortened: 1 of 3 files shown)


insipx (approved):
LGTM. DQ: Is there a better message than just 'Error', not sure if that bubbles up into something else, unfamiliar with the codebase

TarikGul:
Yea just noticed that, putting a better message in there. Good question for sure!

niklasad1 (approved):

dvdplm (approved):

e2e-tests/endpoints/kusama/index.ts
```diff
@@ -19,6 +20,7 @@ import block7354817 from './7354817.json';
 import block7519631 from './7519631.json';
 
 export const kusamaEndpoints = [
+	['/blocks/9253', JSON.stringify(block9253)], //v1020
```
dvdplm, line 23: 👍

Issue #588, opened 2021-06-23
Update auction endpoints (track paritytech/polkadot/pull/3346)


emostov:
https://github.com/paritytech/polkadot/pull/3346 will have some updates to the how the auctions are exposed. Some logic in the auction endpoints will likely need to get tweaked and will need changes to reflect the new `AuctionStatus` enum

TarikGul:
Awesome just saw that it got merged into master, and is going in the next runtime upgrade. Thanks!

TarikGul:
[591](https://github.com/paritytech/substrate-api-sidecar/pull/591) closes this.

TarikGul:
Prematurely closed this issue, and confused it with another. Reopening.

PR #593, opened 2021-06-29
fix: add `vrfDelay` as a `phase` return value for ongoing auctions to reflect VRFdelays in polkadot

TarikGul:
closes: [#588](https://github.com/paritytech/substrate-api-sidecar/issues/588)

### Summary

This PR aims to reflect the addition of the `AuctionStatus` enum in polkadot for parachains, specifically the `VRFDelay`. When an auction is past its `finishEnd` block the auction is then in a `vrfDelay` phase. 

### Changes

* I updated the tests to reflect this change, I also spotted a minor issue in the tests and reversed the output results to reflect a real world scenario with the auctions.

* Added logic inside of the `auctionsCurrent` method for `ParasService` that will add a `vrfDelay` phase as a potential result.

* I added the `vrfDelay` result to the `AuctionPhase` type, and changed `starting`=> `startPeriod`, `ending` => `endPeriod`. 

Overall there are 3 potential return values for a phase:

```
phase :
  startPeriod
  endPeriod
  vrfDelay
```

* Updated the docs

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -2338,12 +2338,15 @@ components:
         phase:
           type: string
           enum:
-          - opening
+          - starting
           - ending
+          - delay
           description: |
-            Whether the auction is in the `opening` or `ending` phase. The
-            `ending` phase is where the eventual winners are retroactively
-            picked from. `null` if there is no ongoing auction.
+            An auction can be in one of 4 phases. Both `starting` and `ending` denoting
+            an ongoing auction. `delay` is part of the VRFDelay inside of Substrate which is a Period 
+            `finishEnd` block number where contributions are no longer accepted. After this `delay` phase 
+            the auction is officially finalized and a winner is declared. If `null` is returned 
+            this means the auction has not started yet.
         auctionIndex:
           type: string
           format: unsignedInteger
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -2338,8 +2338,8 @@ components:
         phase:
           type: string
           enum:
-          - starting
-          - ending
+          - startPeriod
+          - endPeriod
           - vrfDelay
           description: |
             An auction can be in one of 4 phases. Both `startingPeriod` () and `endingPeriod` indicate
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -2340,13 +2340,13 @@ components:
           enum:
           - starting
           - ending
-          - delay
+          - vrfDelay
           description: |
-            An auction can be in one of 4 phases. Both `starting` and `ending` denoting
-            an ongoing auction. `delay` is part of the VRFDelay inside of Substrate which is a Period 
-            `finishEnd` block number where contributions are no longer accepted. After this `delay` phase 
-            the auction is officially finalized and a winner is declared. If `null` is returned 
-            this means the auction has not started yet.
+            An auction can be in one of 4 phases. Both `startingPeriod` () and `endingPeriod` indicate
+            an ongoing auction, while `vrfDelay` lines up with the `AuctionStatus::VrfDelay` . Finally, a value of `null`
+            indicates there is no ongoing auction. Keep in mind the that the `finishEnd` field is the block number the 
+            `endingPeriod` finishes and the `vrfDelay` period begins. The `vrfDelay` period is typically about an 
+            epoch long and no crowdloan contributions are accepted. 
         auctionIndex:
           type: string
           format: unsignedInteger
--- a/src/services/paras/ParasService.spec.ts
+++ b/src/services/paras/ParasService.spec.ts
@@ -188,7 +188,7 @@ describe('ParasService', () => {
 				at: expectedAt,
 				beginEnd: '1000',
 				finishEnd: '21000',
-				phase: 'delay',
+				phase: 'vrfDelay',
 				auctionIndex: '4',
 				leasePeriods: ['39', '40', '41', '42'],
 				winning: [
--- a/src/services/paras/ParasService.spec.ts
+++ b/src/services/paras/ParasService.spec.ts
@@ -1,11 +1,14 @@
 import BN from 'bn.js';
 
 import { sanitizeNumbers } from '../../sanitize/sanitizeNumbers';
+import { polkadotRegistry } from '../../test-helpers/registries';
 import {
 	auctionsInfoAt,
+	blockHash20000,
 	blockHash789629,
 	emptyVectorLeases,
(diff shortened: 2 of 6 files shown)


src/services/paras/ParasService.ts
```diff
@@ -238,7 +238,12 @@ export class ParasService extends AbstractService {
 			}
 
 			finishEnd = beginEnd.add(endingPeriod);
-			phase = beginEnd.gt(blockNumber) ? 'starting' : 'ending';
+
+			if (finishEnd.lt(blockNumber)) {
+				phase = 'delay';
```
emostov, line 243:
```suggestion
				phase = 'vrfDelay';
```

docs/src/openapi-v1.yaml
```diff
@@ -2338,12 +2338,15 @@ components:
...
+          - delay
           description: |
-            Whether the auction is in the `opening` or `ending` phase. The
-            `ending` phase is where the eventual winners are retroactively
-            picked from. `null` if there is no ongoing auction.
+            An auction can be in one of 4 phases. Both `starting` and `ending` denoting
+            an ongoing auction. `delay` is part of the VRFDelay inside of Substrate which is a Period 
+            `finishEnd` block number where contributions are no longer accepted. After this `delay` phase 
+            the auction is officially finalized and a winner is declared. If `null` is returned 
+            this means the auction has not started yet.
```
emostov, line 2349:
```suggestion
            An auction can be in one of 4 phases. Both `startingPeriod` () and `endingPeriod` indicate
            an ongoing auction, while `vrfDelay` lines up with the `AuctionStatus::VrfDelay` . Finally, a value of `null`
            indicates there is no ongoing auction. Keep in mind the that the `finishEnd` field is the block number the 
            `endingPeriod` finishes and the `vrfDelay` period begins. The `vrfDelay` period is typically about an 
            epoch long and no crowdloan contributions are accepted. 
```

docs/src/openapi-v1.yaml
```diff
@@ -2338,12 +2338,15 @@ components:
         phase:
           type: string
           enum:
-          - opening
+          - starting
           - ending
+          - delay
```
emostov, line 2343:
```suggestion
          - startingPeriod
          - endingPeriod
          - vrfDelay
```
I think we should maximize alignment with https://github.com/paritytech/polkadot/blob/60341e125563cbf655ab686ac5cef370ac33d16d/runtime/common/src/traits.rs#L150-L164

src/services/paras/ParasService.spec.ts
```diff
@@ -183,27 +186,27 @@ describe('ParasService', () => {
 
 			const expectedResponse = {
 				at: expectedAt,
-				beginEnd: '39',
-				finishEnd: '20039',
-				phase: 'ending',
+				beginEnd: '1000',
+				finishEnd: '21000',
+				phase: 'delay',
```
emostov, line 191:
```suggestion
				phase: 'vrfDelay',
```

src/services/paras/ParasService.spec.ts
```diff
@@ -214,6 +217,38 @@ describe('ParasService', () => {
...
+					'0x2d5d3fdb96b487d480b08b64ed69a65433c1713ae3579dd23704cb790aa3b2ae',
+				digest: {},
+			};
+			const header = polkadotRegistry.createType('Header', overrideHeader);
+
+			// Override the mockApi
+			(mockApi.rpc.chain.getHeader as unknown) = () =>
+				Promise.resolve().then(() => header);
+
+			const expectedResponse = 'ending';
```
emostov, line 241:
```suggestion
			const expectedResponse = 'endingPeriod';
```

src/services/paras/ParasService.ts
```diff
@@ -238,7 +238,12 @@ export class ParasService extends AbstractService {
...
+			} else {
+				phase = beginEnd.gt(blockNumber) ? 'starting' : 'ending';
```
emostov, line 245:
```suggestion
				phase = beginEnd.gt(blockNumber) ? 'startingPeriod' : 'endingPeriod';
```

src/types/responses/Paras.ts
```diff
@@ -12,7 +12,7 @@ import BN from 'bn.js';
 import { IOption } from '../util';
 import { IAt } from './';
 
-export type AuctionPhase = 'starting' | 'ending';
+export type AuctionPhase = 'starting' | 'ending' | 'delay';
```
emostov, line 15:
```suggestion
export type AuctionPhase = 'startingPeriod' | 'endingPeriod' | 'vrfDelay';
```

emostov (approved):
lgtm - just some suggestions on docs and the name of the enum variants .. although i am not sure if appending `Period` is worth it to align with better with https://github.com/paritytech/polkadot/blob/60341e125563cbf655ab686ac5cef370ac33d16d/runtime/common/src/traits.rs#L150-L164

docs/src/openapi-v1.yaml
dvdplm: > i will not that im not sure what to do for casing on `vrfDelay`

I think `vrfDelay` is fine. It could be `VRFDelay` as well, given it's an acronym? But `vrfDelay` sounds better to me.

dvdplm: I agree we should stay close to the Rust source. I also think it's a bit unfortunate we went with `-ing` rather than `StartPeriod`/`EndPeriod`.

dvdplm (approved):

docs/src/openapi-v1.yaml
TarikGul: Yea I like the idea of `vrfDelay`. One thing I just want to confirm too, if we append `Period` to `starting` and `ending` it will be a breaking change for the API. Just something to note. If you both think its appropriate while its still an experimental endpoint I am down to make the change.

TarikGul:
@emostov This is ready to merge, just want to make sure whether we should append `Period` and change this PR to a breaking change, or keep as if and just add the `vrfDelay`

emostov:
> @emostov This is ready to merge, just want to make sure whether we should append `Period` and change this PR to a breaking change, or keep as if and just add the `vrfDelay`

Its already a breaking change because when we used to show `ending` we will now sometimes show `vrfDelay`. IMO we should change `ending` => `endPeriod` & `starting` => `startPeriod` to better align with the `AuctionStatus` enum

TarikGul:
Ahh yes, the breaking change is there regardless because the new runtime logic is keeping `AuctionInfo` alive during the `vrfDelay` period. Good point. Updated `starting` => `startPeriod`, `ending` => `endPeriod` and reflected it in the docs. I will set up a release after I merge this in once the tests pass.

PR #596, opened 2021-07-01
feat: Basic support for H160 and H256 accounts.

crystalin:
This is a basic support got H160 accounts (like in Moonbeam/Moonriver) substrate chains.
I suspect we could pass an argument to `validateAddressMiddleware` to decide which type to support instead of relying on the `0x` prefix, but I'm not too familiar with the project.

What is your suggestion ?


TarikGul:
@crystalin Thank you for the PR! I will give it a thorough review and response come Monday.

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -32,17 +32,28 @@ export const validateAddressMiddleware: RequestHandler = (req, _res, next) => {
...
-		return [false, (error as Error).message];
-	}
+	if (address.startsWith('0x')) {
+		decoded = Uint8Array.from(Buffer.from(address.slice(2), 'hex'));
 
-	if (!defaults.allowedEncodedLengths.includes(decoded.length)) {
-		return [false, 'Invalid decoded address length'];
-	}
+		if (decoded.length != 20) {
+			return [false, `Invalid decoded address length ${decoded.length}`];
```
emostov, line 39:
```suggestion
			return [false, `Invalid decoded address length ${decoded.length} for H160 account`];
```

emostov:
We need to update tests and make sure to have both success and failure tests for standard substrate addresses and H160.

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -32,17 +32,28 @@ export const validateAddressMiddleware: RequestHandler = (req, _res, next) => {
 function checkAddress(address: string): [boolean, string | undefined] {
 	let decoded;
 
-	try {
-		decoded = base58Decode(address);
-	} catch (error) {
-		return [false, (error as Error).message];
-	}
+	if (address.startsWith('0x')) {
```
emostov, line 35: What if someone submits a normal substrate address in a raw hex form? I think we should double check that `base58Decode` would error on that. But I suspect it will validate the raw address form as well as the checksum, so if that is the case we should continue supporting those formats. If this is indeed the case I suggest we first try `base58Decode` and if that fails _then_ check for `0x` and the length of 20

crystalin: I updated the logic, can you check ?

emostov (approved):
thanks for the contribution - LGTM!

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -22,7 +22,7 @@ export const validateAddressMiddleware: RequestHandler = (req, _res, next) => {
 };
 
 /**
- * Verify that an address is a valid substrate ss58 address.
+ * Verify that an address is a valid substrate address.
  *
  * Note: this is very similar '@polkadot/util-crypto/address/checkAddress,
  * except it does not check the prefix.
```
dvdplm, line 28: Except now it does, for the `H160` case. :/

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -35,11 +35,21 @@ function checkAddress(address: string): [boolean, string | undefined] {
 	try {
 		decoded = base58Decode(address);
```
dvdplm, line 36: Could we use `base58Validate` here instead?

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -35,11 +35,21 @@ function checkAddress(address: string): [boolean, string | undefined] {
...
 	} catch (error) {
+		// If ss58 doesn't pass, we try H160 address.
+		if (address.startsWith('0x')) {
+			decoded = Uint8Array.from(Buffer.from(address.slice(2), 'hex'));
+
+			if (decoded.length != 20) {
+				return [false, `Invalid decoded address length ${decoded.length} for H160 account`];
+			}
```
dvdplm, line 44: There is `isEthereumAddress` that seems to do this job (and also runs more checks with `isEthereumChecksum`). Does it not work here?

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -35,11 +35,21 @@ function checkAddress(address: string): [boolean, string | undefined] {
 	try {
 		decoded = base58Decode(address);
 	} catch (error) {
+		// If ss58 doesn't pass, we try H160 address.
+		if (address.startsWith('0x')) {
+			decoded = Uint8Array.from(Buffer.from(address.slice(2), 'hex'));
```
dvdplm, line 40: "decoded" here is a bit unclear. On line 36 the address is decoded from base58 to hex. Here though, "decoding" means a different thing: cast a hex string to bytes. I would prefer clearer naming to properly distinguish the kind of decoding happening. Maybe `hexAddress`?

dvdplm (changes requested):
Seems like this could be done in a better way, but I could be wrong.

crystalin:
@dvdplm , I tried to improve the logic, please have a look

src/middleware/validate/validateAddressMiddleware.ts
```diff
@@ -22,27 +23,35 @@ export const validateAddressMiddleware: RequestHandler = (req, _res, next) => {
...
-	let decoded;
-
-	try {
-		decoded = base58Decode(address);
-	} catch (error) {
-		return [false, (error as Error).message];
+	let u8Address;
+
+	if (isHex(address)) {
+		u8Address = Uint8Array.from(Buffer.from(address.slice(2), 'hex'));
```
TarikGul, line 37: 👍

TarikGul:
LGTM, I will approve once the lint check is fixed. Thanks again for the PR.

NOTE: I tested this myself against a moonriver node, and everything worked as expected.

TarikGul (approved):
@crystalin This will make it into the next Sidecar release v8.1.0. @dvdplm Once you approve the changes made in response to your suggestions I will get this updated into the next Sidecar release for either tonight or tomorrow morning.

TarikGul:
> This will make it into the next Sidecar release v8.1.0.

This was premature insight of mine.

So we actually had a quick conversation about moving forward with this PR. We are actually going to leave it out of the next release to give this a second audit, we believe there might be a better way to do this, but just need to double check some alternatives. 

Moving forward with the 8.0.1 release and once this is merged we can always do another release.

crystalin:
@TarikGul it is fine for us now, we are hosting our custom version at the moment

dvdplm (approved):
lgtm, thank you @crystalin for humoring me. :)

emostov:
What is the status of this?

TarikGul:
@emostov Im happy to get this merged in.

TarikGul:
My main concern and delay was wondering if there was a better alternative to accepting a set of addresses within our middleware, but I think for now this is acceptable.

emostov:
@TarikGul - yeah I wouldn't say I have a full picture of all the different address schemes, but from what I understand this solution should additionally except ethereum addresses on top of what we already except. If for some reason we found out accepting both is not desirable I suppose we could add some config option. Not sure what other ways we could do this?

TarikGul:
@emostov Yea I agree with everything you said. I am going to do some last minute sanity checks, and if everything checks out again (which it should), I merge this tonight, and have it in tomorrow's release

TarikGul:
Merging.

PR #604, opened 2021-07-13
Make it a little easier to test individual chains

jsdw:
Just a wee tweak to the run_chain_tests script and associated docs so that we can run individual tests more easily (given that they seem to be fairly likely to fail in my case at least)

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -11,7 +11,7 @@
 
 ### Summary
 
-This script calls the local e2e-tests helper library in order to test the current branch or development enviornment against
+This script calls the local e2e-tests helper library in order to test the current branch or development environment against
 a collection of different blocks, across different runtimes. It does this for Polkadot, Kusama, and Westend.
 
 ### Requirements
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -11,13 +11,13 @@
 
 ### Summary
 
-This script calls the local e2e-tests helper library in order to test the current branch or development enviornment against 
-a collection of different blocks, across different runtimes. It does this for Polkadot, Kusama, and Westend. 
+This script calls the local e2e-tests helper library in order to test the current branch or development enviornment against
+a collection of different blocks, across different runtimes. It does this for Polkadot, Kusama, and Westend.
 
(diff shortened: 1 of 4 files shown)


scripts/run_chain_tests.py
```diff
@@ -26,8 +31,11 @@ def run_chain_test(chain):
         return -1
 
     os.environ["SAS_SUBSTRATE_WS_URL"] = url
-    proc = subprocess.Popen(["yarn", "dev"], stdout=subprocess.PIPE,
-                            stderr=subprocess.STDOUT)
+    proc = subprocess.Popen(
+        ["yarn", "dev"],
```
TarikGul, line 35:
```suggestion
        ["yarn", "build", "&&", "yarn", "start"],
```

Just realized this but I think it will also to help to have this not in dev mode so were not using tsc-watch, but instead just running the build itself.

TarikGul (approved):
LGTM, super solid addition!

scripts/run_chain_tests.py
jsdw: I don't think you can use `&&` in that case as it's a shell thing; we'd have to run them separately or use sh/bash like `["sh", "-c", "yarn build && yarn start"]`

scripts/README.md
```diff
@@ -11,13 +11,13 @@
 
 ### Summary
 
-This script calls the local e2e-tests helper library in order to test the current branch or development enviornment against 
-a collection of different blocks, across different runtimes. It does this for Polkadot, Kusama, and Westend. 
+This script calls the local e2e-tests helper library in order to test the current branch or development enviornment against
```
insipx, line 14:
```suggestion
This script calls the local e2e-tests helper library in order to test the current branch or development environment against
```

insipx (approved):

scripts/run_chain_tests.py
jsdw: I'll merge this as-is for now, but happy to make a change like the above (or just spawning both processes, one after the other) in a little separate PR if you like :)

PR #607, opened 2021-07-14
fix: update deps, resolutions

TarikGul:
This updates apps-config adds @polkadot/x-rxjs, and the resolutions in package.json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
     - @polkadot/util-crypto [release notes](https://github.com/polkadot-js/common/releases)
     - @substrate/calc [npm release page](https://www.npmjs.com/package/@substrate/calc)
 
-1. Next make sure to update the resolutions inside of the `package.json` to match polkadot-js [here](https://github.com/polkadot-js/apps/blob/master/package.json).
+1. Next make sure to update the resolutions inside of the `package.json` to match polkadot-js [here](https://github.com/polkadot-js/apps/blob/master/package.json). This is a manual process, and the packages we want to update are only the ones with `@polkadot` appended to it.
 
 1. After updating the dependencies and resolutions, the next step is making sure the release will work against all relevant runtimes for Polkadot, Kusama, and Westend. This can be handled by running `yarn test:init-e2e-tests`. You must have `python3`, and the dependencies inside of `./scripts/requirements.txt` installed to run the script (Read the [README](./scripts/README.md) for more instructions). Before moving forward ensure all tests pass, and if it warns of any missing types feel free to make an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,9 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
     - @polkadot/util-crypto [release notes](https://github.com/polkadot-js/common/releases)
     - @substrate/calc [npm release page](https://www.npmjs.com/package/@substrate/calc)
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -70,6 +71,16 @@
...
+    "@polkadot/api": "^5.0.1",
+    "@polkadot/api-contract": "^5.0.1",
+    "@polkadot/hw-ledger": "^7.0.1",
+    "@polkadot/keyring": "^7.0.1",
+    "@polkadot/networks": "^7.0.1",
+    "@polkadot/phishing": "^0.6.222",
+    "@polkadot/types": "^5.0.1",
+    "@polkadot/util": "^7.0.1",
+    "@polkadot/util-crypto": "^7.0.1",
+    "@polkadot/wasm-crypto": "^4.1.2",
```
jsdw, line 83: so if I recall, these will force the packaegs to be the versions listed.

What happens if two incompatible versions of a package are used throughout? I'd worry that this will lead to the thing expecting a different-enough version hitting errors and such?

TarikGul: So with our resolutions, we are grabbing them directly from the resolutions from @polkadot. That way they will never be mismatched.

jsdw: Should we update the release readme to add this step to make sure that they remain in sync?

jsdw: (I haven't thought through the other ramifications of doing this yet; it makes me feel a bit uneasy offhand, but I'm sure it's been given some thought!)

TarikGul: It's kinda a double edged sword too, because without this we potentially let multiple versions live inside of the node_modules which could also cause issue's. I shared the same concern as you, but after some digging I realized its probably best to mimic and shadow what `@polkadot` is doing which will atleast minimize the package conflictions that can potentially happen.

TarikGul: > Should we update the release readme to add this step to make sure that they remain in sync?

Yes thank you for the reminder this should be updated in the README.md

jsdw: Ok cool! Does the readme want tweaking as part of this PR too or are you happy doing that in a next step?

TarikGul: Yea just added the little bit on updating the resolutions, so you should be good :)

jsdw: Woop, thanks! Looks good!

README.md
```diff
@@ -290,7 +290,9 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
     - @polkadot/util-crypto [release notes](https://github.com/polkadot-js/common/releases)
     - @substrate/calc [npm release page](https://www.npmjs.com/package/@substrate/calc)
 
-1. After updating the dependencies, the next step is making sure the release will work against all relevant runtimes for Polkadot, Kusama, and Westend. This can be handled by running `yarn test:init-e2e-tests`. You must have `python3`, and the dependencies inside of `./scripts/requirements.txt` installed to run the script (Read the [README](./scripts/README.md) for more instructions). Before moving forward ensure all tests pass, and if it warns of any missing types feel free to make an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
+1. Next make sure to update the resolutions inside of the `package.json` to match polkadot-js [here](https://github.com/polkadot-js/apps/blob/master/package.json).
```
dvdplm, line 293: dq: how does one actually perform this check against their `package.js`. Is this a manual process?

dvdplm (approved):

README.md
TarikGul: Yea this would be a manual process. I should update that in the Readme as well. The idea is just too copy and paste the resolutions for the `@polkadot/*` packages that Jaco has set for `@polkadot/apps`.

dvdplm: ah got it, ty.

PR #613, opened 2021-07-15
fix: support for calculating fees for statemint, and statemine and their test nets. 

TarikGul:
Update the calc package to support the Multiplier for statemint and statemine and their testnets. Update the configs and separate statemint from statmine. Everything has been tested locally with the calc package.

After this PR is merged, it should be followed up by a PR to release and update the calc package.


calc/src/calc_fee.rs
```diff
@@ -49,6 +49,12 @@ impl Multiplier {
             ("westend", v) if 10 < v && v < 31 => V1((new_i128(inner), false)),
             ("westend", v) if 31 <= v => V2(new_u128(inner)),
 
+            ("statemine", _v) => V2(new_u128(inner)),
+            ("statemint", _v) => V2(new_u128(inner)),
+
+            ("westmine", _v) => V2(new_u128(inner)),
+            ("westmint", _v) => V2(new_u128(inner)),
```
dvdplm, line 56: Did you consider swapping around the logic a bit here; just check for the weird cases and give every one else `new_i128` by default?

dvdplm (approved):

calc/src/calc_fee.rs
```diff
@@ -49,6 +49,12 @@ impl Multiplier {
             ("westend", v) if 10 < v && v < 31 => V1((new_i128(inner), false)),
             ("westend", v) if 31 <= v => V2(new_u128(inner)),
 
+            ("statemine", _v) => V2(new_u128(inner)),
```
emostov, line 52: One idea here is to just assume all system parachains have the same fee calc as the current runtime of the relay chain .. not sure if this is a dangerous assumption. But if we are willing to make the assumption we could eliminate the new match logic and plug in the relay chain name and runtime version

calc/src/calc_fee.rs
TarikGul: I see what you are saying. Kinda like the equivalent of ending the `match` statement with a 

``` rust
/// account for defaults
_ => V1((new_i128(inner), false)),
```

I did think about this, and in all honesty I am kinda on the fence on whether it would be appropriate for `@substrate/calc` to give a default `Multiplier` or not. We enforce third party chains to update their own fee configuration through the controller-configs. Meaning if they don't have a `minCalcFeeRuntime` in the configuration they wont calculate fees anyways. Then meaning they would have to add their match case inside of the rust package as well which allows for zero room for error on our side or assumptions on another chains fee calculation.

But at the same time, as a package it makes sense for the `Multiplier` to have some kind of default value because then it can support more than just Polkadot, Kusama, parity testnets, and parity parachains.

calc/src/calc_fee.rs
TarikGul: I just asked DQ's, but I am not totally sure. I would side on the case that making the assumption can add potential work in the future but I am not sure if it is necessarily dangerous because we know the Multiplier it needs to use, and if it changes we'll have to update the code no matter what. I do think though that explicitly stating the `Multiplier` directly to the specName (in this case `statemine`) is good for clarity and readability.

TarikGul: Okay so based on this response:

> System parachains may be optimized for specific operations and could have different fee structures as a result. Also, unlike the relay chain, PoV size factors into the weight of a parachain block, which causes a fundamental difference between the relay chain and parachain fee structures

I would say it's probably best we keep it as is, also important for me to keep an eye on potential changes that could happen to the fee structure.

emostov (approved):

calc/src/calc_fee.rs
dvdplm: Ok, let's play it safe then. :)

Issue #612, opened 2021-07-14
GET /blocks/head is super slow, add alternative GET /blocks/head/summary


ngallagher:
**Proposed Change or Idea**

The /blocks/head method is really slow, it takes forever. Given that the polkadot blockchain mines blocks so quickly it would be nice if you could poll to check where the latest block is frequently without downloading data which may not be interesting.

I would suggest two additional methods.

***Existing***

GET /blocks/head
GET /blocks/:number

***Update***

GET /blocks/head
GET /blocks/head/summary
GET /blocks/:number
GET /blocks/:number/summary

For instance something like the following offers useful information for applications following the progress of the blockchain, without all of the other perhaps less interesting information (depending on requirements)

```
	async fetchBlockSummary(targetHash?: BlockHash): Promise<IBlockSummary> {
		const { api } = this;
		const signedBlock = (targetHash ? await api.rpc.chain.getBlock(targetHash) : await api.rpc.chain.getBlock())
		const block = signedBlock?.block;
		const {
			hash,
			parentHash,
			number,
			stateRoot,
			extrinsicsRoot,
		} = block?.header;			

		return {
			number,
			hash,
			parentHash,
			stateRoot,
			extrinsicsRoot
		};
	}
```

TarikGul:
Yea the `/blocks/head` unfortunately does take some to complete a request, especially if there is a lot of data to parse through and format. I would be interested in putting something like this together, get a branch up, and some benchmarks to see what kind of performance upgrade something like this would yield as well. 

cc: @joepetrowski @emostov @dvdplm 

Any thoughts?

joepetrowski:
Sounds reasonable to me.

ngallagher:
Another idea is perhaps use an LRU cache to avoid repeating the heavy RPC calles for GET /blocks/head, as long as it does not change the behaviour.

emostov:
Yeah, endpoint for just the header sounds reasonable.

But I do think there are probably some easy optimizations we can add in with query params to filter out some fields like block author, finality, and even skipping the events query.

I made an issue for documenting how to improve endpoint performance just with pre-existing query params: https://github.com/paritytech/substrate-api-sidecar/issues/489

emostov:
> Another idea is perhaps use an LRU cache to avoid repeating the heavy RPC calles for GET /blocks/head, as long as it does not change the behaviour.

This sounds worth investigating. Are you thinking of situations where the user polls `blocks/head`?

[...]

PR #615, opened 2021-07-20
feat: add `/blocks/:number/header`, and `/blocks/head/header`

TarikGul:
rel: [#612](https://github.com/paritytech/substrate-api-sidecar/issues/612)

## `/blocks/:number/header`

**Summary:** When a block is queried by its block identifier ie: (BlockHash, or BlockNumber),it will return the Header of that block as a summary of its information.

**Returns:**

```javascript
{
  parentHash: string; // type => Hash
  number: string; // type => Compact<BlockNumber>
  stateRoot: string; // type => Hash
  extrinsicsRoot: string; // type => Hash
  digest: Struct { logs: Vec<DigestItem> }; // type => Digest
}
```

## `/blocks/head/header?finalized=boolean`

**Summary:** When a block is queried by its head the response will be the header of the most recent finalized block, but if you want the most recent block that may not be finalized you may add the query parameter `?finalized=false`. It will default to true

**Returns:**

```javascript
{
  parentHash: string; // type => Hash
  number: string; // type => Compact<BlockNumber>
  stateRoot: string; // type => Hash
  extrinsicsRoot: string; // type => Hash
  digest: Struct { logs: Vec<DigestItem> }; // type => Digest
}
```

Todo:

- [X] Tests
- [X] Endpoints
- [X] Docs

Note: I also wrote a naive version that leverages the fetchblock to create a summary endpoint. That can be found on the branch `tarik-blockhead-summary-naive`. I prefer this branch and version though.


jsdw (approved):
I don't have much experience with this codebase, but the changes look good to me offhand :)

(I wonder whether there's a way to reuse the schema pattern in the OpenAPI spec; offhand I don't think it's possible, but I'm rusty with OpenAPI!)

src/controllers/blocks/BlocksController.ts
```diff
@@ -169,4 +166,77 @@ export default class BlocksController extends AbstractController<BlocksService>
...
 		);
 	};
+
+	/**
+	 * Return the Header of a block when a block number is queried
+	 *
+	 * @param req Express Request
+	 * @param res Express Response
+	 */
+	private getBlockByIdSummary: RequestHandler<INumberParam> = async (
```
emostov, line 176:
```suggestion
	private getBlockSummaryById: RequestHandler<INumberParam> = async (
```

emostov: or `getBlockHeaderById`

src/controllers/blocks/BlocksController.ts
```diff
@@ -169,4 +166,77 @@ export default class BlocksController extends AbstractController<BlocksService>
 			await this.service.fetchBlock(hash, options)
 		);
 	};
+
+	/**
+	 * Return the Header of a block when a block number is queried
```
emostov, line 171:
```suggestion
	 * Return the Header of the identified block.
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -169,4 +166,77 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+	 * Return the header of the latest block
+	 *
+	 * @param req Express Request
+	 * @param res Express Response
+	 */
+	private getLatestBlockSummary: RequestHandler = async (
+		{ query: { finalized } },
+		res
+	): Promise<void> => {
+		const paramFinalized = finalized === 'true' ? true : false;
```
emostov, line 198:
```suggestion
		const paramFinalized = finalized !== 'false';
```
If we want the default behavior to get the finalized head I think we should do this: if finalized exactly equals false, we don't get the non-finalized head; otherwise get the finalized head.

src/controllers/blocks/BlocksController.ts
```diff
@@ -169,4 +166,77 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+
+		const hash = !paramFinalized
+			? await this.api.rpc.chain.getFinalizedHead()
```
emostov, line 201: This query gets the finalized head, but the way I read it we do this query when we _don't_ want the finalized head

src/controllers/blocks/BlocksController.ts
```diff
@@ -169,4 +166,77 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+	 * @param req Express Request
+	 * @param res Express Response
+	 */
+	private getLatestBlockSummary: RequestHandler = async (
+		{ query: { finalized } },
+		res
+	): Promise<void> => {
+		const paramFinalized = finalized === 'true' ? true : false;
+
+		const hash = !paramFinalized
```
emostov, line 200:
```suggestion
		const hash = paramFinalized
```

emostov:
Have we considered calling the endpoint `/blocks/latests/header`? I suppose with summary we can put more stuff in later, but atm header is more accurate. @joepetrowski may have an opinion here

src/controllers/blocks/BlocksController.ts
TarikGul: great catch. Resolved.

src/controllers/blocks/BlocksController.ts
TarikGul: I like the header notion here! for now I am going to switch it all over to Header, I think it makes more sense since that is the exact Type we are returning.

TarikGul:
I agree that `/header` is a better endpoint name. I switched it all over.

emostov (approved):

Issue #611, opened 2021-07-14
e2e python script needs some work.


TarikGul:
Few things need to be added to that script or changed.

1. Read the output of sidecar, and when the proper logging is logged, then we can move forward instead of the hardcoded 30 second timeout.

2. Log the STDOUT

3. Wrap the main script in a try catch.

4. Catch kill signals (ctrl+c)

jsdw:
And potentially: catch kill signals (ctrl+c) and kill running processes then also if needbe?

I wonder whether there's a smarter way to have the processes clean up after themselves; it'd be ideal if it just happened pretty automatically.

TarikGul:
Thanks missed the catck kill signals, just added it. 

> I wonder whether there's a smarter way to have the processes clean up after themselves; it'd be ideal if it just happened pretty automatically.

I agree that it should be as easy and smooth as one command, and there shouldn't be any unexpected results.

PR #618, opened 2021-07-23
fix: rewrite sidecar e2e script

TarikGul:
closes: [#611](https://github.com/paritytech/substrate-api-sidecar/issues/611)

Rewritten e2e tests script in typescript.


scripts/config.ts
```diff
@@ -0,0 +1,41 @@
...
+	process: 'sidecar',
+	resolver: 'Build Finished',
+	args: ['build'],
+};
+
+export const config = {
+	polkadot: {
+		wsUrl: 'wss://rpc.polkadot.io',
+		JestProcOpts: {
+			process: 'jest',
```
TarikGul, line 17: Can be more DRY.

jsdw (approved):
Cool! Offhand this looks like a definite step up from the python approach (and I for one like that the python dependency goes away :D)

Does the README need any tweaks now that python is no longer a dependency and such?

scripts/runChainTests.ts
```diff
@@ -0,0 +1,124 @@
+import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+
+import { config, defaultSasBuildOpts } from './config';
+
+type StatusCode = '0' | '1';
```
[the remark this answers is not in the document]
TarikGul: I totally agree! Will update it. Thanks

TarikGul:
@jsdw Yea exactly. I need to update all the docs, and the scripts

TarikGul:
@jsdw I changed it up quite a bit if you want to take a second check. There were still some node processes that would not be handled correctly on random tests, so I ended up spawning them in their own groups and using pid ranges to kill all child processes and any possible descendants. All issues are completely handled now :)

scripts/.eslintrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('@substrate/dev/config/eslint');
```
emostov, line 1: Is eslint smart enough to look in the directory above?

scripts/runChainTests.ts
```diff
@@ -0,0 +1,195 @@
...
+};
+
+/**
+ * Launch any given process. It accepts an options object.
+ *
+ * {
+ *   proc => the name of the process to be saved in our cache
+ *   resolver => string: If the stdout contains the resolver it will resolve the process
+ *   resolverStartupErr => string: If the stderr contains the resolver it will resolve the process
+ *   args => an array of args to be attached to the `yarn` command.
```
emostov, line 22:
```suggestion
 *   proc string => the name of the process to be saved in our cache
 *   resolver string => If the stdout contains the resolver it will resolve the process
 *   resolverStartupErr string => If the stderr contains the resolver it will resolve the process
 *   args string[] => an array of args to be attached to the `yarn` command.
```
Or remove types

scripts/runChainTests.ts
```diff
@@ -0,0 +1,195 @@
...
+		console.error('Error launching sidecar... exiting...');
+		killAll();
+		process.exit(1);
+	}
+};
+
+// Kill all processes spawned and tracked by this file.
+const killAll = () => {
+	console.log('Killing all processes...');
+	for (const key of Object.keys(procs)) {
```
emostov, line 105: I don't fully understand the considerations here, but is there any benefit in have `procs` as a param to this function? Otherwise it is a global-ish object for this module, which is fine but easy to abuse

scripts/runChainTests.ts
```diff
@@ -0,0 +1,195 @@
...
+		if (!procs[key].killed) {
+			try {
+				console.log(`Killing ${key}`);
+				// Kill child and all its descendants.
+				process.kill(-procs[key].pid, 'SIGTERM');
+				process.kill(-procs[key].pid, 'SIGKILL');
+			} catch (e) {
+				/**
```
emostov, line 113: could console.error the message, but up to you

scripts/types.ts
```diff
@@ -0,0 +1,21 @@
+// 0 => success && 1 => failure
+// export type StatusCode = '0' | '1';
+
```
emostov, line 3:
```suggestion
```

scripts/runChainTests.ts
TarikGul: Yea I agree, good point. Will remove them.

scripts/runChainTests.ts
TarikGul: Yea i thought about it too, but it will throw error every time when using `-procs[key].pid` because there will be some 'pid' that sidecar already closed implicitly but is still within the group. The error is harmless, but the major thing is just making sure all process's close.

scripts/runChainTests.ts
TarikGul: > is there any benefit in have procs as a param to this function

I dont think in this situation there is much benefit. Preferably the script is completely stand alone, and doesn't interact with anything else which would stop any collisions or issues happening with `procs`.

scripts/.eslintrc.js
TarikGul: Yes! 100% is. TIL :)

removed all the eslint configs from the scripts folder.

TarikGul: except the eslintignore

insipx (approved):
LGTM, scripts ran on the first try, looks like a good upgrade

README.md
```diff
@@ -292,9 +292,9 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
...
+1. After updating the dependencies and resolutions (if applicable), the next step is making sure the release will work against all relevant runtimes for Polkadot, Kusama, and Westend. This can be handled by running `yarn test:init-e2e-tests`. If you would like to test on an individual chain, you may run the same command followed by its chain, ex: `yarn test:init-e2e-tests:polkadot`. Before moving forward ensure all tests pass, and if it warns of any missing types feel free to make an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
 
-    Note that the e2e tests will connect to running nodes in order to test sidecar against real data, and they may fail owing to those connections taking too long to establish. If you run into any failures, try running tests just for the chain that failed with something like `yarn test:init-e2e-tests:polkadot`.
+    Note: that the e2e tests will connect to running nodes in order to test sidecar against real data, and they may fail owing to those connections taking too long to establish. If you run into any failures, try running tests just for the chain that failed with something like `yarn test:init-e2e-tests:polkadot`.
```
jsdw, line 297: Nit: Maybe we don't need to repeat the per-chain bit now it's added above (but no harm in it really!)

scripts/runChainTests.ts
```diff
@@ -0,0 +1,195 @@
...
+parser.add_argument('--chain', {
+	choices: ['polkadot', 'kusama', 'westend'],
+});
+
+const args = parser.parse_args() as Namespace;
+
+/**
+ * Signal interrupt
+ */
+process.on('SIGINT', function () {
```
jsdw, line 189: I wonder whether iot's worth covering a few more bases and also cleaning up when SIGKILL, SIGTERM (eg when kill command used) and SIGHUP (eg when terminal window closed) are sent?

jsdw (approved):
Looks great! I tried it and, for the first time ever, all tests passed without a hitch. I'll consider that a win :D

README.md
TarikGul: It might be worth keeping for now, but yea I am happy to remove it if it seems like extra bloat.

scripts/runChainTests.ts
TarikGul: Yea I agree, I just added SIGHUP. 

> cleaning up when SIGKILL, SIGTERM (eg when kill command used)

Curious what you mean by that?

jsdw: Oh, I just meant that SIGTERM is the default signal sent when `kill` is used. Actually, I don't think you can catch SIGKILL looking at the `kill` man page so that's not useful anyway!

jsdw: (I thnk that catching the result of ctrl+c is most important though and you've done that, and now you've added the "terminal" closed signal that seems good to me!)

TarikGul: Ahh okay cool i see what you are referencing at :)

Issue #484, opened 2021-03-22
Add a `--version` flag


emostov:
**Proposed Change or Idea**

Add a `--version` flag that simply returns the version of substrate-api-sidecar

**Alternatives**

Recently, #480 added a log message with the version. It could be left at that.

PR #620, opened 2021-07-24
fix: add --version flag

TarikGul:
closes: [#484](https://github.com/paritytech/substrate-api-sidecar/issues/484)

This adds the version flag for sidecar, but currently there is some unknown default behavior that should be resolved before this is merged. 

Run: `yarn build && yarn start --version`

Current output:

```
SAS:
  📦 LOG:
     ✅ LEVEL: "info"
     ✅ JSON: false
     ✅ FILTER_RPC: false
     ✅ STRIP_ANSI: false
  📦 SUBSTRATE:
     ✅ WS_URL: "wss://rpc.polkadot.io"
     ✅ TYPES_BUNDLE: undefined
     ✅ TYPES_CHAIN: undefined
     ✅ TYPES_SPEC: undefined
     ✅ TYPES: undefined
  📦 EXPRESS:
     ✅ BIND_HOST: "127.0.0.1"
     ✅ PORT: 8080
@substrate/api-sidecar v9.0.0
```

I am under the impression that [this](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/logging/Log.ts#L6) could be the root of the issue, but I need to do some more digging. 

It originated from [this import](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/main.ts#L27), then trickles down to `./controllers` all the way to anywhere `validateAddress` is imported in, such as all the `Account` Controllers.

 
 Update: 
 
 This is fixed, and working as it should!


src/main.ts
```diff
@@ -23,8 +23,10 @@ import { OverrideBundleType, RegistryTypes } from '@polkadot/types/types';
 import { json } from 'express';
 
 import packageJSON from '../package.json';
+import { version } from '../package.json';
```
TarikGul, line 26: remove and use `packageJSON`

src/Cli.ts
```diff
@@ -0,0 +1,9 @@
+import { ArgumentParser, Namespace } from 'argparse';
+
+export const parseArgs = (): Namespace => {
+	const parser = new ArgumentParser();
+
+	parser.add_argument('-v', '--version', { action: 'store_true' });
+
+	return parser.parse_args() as Namespace;
+};
```
emostov, line 9: maybe rename file `cli` or `parseArgs`? `Cli` makes me think its a class with that name

emostov:
> but currently there is some unknown default behavior that should be resolved before this is merged.

Whats the exact behavior that is unexpected?

src/Cli.ts
TarikGul: Yea i totally agree. I'll change it to lowercase.

TarikGul:
@emostov The "unknown" issue is just SAS is logging/printing the below by just importing files. As in without initializing anything this will always print if `main.ts` is called. 

```
SAS:
  📦 LOG:
     ✅ LEVEL: "info"
     ✅ JSON: false
     ✅ FILTER_RPC: false
     ✅ STRIP_ANSI: false
  📦 SUBSTRATE:
     ✅ WS_URL: "wss://rpc.polkadot.io"
     ✅ TYPES_BUNDLE: undefined
     ✅ TYPES_CHAIN: undefined
     ✅ TYPES_SPEC: undefined
     ✅ TYPES: undefined
  📦 EXPRESS:
     ✅ BIND_HOST: "127.0.0.1"
     ✅ PORT: 8080
```

emostov:
@TarikGul try moving this line inside of main:

https://github.com/paritytech/substrate-api-sidecar/blob/5a7890311ab6ae87b40c371353d5bad25bba7754/src/main.ts#L34

I think its calling the getter here:

https://github.com/paritytech/substrate-api-sidecar/blob/5a7890311ab6ae87b40c371353d5bad25bba7754/src/SidecarConfig.ts#L70-L72

TarikGul:
> @TarikGul try moving this line inside of main:
> 
> https://github.com/paritytech/substrate-api-sidecar/blob/5a7890311ab6ae87b40c371353d5bad25bba7754/src/main.ts#L34
> 
> I think its calling the getter here:
> 
> https://github.com/paritytech/substrate-api-sidecar/blob/5a7890311ab6ae87b40c371353d5bad25bba7754/src/SidecarConfig.ts#L70-L72

Yea I had tried this earlier when trying to debug it and it didn't work, that could be another source of where it could be coming from though, but I also can confirm that this file specifically [this line](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/logging/Log.ts#L6). 

But that being said it I think it could be initializing in more than one place too.

TarikGul:
@emostov I did make some progress on it, looks like you were right in one respect. `consoleTransport` had the same issue with `SidecarConfig` which you pointed out in main. So three cases where it shows up is now solved (main.ts, consoleTransport.ts). The issue is still coming up, but I think I should have it solved shortly after having those 3 sorted out.

emostov (approved):
💯

jsdw (approved):
Code looks good to me, and it does what it says on the tin :)

src/main.ts
```diff
@@ -150,3 +146,17 @@ function startUpPrompt(wsUrl: string, chainName: string, implName: string) {
 		);
 	}
 }
+
+process.on('SIGINT', function () {
+	console.log('Caught interrupt signal, exiting...');
+	process.exit(0);
+});
```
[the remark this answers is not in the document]
TarikGul: You're exactly correct here, its more for the exit code, and the console to the terminal.

PR #625, opened 2021-07-30
fix: Added SORA network controller

pgolovkin:
Added SORA network controller


src/chains-config/soraControllers.ts
```diff
@@ -0,0 +1,33 @@
+import { ControllerConfig } from '../types/chains-config';
+
+/**
+ * SORA configuration fro sidecar.
```
TarikGul, line 4:
```suggestion
 * SORA configuration for sidecar.
```

TarikGul (approved):
lgtm, I tested it against `wss://mof3.sora.org` and it all checked out. Just one minor grumble with a comment spelling error.

TarikGul:
Also before merging, @pgolovkin if you could run `yarn lint --fix` to fix the CI check. 

Thanks again.

src/chains-config/soraControllers.ts
pgolovkin: fixed, thanks!

insipx (approved):

PR #631, opened 2021-08-09
fix: finalization for `/blocks/head`

TarikGul:
This fixes the default behavior for querying the finalized head. The query param and its expected behavior still works. The only issue is the `/blocks/head` endpoint is not returning the finalized head by default, which means you have to tag the finalized param onto the query. This fixes the slight bug introduced in this PR [#615](https://github.com/paritytech/substrate-api-sidecar/pull/615).

### DIFF
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -115,8 +115,8 @@ export default class BlocksController extends AbstractController<BlocksService> 
 			hash = (await this.api.rpc.chain.getHeader()).hash;
 		} else if (finalized === 'false') {
 			// We query the finalized head to know where the latest finalized block
-			// is. It is a way to confirm whether the queried block is less than or greater
-			// than the finalized head.
+			// is. It is a way to confirm whether the queried block is less than or
+			// equal to the finalized head.
 			omitFinalizedTag = false;
 			queryFinalizedHead = true;
 			hash = (await this.api.rpc.chain.getHeader()).hash;
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -115,7 +115,10 @@ export default class BlocksController extends AbstractController<BlocksService> 
 		const extrinsicDocsArg = extrinsicDocs === 'true';
 
 		const { hash, queryFinalizedHead, omitFinalizedTag } =
-			await this.parseFinalizationOpts(this.options.finalizes, finalized as string);
+			await this.parseFinalizationOpts(
+				this.options.finalizes,
+				finalized as string
+			);
 
 		const options = {
 			eventDocs: eventDocsArg,
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -114,6 +114,9 @@ export default class BlocksController extends AbstractController<BlocksService> 
 			queryFinalizedHead = false;
 			hash = (await this.api.rpc.chain.getHeader()).hash;
 		} else if (finalized === 'false') {
+			// We query the finalized head to know where the latest finalized block
+			// is. It is a way to confirm whether the queried block is less than or greater
+			// than the finalized head. 
 			omitFinalizedTag = false;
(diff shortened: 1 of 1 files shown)


src/controllers/blocks/BlocksController.ts
```diff
@@ -215,15 +217,15 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+		paramFinalized: string
 	): Promise<IFinalizationOpts> => {
 		let hash, queryFinalizedHead, omitFinalizedTag;
 		if (!optFinalizes) {
 			// If the network chain doesn't finalize blocks, we dont want a finalized tag.
 			omitFinalizedTag = true;
 			queryFinalizedHead = false;
 			hash = (await this.api.rpc.chain.getHeader()).hash;
-		} else if (!paramFinalized) {
+		} else if (paramFinalized === 'false') {
```
niklasad1, line 228: oh, this check is changed just because it is a `String` now?

src/controllers/blocks/BlocksController.ts
```diff
@@ -114,9 +114,11 @@ export default class BlocksController extends AbstractController<BlocksService>
 		const eventDocsArg = eventDocs === 'true';
 		const extrinsicDocsArg = extrinsicDocs === 'true';
 
-		const paramFinalized = finalized === 'true' ? true : false;
 		const { hash, queryFinalizedHead, omitFinalizedTag } =
-			await this.parseFinalizationOpts(this.options.finalizes, paramFinalized);
+			await this.parseFinalizationOpts(
```
niklasad1, line 118: I don't get this, how does it fix it by converting passing it as a `String` rather than a `boolean`?

src/controllers/blocks/BlocksController.ts
TarikGul: So initially before i abstracted this logic out it was always a string. Then I abstracted out of the main controller logic, and wanted to get it more organized, which made the logic more convoluted and confusing which I think is what introduced the mix in logic. So for simplicity sake, and making sure it works, I am reverting it back to what it was originally.

src/controllers/blocks/BlocksController.ts
TarikGul: hmmm, you make a good point that it is confusing too outside eyes as to what is happening, so I am going to just revert it completely to what it use to be so that there is no confusion.

jsdw: I'm a bit confused about the string usage; can the string just be `true` or `false` and, why a string and not a bool? I'm probably missing too much context really :)

Sounds like you'll revert anyway so maybe I don't need to know :D

TarikGul: @niklasad1 @jsdw Okay, so I reverted it back to what it was originally before I abstracted out of the controller. Hopefully it's a little clearer

insipx (commented):
So IIUC, `finalizes` checks if the network itself supports finalization, and `finalized` is a user parameter, but I'm not sure what the use of `FinalizedTag`/`queryFinalizedHead` is. What does `FinalizedTag`do that `queryFinalizedHead`does not? I'm trying to understand why we use `queryFinalizedHead` with the `chain.getHeader()` func. Or why when we use `getHeader`when `queryFinalized` head is false, and `getFinalizedHead` when `queryFinalizedHead` is true.

Otherwise I think this looks good.

TarikGul:
@insipx Great question, I just added a comment to clarify. But the idea behind `queryFinalizedHead` is to confirm that the queried blockhead is either finalized or not. We do that by comparing the results of the finalized head blocks height, and see if it is greater than or equal to the blockhead that's queried. That is why we have the `queryFinalizedHead` boolean. (The naming can be confusing).

The logic where this all takes place when this is queried (`/blocks/head?finalized=false`) is [here](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/blocks/BlocksService.ts#L747)

insipx (approved):
makes sense to me :+1:

CurlyBracketEffect (approved):

PR #637, opened 2021-08-18
fix: imports for metadata and tests, update deps

TarikGul:
This PR updates the deps, and fixes the imports from polkadot-js regarding metdata. Residually some tests had to be reconfigured.


src/services/blocks/BlocksService.spec.ts
```diff
@@ -356,40 +356,11 @@ describe('BlocksService', () => {
...
-			});
-
-			const sudo = createCall('sudo', 'sudo', {
-				call: proxy,
-			});
-
+		it('handles a batch transfer correctly', () => {
 			const batch = createCall('utility', 'batch', {
-				calls: [sudo, sudo, sudo],
+				calls: [],
```
TarikGul, line 361: This test just needs to be updated

insipx (approved):

src/services/blocks/BlocksService.spec.ts
TarikGul: fixed this test, inputted proxy calls

emostov (approved):

Issue #640, opened 2021-08-19
GET /experimental/paras/ return error 500


CrommVardek:
Hello, not a huge deal, since it is on an experimental endpoint, but I'd still point it out: 

**Description**

Sidecar version : 9.1.4 (latest docker image)
Connected node is a Polkadot Node, v0.9.8 (tested on latest block which is 6452683 at the time of the test)

GET on /experimental/paras return the following : 
{"code":500,"message":"Cannot read property 'paraLifecycles' of undefined","stack":"TypeError: Cannot read property 'paraLifecycles' of undefined\n    at ParasService.paras (/usr/src/app/build/src/services/paras/ParasService.js:256:34)\n    at ParasController.getParas (/usr/src/app/build/src/controllers/paras/ParasController.js:13:67)\n    at processTicksAndRejections (node:internal/process/task_queues:94:5)\n    at async /usr/src/app/build/src/controllers/AbstractController.js:159:9","level":"error"}

Which is not really expected. 

**Steps to Reproduce**

Running latest docker image of sidecar.

HTTP GET on /experimental/paras

**Expected vs. Actual Behavior**

Expected response from GET : 

```
{
  "at": {
    "hash": "0xdb4c18773a415e9d12ccf4a49dcc247dfc2e89a36b700ad46d1bba9cabfffa85",
    "height": "6452683"
  },
  "paras": []
}
```

Actual response (see 1st paragraph).

TarikGul:
Hey thanks for reporting the issue.

The issue here is you are connected to a polkadot node which doesn't have parachain's active yet. Therefore the runtime metadata doesn't contain the types necessary to query any of the `paras` endpoints. 

If you connect to `wss://kusama-rpc.polkadot.io` you'll see that the endpoints start working as should. The idea is once parachains are active on polkadot to remove it out of `experimental`. We could also remove the `Paras` controller out of polkadot config until it is active on the relay chain so there is less confusion.

CrommVardek:
Thanks for the confirmation, it was my guess that the exception occurred because Polkadot does not have parachains yet. It's more that the error code is not helpful in understanding the really reason behind it. As Polkadot's parachains will come in the next few months I'm not sure it's that important to remove Paras for polkadot config. But it's your call.

TarikGul:
So after some consideration I agree that removing it from the config is not really useful. 

> it's more that the error code is not helpful in understanding the really reason behind it. 

[...]

PR #641, opened 2021-08-19
fix: remove polkadot config for paras controller

TarikGul:
closes: [#640](https://github.com/paritytech/substrate-api-sidecar/issues/640)

Remove `Paras` controller from the polkadot chains-config until it's time to start testing parachains against polkadot before it's launched. I think this will cause less confusion for others, and is the most straight forward way to address the issue of anyone trying to use the endpoint.


TarikGul:
After some discussion I am going to close this PR, with the paras endpoint as experimental currently, and polkadot not having yet supporting parachains, the errors received are a none issue.

PR #648, opened 2021-08-25
improve security of the docker container

chevdor:
details to come

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,5 +36,6 @@ COPY --from=builder /opt/builder /usr/src/app
 ENV SAS_EXPRESS_PORT=8080
 ENV SAS_EXPRESS_BIND_HOST=0.0.0.0
 
+USER sidecar
 EXPOSE ${SAS_EXPRESS_PORT}
 CMD [ "node", "build/src/main.js" ]


README.md
```diff
@@ -252,12 +252,14 @@ yarn build:docker
...
 # For default use run:
-docker run --rm -it -p 8080:8080 substrate-api-sidecar
+docker run --rm -it --read-only -p 8080:8080 substrate-api-sidecar
 
 # Or if you want to use environment variables set in `.env.docker`, run:
-docker run --rm -it --env-file .env.docker -p 8080:8080 substrate-api-sidecar
+docker run --rm -it --read-only --env-file .env.docker -p 8080:8080 substrate-api-sidecar
 ```
 
+**NOTE**: While you could ommit the `--read-only` flag, it is **strongly recommended for containers used in production**.
```
jsdw, line 261:
```suggestion
**NOTE**: While you could omit the `--read-only` flag, it is **strongly recommended for containers used in production**.
```

docker-compose.yml
```diff
@@ -7,11 +7,12 @@ services:
       - ~/polkadot-data:/data
     # ports:
     #   - "9944:9944"
-    command: 'polkadot --chain polkadot --unsafe-ws-external --rpc-cors=all'
+    command: '--chain polkadot --unsafe-ws-external --rpc-cors=all'
```
jsdw, line 10: Did this line work at all previously (or did older images not default to running `polkadot`)?

jsdw (approved):
LGTM!

I haven't tested the `justfile`, but I can see that you have, and the other changes look fairly trivial assuming the `docker-compose` command is all gravy :)

docker-compose.yml
chevdor: I did not check but it is **very** likely that I initially used my `chevdor/polkadot` image here. The difference with the parity one being that the parity image uses `polkadot` as entrypoint and my image does not. I guess someone updated from `chevdor` to `parity` without changing the call.

So the only 2 options are:
- use `chevdor/polkadot` **and** start  the command with `polkadot`
- use `parity/polkadot` and omit the `polkadot` binary in the command

My image is not built from CI so I do NOT recommend any longer using it (expect for people who want to build their own image themself).

TarikGul (approved):
LGTM

> While this PR makes the Docker image and the container deployed with it more secure, just using the new image does not get all the benefits if the user is not running the container as --read-only. This is why this flag as been set as default in the doc and in the docker-compose. It may be wise to let our known users know about this.

👍

Issue #634, opened 2021-08-12
Cannot query historical balances on Kusama


pepyakin:
I queried substrate-api-sidecar for historical account and it returned an error. My polkadot node is of the 0.9.8 release running kusama.

The query was:

`GET /accounts/{accountId}/balance-info?at=0x3361f6446c72aaa6679aa40a03e493a851161fd08cb53237968a9967902691c0`

The result was:

```
{
  "code": 500,
  "message": "Unable to decode storage system.account:: createType(AccountInfo):: {\"nonce\":\"Index\",\"refcount\":\"RefCount\",\"data\":\"AccountData\"}:: Decoded input doesn't match input, received 0x0000000000000000000000000000000000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (80 bytes), created 0x0000000000000000000000000000000000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (69 bytes)",
  "stack": "Error: Unable to decode storage system.account:: createType(AccountInfo):: {\"nonce\":\"Index\",\"refcount\":\"RefCount\",\"data\":\"AccountData\"}:: Decoded input doesn't match input, received 0x0000000000000000000000000000000000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (80 bytes), created 0x0000000000000000000000000000000000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (69 bytes)\n    at RpcCore._newType (/home/lilpep/dev/substrate-api-sidecar/node_modules/@polkadot/rpc-core/bundle.cjs:450:13)\n    at RpcCore._formatStorageData (/home/lilpep/dev/substrate-api-sidecar/node_modules/@polkadot/rpc-core/bundle.cjs:399:17)\n    at RpcCore._formatOutput (/home/lilpep/dev/substrate-api-sidecar/node_modules/@polkadot/rpc-core/bundle.cjs:376:19)\n    at callWithRegistry (/home/lilpep/dev/substrate-api-sidecar/node_modules/@polkadot/rpc-core/bundle.cjs:255:42)\n    at processTicksAndRejections (internal/process/task_queues.js:93:5)",
  "level": "error"
}
```

Sidecar revision is: 5a3023b12a4282852afc039cf22b84f0f5207425

TarikGul:
[...]

PR #653, opened 2021-09-01
fix: historical queries for account balances

TarikGul:
closes: [#634](https://github.com/paritytech/substrate-api-sidecar/issues/634)

This fixes historical queries for `/accounts/{accountId}/balance-info`.

### DIFF
--- a/src/services/accounts/AccountsBalanceInfoService.spec.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.spec.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/require-await */
 import { AccountInfo, Address, Hash } from '@polkadot/types/interfaces';
 
 import { sanitizeNumbers } from '../../sanitize/sanitizeNumbers';
@@ -19,24 +20,13 @@ const accountsBalanceInfoService = new AccountsBalanceInfoService(mockApi);
 const historicalMockApi = {
 	query: {
 		balances: {
-			freeBalance: () =>
-				Promise.resolve().then(() => {
-					return polkadotRegistry.createType('Balance', 123456789);
-				}),
-			reservedBalance: () =>
-				Promise.resolve().then(() => {
-					return polkadotRegistry.createType('Balance', 1);
-				}),
-			locks: () =>
-				Promise.resolve().then(() => {
-					return polkadotRegistry.createType('Vec<BalanceLock>', []);
-				}),
+			freeBalance: async () =>
+				polkadotRegistry.createType('Balance', 123456789),
+			reservedBalance: async () => polkadotRegistry.createType('Balance', 1),
+			locks: async () => polkadotRegistry.createType('Vec<BalanceLock>', []),
 		},
 		system: {
-			accountNonce: () =>
-				Promise.resolve().then(() => {
-					return polkadotRegistry.createType('Index', 1);
-				}),
+			accountNonce: async () => polkadotRegistry.createType('Index', 1),
 		},
 	},
 };
@@ -171,10 +161,8 @@ describe('AccountsBalanceInfoService', () => {
 
 				expect(sanitizeNumbers(result)).toStrictEqual(expectedResponse);
 
-				mockApi.at = () =>
-					Promise.resolve().then(() => {
-						return createApiWithAugmentations(polkadotMetadata.toHex());
-					});
+				mockApi.at = async () =>
+					createApiWithAugmentations(polkadotMetadata.toHex());
 			});
 		});
 	});
--- a/src/services/accounts/AccountsBalanceInfoService.spec.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.spec.ts
@@ -4,11 +4,42 @@
 import { AccountInfo, Address, Hash } from '@polkadot/types/interfaces';
 
 import { sanitizeNumbers } from '../../sanitize/sanitizeNumbers';
+import { polkadotMetadata } from '../../test-helpers/metadata/metadata';
+import { polkadotRegistry } from '../../test-helpers/registries';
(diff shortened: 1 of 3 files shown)


insipx (approved):

TarikGul:
I need to confirm that this works for all runtimes before block ~1500000, currently they all work except when querying block 1400000's runtime.

TarikGul:
All historical runtimes are now covered.

The initial issue was present in two cases: For example, when querying Kusama, the following 2 groups of runtimes have different storage formats. 

v1054 --> v1050
v1049 --> ... (earliest runtime)

--------

Presently we use 2 formats to get account storage information:

a) `api.query.system.accounts.at(hash, address)`
b) `api.query.balances.account.at(hash, address)`

But for runtime v1054 --> v1050 the api requires a historicalApi via `await api.at(`${BlockHash}`)`

`historicalApi.query.system.account(address)`

And for runtime v1049 --> ...(earliest runtime) also requires a historicalApi

`historicalApi.query.balances.freeBalance(address)`,
`historicalApi.query.balances.locks(address)`,
`historicalApi.query.balances.reservedBalance(address)`,
`historicalApi.query.system.accountNonce(address)`,

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,6 +27,88 @@ export class AccountsBalanceInfoService extends AbstractService {
...
 
+		// Retrieve the historic api for the given hash
+		const historicalApi = await api.at(hash);
+
+		/**
+		 * Check two different cases where a historicalApi is needed in order
+		 * to have the correct runtime methods.
+		 *
+		 * a) Does the block use the oldest api where the free balance is found
+		 * using `historicalApi.query.balances.freeBalance`
```
emostov, line 38:
```suggestion
		 * using `historicalApi.query.balances.freeBalance`.
```

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,6 +27,88 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+				historicalApi.query.balances.locks(address),
+				historicalApi.query.balances.reservedBalance(
+					address
+				) as Promise<Balance>,
+				historicalApi.query.system.accountNonce(address) as Promise<Index>,
+			]);
+
+			// Values dont exist for these historic runtimes
+			const miscFrozen = api.registry.createType('Balance', 0),
+				feeFrozen = api.registry.createType('Balance', 0);
```
emostov, line 56: TIL you can do this with a `cont`

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,6 +27,88 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+
+			const at = {
+				hash,
+				height: header.number.toNumber().toString(10),
+			};
+
+			if (locks && free && reserved && nonce) {
```
emostov, line 63: I wonder if this is necessary, could we just do null if one of these fields doesn't exist?

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,6 +27,88 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+					reserved,
+					miscFrozen,
+					feeFrozen,
+					locks,
+				};
+			} else {
+				throw new BadRequest('Account not found');
+			}
+		} else if (
+			!Object.keys(historicalApi.query.system.account).includes('at')
```
emostov, line 78: I assume checking `!historicalApi.query.system.account.at` does not work? if it does i wonder if its slightly faster than pulling all the keys and then doing an O(n) includes

emostov (approved):
LGTM % a few questions where we _might_ be able to improve performance. Also, if these derives  https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/balances/account.ts start supporting `.at` I wonder if it would be more robust to start using them in parts of this service in the future

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: It actually does work, thanks for the suggestion. Way cleaner.

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: When you say 'do null' do you mean a `null` response for the key value pair? Or do you mean just check one value and then throw the error. 

Here I am just mimicking the error response logic that is in the existing service to keep it consistent with what response/errors it reports. So if we change this to something different it might be worth updating the rest as well.

TarikGul: Ahh, I see what you are saying, It only makes sense to really check for one value. Will update

TarikGul:
> Also, if these derives https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/balances/account.ts start supporting .at I wonder if it would be more robust to start using them in parts of this service in the future

Yea I totally agree with this.

src/services/accounts/AccountsBalanceInfoService.spec.ts
```diff
@@ -4,12 +4,43 @@
...
+/**
+ * Historical Queries MockApi
+ */
+const historicalMockApi = {
+	query: {
+		balances: {
+			freeBalance: () =>
+				Promise.resolve().then(() => {
+					return polkadotRegistry.createType('Balance', 123456789);
+				}),
```
maciejhirsz, line 25: Wouldn't that be easier?

```suggestion
			freeBalance: async () => polkadotRegistry.createType('Balance', 123456789),
```

maciejhirsz (approved):
Looks good to me.

There is a bunch of places where you mock an async function by chaining a `Promise.resolve` with no value, and I'm not sure it's necessary when a regular async function could be used?

src/services/accounts/AccountsBalanceInfoService.spec.ts
TarikGul: Nice catch, yea I definitely agree, will fix it up now.

TarikGul: fixed

PR #661, opened 2021-09-09
update @polkadot/apps-config

xlc:
Solves compatibility issue with Karura and likely for other networks as well.

PR #664, opened 2021-09-13
chore(release): 9.1.9

jsdw:


package.json
```diff
@@ -49,8 +49,8 @@
     "lint:scripts": "cd scripts && substrate-dev-run-lint"
   },
   "dependencies": {
-    "@polkadot/api": "^5.8.4-1",
-    "@polkadot/apps-config": "^0.95.1",
+    "@polkadot/api": "^5.9.1",
```
jsdw, line 52: The resolution is set to an older version still, but 5.9.1 is released now so the dependency updated. Is this OK or should this be bought down to match the resolution?

package.json
```diff
@@ -49,8 +49,8 @@
...
+    "@polkadot/apps-config": "^0.95.2-114",
```
jsdw, line 53: This is going from a stable release (a month ago) to one of the daily beta releases (see https://github.com/paritytech/substrate-api-sidecar/pull/661). I'm assuming that this is OK!

niklasad1 (commented):
Looks good but I'm not sure whether `polkadot/api`  should be ^5.8 or ^5.9...

package.json
dvdplm: Looks alright to me. I might be wrong but isn’t it pretty risk-free to bump the resolutions too?

dvdplm (approved):
Lgtm

package.json
jsdw: Basically, the resolutions are copied from https://github.com/polkadot-js/apps/blob/master/package.json, so I'm a little hesitant to deviate I guess. That said, everything seems to have decided on 5.9.1 regardless, so the resolution isn't really having an impact:

```
 % yarn why @polkadot/api
├─ @polkadot/api-derive@npm:5.9.1
│  └─ @polkadot/api@npm:5.9.1 (via npm:^5.8.4-4)
│
└─ @substrate/api-sidecar@workspace:.
   └─ @polkadot/api@npm:5.9.1 (via npm:^5.8.4-4)
```

I guess I'd be inclined to leave it as is I guess.

insipx (approved):

jsdw:
I'll merge and release this as-is, since it's passing the tests and such and the new apps-config version will hopefully resolve the issue that bought up #661

PR #677, opened 2021-09-19
fix: update polkadot-js deps

TarikGul:
This updates polkadot-js to 6.0.5 which has some major changes. 

Places where we were affected:

1) `AccountAssetsBalances`: We no longer receive the `AssetBalance` type but `PalletAssetsAssetBalance` instead which changes `isSufficient` -> `sufficient`. I made sure to account for runtimes that have either `isSufficient` or `sufficient` keys. MockApi also had to be adjusted.

2) Changed `BalanceOf` to `u128` in the `BlocksService` test suite

3) The `PalletsStorage` service uses V12 as (`asLatest`) in order to get the `{ modules }`, but now with V14 the metadata return `{ pallets }`, this fix is to make sure the current api doesnt break. But that being said the endpoint needs to be updated to deal with V14 as well. Tracking issue [here Issue #678](https://github.com/paritytech/substrate-api-sidecar/issues/678);


TarikGul:
Not Ready to merge as this breaks a lot of `getStorage` rpc calls. 

Example: querying `http://127.0.0.1:8080/blocks/1603025`

```
2021-09-19 11:11:16 error: RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: Unable to decode storage system.events:: createType(LastRuntimeUpgradeInfo):: {"specVersion":"Compact<u32>","specName":"Text"}:: Decoded input doesn't match input, received 0x1c0000000000000080e36a090000000002000000010000000000000000000000…0100000000000000000000000000000300000000008093dc1400000000000000 (209 bytes), created 0x1c00 (2 bytes)
2021-09-19 11:11:16 error: RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: Unable to decode storage session.validators:: createType(ValidatorId):: AccountId:: Decoded input doesn't match input, received 0x15030ea81d109d526eebecfc18c680281235a4bf23fade14e838d120a2943a48…9c6bae578812080b2f9ecdef913f4f68abc3dcee697f505625e8430c8ce3122c (6306 bytes), created 0x15030ea81d109d526eebecfc18c680281235a4bf23fade14e838d120a2943a48 (32 bytes)
```

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -117,7 +117,9 @@ export class AccountsAssetsService extends AbstractService {
 					assetId,
 					balance: assetBalance.balance,
 					isFrozen: assetBalance.isFrozen,
-					isSufficient: assetBalance.isSufficient,
+					isSufficient: assetBalance.sufficient
+						? assetBalance.sufficient
+						: (assetBalance['isSufficient'] as bool),
```
dvdplm, line 122:
```suggestion
					isSufficient: assetBalance.sufficient || (assetBalance['isSufficient'] as bool),
```

maybe?

src/services/blocks/BlocksService.spec.ts
```diff
@@ -143,7 +138,7 @@ describe('BlocksService', () => {
 
 		it('Return an error with a null calcFee when perByte is undefined', async () => {
 			mockApi.consts.transactionPayment.transactionByteFee =
-				undefined as unknown as BalanceOf & AugmentedConst<'promise'>;
+				undefined as unknown as u128 & AugmentedConst<'promise'>;
```
dvdplm, line 141: What does the `& AugmentedConst<'promise'>` part do?

dvdplm (approved):

Issue #486, opened 2021-03-23
add era info in extrinsic


jqriyv:
era info in extrinsic is a important message

emostov:
What is your use case for this info?

yimomo:
I want to look a extrinsic's left time. 
https://polkadot.subscan.io/extrinsic/4337373-2

PR #685, opened 2021-09-29
feat: add `era` to `blocks/{blockId}` response for extrinsics

TarikGul:
closes: [#486](https://github.com/paritytech/substrate-api-sidecar/issues/486)

This adds the `era` field.  @joepetrowski . 

The `era` enum has two possibilities, `immortalEra`, and `mortalEra`. Example return values for each:

```
era: {"immortalEra": "0x00"}
era: {"mortalEra": ["64", "11"]}
```  

Todo: 
- [X] Update e2e-tests
  - [X] kusama
  - [X] polkadot
  - [X] westend (Under the exception 2 tests are failing but they are unrelated to this PR)
- [X] Update unit tests


insipx (approved):

docs/src/openapi-v1.yaml
```diff
@@ -1793,6 +1795,25 @@ components:
...
           format: unsignedInteger
+    GenericExtrinsicEra:
+      type: object
+      description: |
+        The return value for eraInfo can either be `mortalEra`, or `immortalEra` and is represented as an enum in substrate. `immortalEra` meaning 
+        the transaction is valid forever. `mortalEra` consists of a tuple containing a period and phase.
+        ex: `"{"mortalEra": ["64", "11"]}"`. The Period is the period of validity from the block hash found in the signing material. 
+        The Phase is the period that this transaction's lifetime begins (and, importantly, 
+        implies which block hash is included in the signature material). If the `period` is 
+        greater than 1 << 12, then it will be a factor of the times greater than 1<<12 that
```
insipx, line 1806: This was copied from substrate docs, i think the wording is confusing but maybe OK?

TarikGul: Yea i actually totally agree, yea I just took them straight from the substrate docs. Maybe I leave out the technical part?

insipx: yeah tbh i think it would cause more confusion than it would help

TarikGul: Awesome I updated and removed the technical bit.

docs/src/openapi-v1.yaml
```diff
@@ -1793,6 +1795,23 @@ components:
...
+        The Phase is the period that this transaction's lifetime begins (and, importantly, 
+        implies which block hash is included in the signature material).
+      properties:
+        mortalEra: 
+          type: string
+          nullable: true
+        immortalEra:
+          type: string
+          nullable: true
+          default: ['0x00']
```
dvdplm, line 1813: If I understand this correctly, this is the only possible value for the `immortal` case? So rather than a default, it's just the hard coded value that must be there for Reasons™? If that is correct, can we get rid of it somehow, and have just `immortalEra`?

docs/src/openapi-v1.yaml
```diff
@@ -1793,6 +1795,23 @@ components:
...
+      example: "{\"immmortalEra\": ['0x00']}"
```
dvdplm, line 1814:
```suggestion
      example: "{\"immortalEra\": ['0x00']}"
```

dvdplm (approved):

docs/src/openapi-v1.yaml
```diff
@@ -1793,6 +1795,23 @@ components:
...
           type: string
           format: unsignedInteger
+    GenericExtrinsicEra:
+      type: object
+      description: |
+        The return value for eraInfo can either be `mortalEra`, or `immortalEra` and is represented as an enum in substrate. `immortalEra` meaning 
+        the transaction is valid forever. `mortalEra` consists of a tuple containing a period and phase.
+        ex: `"{"mortalEra": ["64", "11"]}"`. The Period is the period of validity from the block hash found in the signing material. 
+        The Phase is the period that this transaction's lifetime begins (and, importantly, 
+        implies which block hash is included in the signature material).
```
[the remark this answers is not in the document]
jsdw: This always confuses me in the documentation here too. I assumed a mortal era was just a block number, and a number of blocks it'd be valid from after that (and the signature material contains the block hash so we can verify that the block numbered is what you'd expect), but I don't think I'm quite right?!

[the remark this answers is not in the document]
jsdw: (to build on this; iirc we don't pass a block hash at all in the transaction; we just sign it against a block hash, so you need to know the block number you'll be getting the hash of to verify the signature I think?)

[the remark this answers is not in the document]
joepetrowski: Yes, the runtime keeps a mapping of `BlockNumber => BlockHash` for the last `BlockHashCount` blocks: https://github.com/paritytech/polkadot/blob/v0.9.10/runtime/common/src/lib.rs#L81

The mortality is a number of blocks, but it's encoded in one byte, so it's `2 ** mortality`. Front ends abstract this by letting you "choose" a validity period, but if you choose e.g. 50 blocks it would actually be mortal for 64.

docs/src/openapi-v1.yaml
jsdw: One way to represent the return types I guess could be:

```
era: {
   type: "immortal"
}
```

or

```
era: {
   type: "mortal",
   period: 64,
   phase: 12
}
```

(with a typescript type like `{ type: "immortal" } | { type: "mortal", period: number, phase: number }`)

dvdplm: > A description explaining that the value is hardcoded or will always be '0x00' would possibly suffice?

That's ok for this PR. Perhaps add an issue to investigate James' suggestion?

TarikGul: Sweet, so I added a concise description for `immortal` -> `Hardcoded constant '0x00'`. I do like the idea of the return type being as @jsdw put it. The one reason I like it is the natural response from substrate has `mortal` being iterable, and `immortal` just being a string. Having different data structures in the same response field is a little odd for our circumstances. 

I do think for now it might be wise to keep the response as close to substrate as possible. The idea being if anything were to change in substrate (low likely hood), the little logic there is for getting `era` wont change.

TarikGul:
Merging

PR #687, opened 2021-09-29
fix: refactor e2e chain tests

TarikGul:
rel: [#684](https://github.com/paritytech/substrate-api-sidecar/pull/684)

This is a follow up to the PR mentioned above. It refactors the `runChainTests.ts` to use the scriptApi, and also lints and cleans up code in the scripts directory.

### DIFF
--- a/scripts/runChainTests.ts
+++ b/scripts/runChainTests.ts
@@ -1,68 +1,11 @@
 import { ArgumentParser, Namespace } from 'argparse';
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 
 import { config, defaultSasBuildOpts } from './config';
-import { IProcOpts, StatusCode } from './types';
+import { StatusCode, ProcsType } from './types';
+import { setWsUrl, launchProcess, killAll } from './sidecarScriptApi';
 
 // Stores all the processes
-const procs: { [key: string]: ChildProcessWithoutNullStreams } = {};
-
-// Set the env variable for SAS_SUBSTRATE_WS_URL
-const setWsUrl = (url: string): void => {
-	process.env.SAS_SUBSTRATE_WS_URL = url;
-};
-
-/**
- * Launch any given process. It accepts an options object.
- *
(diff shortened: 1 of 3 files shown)


jsdw (approved):
I think this LGTM offhand!

(I'm not a huge fan of the global `procs` that is passed to `launchProcess` (I'd rather `launchProcess` returneda  thing that you could call `kill` on) but that's just personal preference really 🙂)

TarikGul:
> (I'm not a huge fan of the global procs that is passed to launchProcess (I'd rather launchProcess returned a thing that you could call kill on) but that's just personal preference really 🙂)

@jsdw I actually agree with this sentiment. I put a lot of thought into it yesterday, and my initial thought was its okay to have the procs initialized as a global variable inside of the script on the notion that the scripts should always be isolated from eachother, and never intertwined or ran in conjunction.

 My other idea was to create a `ScriptApi` class which is then instantiated on initiailization. This could populate a `procs` object specific to that class, and have getters and setters.  <<-- If this is something you would prefer and think is most obvious and better for longevity I am happy to set it up in this PR to avoid future technical debt. The idea being that if the script api ever grows it would probably be a no brainer.

maciejhirsz (approved):

jsdw:
I guess my thinking was that if the `launchProcess` method returned something with a shape like`{ kill: () => void, ... }`, the individual scripts using it could, if they decided, stick that in a global object if they wanted, orr they might not want/need to keep track of a bunch of stuff in a global and have the choice to just call the `kill` method directly as needed. Having `launchProcess` take in the object to populate sortof pushes scripts down a more "global first" route :)

TarikGul:
@jsdw Filed issue [#689](https://github.com/paritytech/substrate-api-sidecar/issues/689). I actually think your design around `launchProcess` makes a lot of sense, and would improve the API's flexibility with scripts. Since it causes for some heavy refactoring I'll have it be a follow up PR. 

Merging.

PR #688, opened 2021-09-30
fix: add block weights for Shiden network to support fee calculation

hoonsubin:
This PR adds a controller and metadata consts for Shiden network


TarikGul:
Hey @hoonsubin , thanks for the PR. I activated the tests for you. Let me know when this is ready for review and I will run it locally as well.

hoonsubin:
Thank you @TarikGul 
I'm currently having an issue with fixing the fee calculation for Shiden network.
Screen Shot 2021-09-30 at 4 39 41 PM

I tried specifying the block weights and spec versions, but `CalcFee.from_params` (https://github.com/hoonsubin/substrate-api-sidecar/blob/9f7a29f7bffd4ce225224234865d9c78d2b7f941/src/services/blocks/BlocksService.ts#L532) seems to always return `undefined`. Do you know what I'm doing wrong here?

I'm running a local dev instance with the following environment variables.

```env
# file name `.env.shiden`
SAS_SUBSTRATE_WS_URL=wss://rpc.shiden.astar.network
SAS_EXPRESS_LOG_MODE=all
SAS_LOG_LEVEL=http
```

TarikGul:
Ill have a look.

hoonsubin:
@TarikGul Sorry for the late response and thank you for the suggestion. I added the shiden network's sepc multiplier, please let me know if there is any changes I need to make!

TarikGul:
> @TarikGul Sorry for the late response and thank you for the suggestion. I added the shiden network's sepc multiplier, please let me know if there is any changes I need to make!

No worries at all, I'll have a run at this locally to confirm it works.

.env.local
```diff
@@ -3,4 +3,4 @@
 # For more information on how to use .env files and environment variables
 # consult the Configuration section in the README.
 
-SAS_SUBSTRATE_WS_URL=ws://127.0.0.1:9944
+SAS_SUBSTRATE_WS_URL=ws://127.0.0.1:9944
```
TarikGul, line 6: Could you add a new line at the end of this file :)

TarikGul:
Okay sweet, its working on my end. Thanks again for the PR. If you could also run sh ./calc/build.sh from the root directory to update the target wasm pkg.

hoonsubin:
> Okay sweet, its working on my end. Thanks again for the PR. If you could also run sh ./calc/build.sh from the root directory to update the target wasm pkg.

It seems like this was the result of the build https://github.com/paritytech/substrate-api-sidecar/pull/688/files#diff-a252d41fbf4a811e25b7ff2afecb973ca578efa0756c6d7ca14f6fab09bd3f62. Running `yarn build:calc` results with the same diff with Git.

TarikGul (approved):
LGTM, ran it locally against sidecar and its calculating fees correctly..

insipx (approved):
LGTM!

PR #691, opened 2021-10-02
fix(deps): bump polkadot/api, cleanup resolutions

TarikGul:
This PR cleans up the resolutions and bumps `@polkadot/api` to v6.1.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
     - @polkadot/util-crypto [release notes](https://github.com/polkadot-js/common/releases)
     - @substrate/calc [npm release page](https://www.npmjs.com/package/@substrate/calc)
 
-1. Next make sure the resolutions are up to date inside of the `package.json` to match polkadot-js [here](https://github.com/polkadot-js/apps/blob/master/package.json). This is a manual process, and the packages we want to update are only the ones with `@polkadot` appended to it. If any issues may surface, contact the maintainers. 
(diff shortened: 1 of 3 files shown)


insipx (approved):

README.md
```diff
@@ -295,7 +295,7 @@ All the commits in this repo follow the [Conventional Commits spec](https://www.
     - @polkadot/util-crypto [release notes](https://github.com/polkadot-js/common/releases)
     - @substrate/calc [npm release page](https://www.npmjs.com/package/@substrate/calc)
 
-1. Next make sure the resolutions are up to date inside of the `package.json` to match polkadot-js [here](https://github.com/polkadot-js/apps/blob/master/package.json). This is a manual process, and the packages we want to update are only the ones with `@polkadot` appended to it. If any issues may surface, contact the maintainers. 
+1. Next make sure the resolutions are up to date inside of the `package.json` for all `@polkadot/*` packages, please refer to the releases of each polkadot package we update as a dependency, and reach out to the maintainers for any questions.
```
niklasad1, line 298: hmm, that sounds like polkadot-js changelog/release log is not detailed enough but fine...

niklasad1 (approved):

README.md
TarikGul: Well actually its that, more often than not were are a little out of line with polkadot-js/apps, because its the last thing in the chain of releases. Sometimes it off for what we need, and causes confusion, or may even cause a bug if our resolutions don't match somewhere.

Issue #690, opened 2021-10-02
[BUG]E2E tests fail for westend on blocks 5495855, and 5493461


TarikGul:
Currently there is a bug for '/blocks/{blockId}' when connected to westend:

Blocks:
5495855,
5493461

Error:
```json
{
  "code": 500,
  "message": "createType(Lookup36):: Unable to create Enum via index 112, in InvalidEquivocationProof, InvalidKeyOwnershipProof, DuplicateOffenceReport",
  "stack": "Error: createType(Lookup36):: Unable to create Enum via index 112, in InvalidEquivocationProof, InvalidKeyOwnershipProof, DuplicateOffenceReport\n    at createTypeUnsafe (/Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/create/createType.cjs:68:18)\n    at TypeRegistry.createTypeUnsafe (/Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/create/registry.cjs:364:45)\n    at /Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/metadata/decorate/constants/index.cjs:28:30\n    at Array.reduce (<anonymous>)\n    at /Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/metadata/decorate/constants/index.cjs:27:58\n    at Array.reduce (<anonymous>)\n    at decorateConstants (/Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/metadata/decorate/constants/index.cjs:18:18)\n    at expandMetadata (/Users/tarik/Desktop/parity/substrate-api-sidecar/node_modules/@polkadot/types/metadata/decorate/index.cjs:63:42)\n    at BlocksService.getWeight (/Users/tarik/Desktop/parity/substrate-api-sidecar/build/src/services/blocks/BlocksService.js:385:68)\n    at processTicksAndRejections (internal/process/task_queues.js:93:5)",
  "level": "error"
}
```

In order to reproduce the bug in polkadot.js:

```javascript
const main = async () => {
    const api = await ApiPromise.create({
        provider: new WsProvider('wss://westend-rpc.polkadot.io'),
    });

    const hash = await api.rpc.chain.getBlockHash(5493461)
    const metadata = await api.rpc.state.getMetadata(hash);

    const decorate = expandMetadata(api.registry, metadata);
[...]

PR #692, opened 2021-10-02
fix: expandMetadata for historic runtimes

TarikGul:
closes: [#690](https://github.com/paritytech/substrate-api-sidecar/issues/690)

This fixes a bug where when expanding metadata in order to capture the weights used in calculating we dont use the correct registry per historic block. This changes that to have a historicApi return the correct registry for the block.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -558,7 +558,7 @@ export class BlocksService extends AbstractService {
 		blockHash: BlockHash
 	): Promise<WeightValue> {
 		const metadata = await api.rpc.state.getMetadata(blockHash);
-		const historicApi = await api.at(blockHash);
+		const historicApi = await api.at(blockHash) as ApiPromise;
 		const {
 			consts: { system },
 		} = expandMetadata(historicApi.registry, metadata);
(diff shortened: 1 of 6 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -558,9 +558,10 @@ export class BlocksService extends AbstractService {
 		blockHash: BlockHash
 	): Promise<WeightValue> {
 		const metadata = await api.rpc.state.getMetadata(blockHash);
+		const historicApi = (await api.at(blockHash)) as ApiPromise;
```
niklasad1, line 561: @TarikGul I suppose the registry contains metadata, chain info and other stuff that is required for that data to match for historical blocks?

niklasad1 (approved):
LGTM

src/services/blocks/BlocksService.ts
TarikGul: Yup exactly! 

Example: The specific error that brought up this issue too was a type error where Westend has a `Lookup36` type in its Metadata, but it didnt match the registry. So when trying to expand the metadata, types didn't align.

insipx (approved):

Issue #698, opened 2021-10-07
Port over all api.query.<pallet>.<method>.at() -> api.at() historic api


TarikGul:
All services that use api.query.<pallet>.<method>.at() will be deprecated via [affa6ab](https://github.com/polkadot-js/api/commit/affa6ab1bd19b7819766fd8015f3c72ecf975d9c) in polkadot-js. It's important to refactor each service to now use a `api.at()` historicApi. 

List of services to refactor:

- [X] BlocksService [#699]
- [x] AccountAssetsBalanceInfoService [#721]
- [X] AccountsBalanceInfoService [#709]
- [x] AccountsStakingInfoService [#741]
- [x] AccountsStakingPayoutsService [#762]
- [X] AccountsVestingInfoService [#709]
- [x] PalletsStakingProgressService [#761]
- [X] PalletsStorageService [#710]
- [X] ParasService [#735]
- [X] [ readjust mockApi, and historicApi for testing env [#702]

rel: [#697](https://github.com/paritytech/substrate-api-sidecar/issues/697)

TarikGul:
Before I can go any further with this issue, [#702](https://github.com/paritytech/substrate-api-sidecar/pull/702) needs to be merged

jacogr:
Just a note - these won't be removed anytime soon (certainly not in the mids of the 6.x range). The reality is that `api.at` is cleaner and does a whole host of things better, i.e. giving you an actual instance at that point, correctly decorated. Additionally this is used everywhere, so it certainly won't be dropped soon (probably will still live in the 7.x version as well), generally don't like breaking other people's code.

Currently the older versions now use `api.at` internally so it should yield less breakages, which with the new metadata + types is quite tricky with the (now deprecated) versions.

TarikGul:
Okay sweet, good to know. I Appreciate the explanation. That helps a lot because I currently see some instances where I might have to leave the `api.query.*.*.at` in for the time being.  I have actually been finding `api.at` to be a really great upgrade in general as well, having the centralized source of truth in the `api` is such an incredibly useful feature.

Issue #693, opened 2021-10-04
Benchmarks showing regression from v9.1.9 -> v9.1.10


TarikGul:
## Commit hash `fcf60eec106d0ac267ed3c9cf7d86dbc0009a86c ` Pre-v9.1.10
```
Running 2m test @ http://127.0.0.1:8080
  4 threads and 12 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.41s     3.08s   14.90s    66.43%
    Req/Sec     1.99      4.11    20.00     87.80%
  Latency Distribution
     50%    4.89s 
     75%    7.76s 
     90%    9.29s 
     99%   14.12s 
  268 requests in 2.00m, 296.39MB read
Requests/sec:      2.23
Transfer/sec:      2.47MB
Total completed requests:       268
Failed requests:        0
Timeouts:       0
Average latency:        5408.6096s
```

## Commit Hash `f14f2c2ed9da295d31dfe63c7d89d4f8613ad0db` Pre-v9.1.10
```
Running 2m test @ http://127.0.0.1:8080
  4 threads and 12 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    21.14s    17.07s   46.86s    45.76%
    Req/Sec     1.21      2.30    10.00     89.58%
  Latency Distribution
     50%   25.25s 
     75%   29.13s 
     90%   46.76s 
     99%   46.86s 
  59 requests in 2.00m, 74.90MB read
Requests/sec:      0.49
Transfer/sec:    638.79KB
Total completed requests:       59
Failed requests:        0
Timeouts:       0
Average latency:        21138.197677966s
```

The issue lies around commit [#677](https://github.com/paritytech/substrate-api-sidecar/pull/677). Whether or not this is an implementation issue or an regression that stems from polkadot-js is still yet to be known. But we do know the regression started when we bumped the deps from 5.x.x -> 6.x.x

TarikGul:
rel: [#4019](https://github.com/polkadot-js/api/issues/4019)

This solves a residual issue found in polkadot-js where the cache was getting reset each call. This actually bumps up our performance a bit, but doesnt solve the main underlying issue that arise's after v5.9.1.

TarikGul:
Large part of the regression is now resolved. Will keep the issue open while we transition to v14

TarikGul:
[...]

PR #699, opened 2021-10-07
fix(BlocksService): refactor api.derive for performance, and add historicApi to BlocksService

TarikGul:
rel: [#693](https://github.com/paritytech/substrate-api-sidecar/issues/693)

### Performance bump

Benchmarks were done against `/blocks` where the main known regression took place. There are 3 versions we are testing here: v9.1.9, v9.2.0, and this current branch which I will denote as 'beta'.

```
// This benchmark reflects sidecar with an empty cache in all respects
4 threads 12 connections 2 mins
v9.1.9 -> Completed Reqs: 162, Avg Latency: 8.61s
v9.2.0 -> Completed Reqs: 24, Avg Latency: 45.08s
beta -> Completed Reqs: 55, Avg Latency: 22.75s

// The below benchmarks reflects sidecar with a filled cache
4 threads 6 connections 2 mins
v9.1.9 -> Completed Reqs: 239, Avg Latency: 1.98s
v9.2.0 -> Completed Reqs: 71, Avg Latency: 7.86s
beta -> Completed Reqs: 147, Avg Latency: 3.86s

4 threads 4 connections 2 mins
v9.1.9 -> Completed Reqs: 255, Avg Latency: 1.88s
v9.2.0 -> Completed Reqs: 107, Avg Latency: 5.94s
beta -> Completed Reqs: 211, Avg Latency: 2.28s
```

### `Api.at()`

Local Benchmarks were also ran against instantiating `api.at()`. Initially for a particular runtime if the information isn't cached inside of the api it will take about 3 seconds too initialize. But after its cached it usually takes 1-50 milliseconds to recall. 

All instances of `api.query....at()` in `BlocksService` are now using a historicApi to get any information, regarding `api.query`. During this I found a bug where we were getting the wrong `transactionByteFee` for historic runtimes because the api was always using the latest metadata. This is also now fixed, and all e2e tests for kusama reflect this change. 

### `api.derive`

I noticed `api.derive` was a large area of regression for sidecar while doing local isolated benchmarks, therefore I took the method apart and abstracted the things we need from it in order to speed the request time up. 

### extra notes

I fixed the mockApi for tests in `PalletsStakingProgressService.spec.ts`. Moving forward until all services are switched to using the `historicApi` the mockApi enviornment will be going through some adjustments. Once the final Service is fixed up, the mockApi will be readjusted to have a generated historicApi, as well as mockApi. 

Part 2: [#702](https://github.com/paritytech/substrate-api-sidecar/pull/702)

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -509,9 +509,8 @@ export class BlocksService extends AbstractService {
 		 * This will remain using the original api.query.*.*.at to retrieve the multiplier
 		 * of the `parentHash` block.
 		 */
-		const multiplier = await api.query.transactionPayment?.nextFeeMultiplier?.at(
-			parentHash
-		);
+		const multiplier =
+			await api.query.transactionPayment?.nextFeeMultiplier?.at(parentHash);
 
 		const perByte = historicApi.consts.transactionPayment?.transactionByteFee;
 		const extrinsicBaseWeightExists =


src/services/blocks/BlocksService.ts
```diff
@@ -496,13 +507,13 @@ export class BlocksService extends AbstractService {
 		}
 
 		const multiplier =
-			await api.query.transactionPayment?.nextFeeMultiplier?.at(parentHash);
+			await historicApi.query.transactionPayment?.nextFeeMultiplier();
```
emostov, line 510: From what I understand we may actually need `(await api.at(parentHash)).query.transactionPayment?.nextFeeMultiplier()`

src/services/blocks/BlocksService.ts
```diff
@@ -98,22 +101,24 @@ export class BlocksService extends AbstractService {
 			return isBlockCached;
 		}
 
-		const [deriveBlock, events, finalizedHead] = await Promise.all([
-			api.derive.chain.getBlock(hash),
-			this.fetchEvents(api, hash),
+		const [{ block }, validators, events, finalizedHead] = await Promise.all([
+			api.rpc.chain.getBlock(hash),
```
emostov, line 105: Is there a reason we don't completely get of `api` and instead use `historicApi` everywhere?

Like at the top of all these services we could replace `const { api } = this;` with `const  api = this.api.at(has)`;

TarikGul: So we still need to use the api for rpc methods, which is why we would keep the `this.api`. Whats good is `this.api` will cache the registries for each runtime it creates a historicApi for, so it gets pretty fast.

src/services/blocks/BlocksService.ts
```diff
@@ -555,10 +567,10 @@ export class BlocksService extends AbstractService {
...
 	private async getWeight(
 		api: ApiPromise,
+		historicApi: ApiDecoration<'promise'>,
 		blockHash: BlockHash
 	): Promise<WeightValue> {
 		const metadata = await api.rpc.state.getMetadata(blockHash);
-		const historicApi = (await api.at(blockHash)) as ApiPromise;
 		const {
 			consts: { system },
 		} = expandMetadata(historicApi.registry, metadata);
```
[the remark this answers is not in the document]
TarikGul: Good point, ill check it out.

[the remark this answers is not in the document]
TarikGul: Yup removed expandMetadata and optimized getWeights!

e2e-tests/endpoints/kusama/blocks/2786879.json
```diff
@@ -148,7 +148,7 @@
             "info": {
                 "weight": "195000000",
                 "class": "Normal",
-                "partialFee": "3333333"
+                "partialFee": "166666666"
```
niklasad1, line 151: why is this partialFee changed?

I guess because the historialApi is used or the calculate fee bug....

TarikGul: Yuo so I found a bug for calculating fees in regards to historic blocks. I fixed it in this PR thats why I changed the e2e tests too for partialFee responses.

PR #717, opened 2021-10-14
fix!: change the response for vesting-info

TarikGul:
## BREAKING CHANGE

With runtime v9110, the vesting-info response now returns an array for `vesting`, with the `VestingInfo` type now being `Vec<PalletVestingVestingInfo>`. Sidecar previously returned responses under the `vesting` key as objects, therefore when there was no information the response was an empty object. 

This is now changed as the response will default to an empty array when there is no vesting-info on the account, and the past responses that were objects, will now be wrapped in an array. 

### Examples:

Before: 
```
{
  "at": {
    "hash": "0x89f4608b568a05a89737bc826a32f8677f2646a5773312d7dee3a667458e5a98",
    "height": "6000000"
  },
  "vesting": {
      "locked": "1749990000000000",
      "perBlock": "166475460",
      "startingBlock": "4961000"
    }
}
```

After
```
{
  "at": {
    "hash": "0x89f4608b568a05a89737bc826a32f8677f2646a5773312d7dee3a667458e5a98",
    "height": "6000000"
  },
  "vesting": [
    {
      "locked": "1749990000000000",
      "perBlock": "166475460",
      "startingBlock": "4961000"
    }
  ]
}
```

### DIFF
--- a/src/services/accounts/AccountsVestingInfoService.ts
+++ b/src/services/accounts/AccountsVestingInfoService.ts
@@ -26,13 +26,18 @@ export class AccountsVestingInfoService extends AbstractService {
 			height: number.unwrap().toString(10),
 		};
 
-		return {
-			at,
-			vesting: vesting.isNone
-				? []
-				: Array.isArray(vesting.unwrap())
-				? vesting.unwrap()
-				: [vesting.unwrap()],
-		};
+		if (vesting.isNone) {
+			return {
+				at,
+				vesting: [],
+			};
+		} else {
+			const unwrapVesting = vesting.unwrap();
+
+			return {
+				at,
+				vesting: Array.isArray(unwrapVesting) ? unwrapVesting : [unwrapVesting],
+			};
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


src/services/accounts/AccountsVestingInfoService.ts
```diff
@@ -28,7 +28,11 @@ export class AccountsVestingInfoService extends AbstractService {
 
 		return {
 			at,
-			vesting: vesting.isNone ? {} : vesting.unwrap(),
+			vesting: vesting.isNone
+				? []
+				: Array.isArray(vesting.unwrap())
+				? vesting.unwrap()
+				: [vesting.unwrap()],
```
maciejhirsz, line 35: Would like to avoid double unwrap here, but not sure what a good way would be, TS would likely scream about something like:

```suggestion
				: Array.isArray(vesting = vesting.unwrap())
				? vesting
				: [vesting],
```

Could use a different variable for the unwrapped value (declared before the return, assigned in place)? :thinking:

TarikGul: Ahh very good point. I should have noticed that earlier when I tried to abstract the `vesting.unwrap()` before calling vesting.isNone which caused me issues. 

Good call. Going to declare 2 variables, one for the bool value, and the other to unwrap vesting

Issue #719, opened 2021-10-15
Timeouts getting blocks with Sidecar v10.0.0, node v0.9.11


pablojorge:
We've been experiencing serious timeouts fetching blocks on 9.2.0 and 10.0.0 after runtime v9100/v9110 went live.

Example:

```
"0262c8","2021-10-15T01:10:52Z","2021-10-15 01:10:52 error: GET /blocks/7262958 500 232588ms"
"0262c8","2021-10-15T01:10:52Z","2021-10-15 01:10:52 error: Failed while trying to get the latest header. "
"0262c8","2021-10-15T01:10:52Z"," InternalServerError: Failed while trying to get the latest header."
"0262c8","2021-10-15T01:10:52Z","    at /substrate-sidecar/build/src/controllers/AbstractController.js:104:23"
"0262c8","2021-10-15T01:10:52Z","    at runMicrotasks (<anonymous>)"
"0262c8","2021-10-15T01:10:52Z","    at processTicksAndRejections (node:internal/process/task_queues:96:5)"
"0262c8","2021-10-15T01:10:52Z","    at async BlocksController.getHashForBlock (/substrate-sidecar/build/src/controllers/AbstractController.js:103:32)"
"0262c8","2021-10-15T01:10:52Z","    at async BlocksController.getBlockById (/substrate-sidecar/build/src/controllers/blocks/BlocksController.js:118:26)"
"0262c8","2021-10-15T01:10:52Z","    at async /substrate-sidecar/build/src/controllers/AbstractController.js:168:9"
```

On the node side there are no errors, everything seems fine.

After manually restarting the services (node and sidecar) it goes back to normal, but some hours later we start getting timeouts again.

pablojorge:
This is happening with Polkadot mainnet (the only chain we track), and we run our own internal nodes. In fact the nodes seem to be perfectly fine, and never stop receiving and processing blocks.

Memory patterns are not different from previous versions, so nothing really stands out. I'll collect more logs

pablojorge:
Going back in the logs It seems the issue is that Sidecar for some reason disconnects from the node (we're connecting over websocket), and the timeouts are ridiculously high, so it takes a _lot_ of time to react and reconnect.

Full logs: (using [...] to reduce duplication)
```
[...]

Issue #418, opened 2021-02-03
sidecar does not reconnect if substrate node restarts


jadechip:
**Proposed Change or Idea**

The substrate-api-sidecar service does not attempt to reconnect if the associated substrate node goes down or restarts.
This is particularly troublesome in a Kubernetes environment where deployments often trigger pods to restart.

**Alternatives**

The substrate-api-sidecar service could periodically attempt to reconnect to the substrate node.

**Additional Information**

[Add any other details about the feature request here.]

emostov:
Possibly relates to #326

TarikGul:
Closing in favor of this tracking issue [#719](https://github.com/paritytech/substrate-api-sidecar/issues/719).

Polkadot-js has a few built it ws-options, and default behavior for reconnecting. One way to confirm this locally, is have sidecar connected to a public node, then turn off the wifi, and turn it back on.

Issue #326, opened 2020-10-22
Investigate strategies to maintain service through lossy websocket connection


emostov:
Relates to #317 

Strategies should be investigated that intend to make sure sidecar can keep running when websocket connections disconnect

This may be a modification at the `ApiPromise` level, or could even involve sidecar restarting itself.

dvdplm:
I think doing this is a good idea. The tricky part might be knowing (reliably) when a websocket is disconnects in such a way that a reconnect makes sense, but it can be a QoL improvement for 80% of users.
I suggest we make it a config option that is off by default and with a default value set fairly high, 30sec or so, to avoid spam. If it pans out and users are happy it can become an on-by-default value in a subsequent breaking release.

anhcao142:
Hi, sorry to bring this up again, but why don't you consider using the HTTP/HTTPS method?

TarikGul:
@anhcao142 Do you mind going a little more in detail. 

Under the hood the connection is being established by `@polkadot/api` which leverages websocket connection. I am not sure I understand the use case for HTTP/HTTPs here but I might be missing something.

anhcao142:
Hi @TarikGul, the `@polkadot/rpc-provider` provides both `WsProvider` and `HttpProvider` to respectively connect to the node through WebSocket and RPC. So it's possible for `substrate-api-sidecar` to connect to the node through WebSocket and HTTP too.

anhcao142:
To demonstrate, here is my modified version of `substrate-api-sidecar` in order to connect to the node through the RPC port

```
import { WsProvider, HttpProvider } from '@polkadot/rpc-provider';

async function main() {
	const { config } = SidecarConfig;
	const { logger } = Log;
	// Overide console.{log, error, warn, etc}
	consoleOverride(logger);
	logger.info(`Version: ${packageJSON.version}`);
	const { TYPES_BUNDLE, TYPES_SPEC, TYPES_CHAIN, TYPES } = config.SUBSTRATE;
	// Instantiate a web socket connection to the node and load types
	const api = await ApiPromise.create({
		provider: config.SUBSTRATE.URL.startsWith('http')
			? new HttpProvider(config.SUBSTRATE.URL) // <-- I connect to the node using RPC here
			: new WsProvider(config.SUBSTRATE.URL),
		/* eslint-disable @typescript-eslint/no-var-requires */
		typesBundle: TYPES_BUNDLE
[...]

PR #722, opened 2021-10-18
fix(middleware): create `reconnectMiddleware` to attempt to reconnect to the api when down

TarikGul:
closes: [#418](https://github.com/paritytech/substrate-api-sidecar/issues/418)
closes: [#326](https://github.com/paritytech/substrate-api-sidecar/issues/326)
rel: [#719](https://github.com/paritytech/substrate-api-sidecar/issues/719)

This creates a middleware that will reconnect the API-WS, and block all requests until it is reconnected. This:

1) Make sure api is disconnected manually (i.e: api.disconnect)
2) Attempt to reconnect recursively, max at 30 reconnect attempts (Subject to change after discussion)
3) Stop requests from being processed when the websocket connection is down
4) Once reconnected allow requests to go through. 

I should implement an incoming requests queue count. This keeps track of the number of requests that are blocked, and allows each to resolve in order with a time buffer. This way all requests wont go through at the same time when the API-WS is reconnected. 

I also need to finish writing tests for this. Currently works in a live environments, but it needs unit tests.


TarikGul:
As an update as well. This PR wont be merged until able to reproduce it locally.

TarikGul:
Going to close this PR for now to avoid clutter. 

Some takeaways:

1. Middleware like this complicates Sidecar in an unnecessary way, when there is already an underlying built in reconnection logic in polkadot-js. 
2. If something like this were to go in, it would need to be off by default. Some type of ENV config to turn it on.
3. We are not able to reproduce this against polkadot, kusama, or westend so it's hard to justify its use.

PR #724, opened 2021-10-19
Address verification enhancement

toolsopen:


src/main.ts
```diff
@@ -56,7 +57,9 @@ async function main() {
 		types: TYPES ? (require(TYPES) as RegistryTypes) : undefined,
 		/* eslint-enable @typescript-eslint/no-var-requires */
 	});
-
+	if(api.registry.chainSS58 != undefined){
```
TarikGul, line 60: So this logic would actually have to be contained inside of the middleware. It would have to check if the requests are grabbing information from a past block that has a historic runtime OR if there is no `at` query parameter attached then to just use the current api's registry. This also stops us from mutating any data in polkadot-js through `{ defaults }`, and potentially having other files import that mutated dep.

emostov (changes requested):
I don't think we should do address validation based on the prefix. If we want that level of checks we could add an endpoint that tells some info about an address, like what network its associated with and generally if its valid

PR #735, opened 2021-10-27
fix(ParasService): adjust endpoint to use historicApi, fix endingOffset bug

TarikGul:

This PR refactors the the ParasService to use a historicApi for querying the chain. 

### Bug Fix

There was a bug for `/experimental/paras/auctions/current?at=8400000` where the incorrect `endingOffset` was being calculated and passed into `auctions.winning` which would return the following:

```
RPC-CORE: getStorage(key: StorageKey, at?: BlockHash): StorageData:: Unable to decode storage auctions.winning:: createType(WinningData):: [Option<(AccountId,ParaId,BalanceOf)>;10]:: Decoded input doesn't match input, received 0x00000000000000016d6f646c70792f6366756e64d10700000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (88 bytes), created 0x00000000000000016d6f646c70792f6366756e64d107000000000000000000000000000000000000d10700002550b60924ae410100000000000000000000 (62 bytes)
Error: Unable to decode storage auctions.winning:: createType(WinningData):: [Option<(AccountId,ParaId,BalanceOf)>;10]:: Decoded input doesn't match input, received 0x00000000000000016d6f646c70792f6366756e64d10700000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (88 bytes), created 0x00000000000000016d6f646c70792f6366756e64d107000000000000000000000000000000000000d10700002550b60924ae410100000000000000000000 (62 bytes)
```

The endingOffset for block 8400000 should be 1478008, but we were receiving 412. 

I used the implementation inside of `@polkadot/apps` in order to fix the current bug. 

It also takes into account an edgecase where if the current now block substracted by the offset is less than zero than the leasePeriodIndex will be `null`.


src/services/paras/ParasService.ts
```diff
@@ -370,11 +392,14 @@ export class ParasService extends AbstractService {
...
+	 * @param now Current blockHeight
 	 */
-	private leasePeriodIndexAt(now: BN): BN {
-		const leasePeriod = this.api.consts.slots.leasePeriod as BlockNumber;
+	private leasePeriodIndexAt(
+		historicApi: ApiDecoration<'promise'>,
+		now: BN
+	): BN {
+		const leasePeriod = historicApi.consts.slots.leasePeriod as BlockNumber;
 		return now.div(leasePeriod);
```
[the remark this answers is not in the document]
TarikGul: @emostov updated if. Could you take a look when you get a chance

jsdw (approved):
Looks sane; I'm happy if others are! :)

src/services/paras/ParasService.ts
```diff
@@ -369,13 +391,20 @@ export class ParasService extends AbstractService {
...
-		return now.div(leasePeriod);
+	private leasePeriodIndexAt(
+		historicApi: ApiDecoration<'promise'>,
+		now: BN
+	): BN {
+		const leasePeriod = historicApi.consts.slots.leasePeriod as BlockNumber;
+		const offset =
+			(historicApi.consts.slots.leaseOffset as BlockNumber) || BN_ZERO;
+
+		return now.sub(offset).div(leasePeriod);
```
emostov, line 407: This needs to be updated slightly. There is an edge case where offset > now and we decided to consider this has have _no_ lease period index. In the substrate code this is represented by None. Here is the line https://github.com/paritytech/polkadot/pull/3980/commits/3668966dc02ac793c799d8c8667e8c396d891734#diff-3e19d8a242311f202d80bcf9998ffdc22ddd933ed340f308276f353f137e2304R444

So basically it says "if now - offset is less than 0, return None". So we need a update for the logic that uses this function to basically say there is no lease period at `now` and we could add the first lease period start at block `offset`. 

 (This logic which got added to the substrate PR a bit late).

emostov: (You can see the changes for this in substrate commit 3668966dc02ac793c799d8c8667e8c396d891734)

src/services/paras/ParasService.ts
```diff
@@ -151,11 +151,13 @@ export class ParasService extends AbstractService {
...
 				blockNumber
-			).toNumber();
+			);
 
 			leasesFormatted = leases.reduce((acc, curLeaseOpt, idx) => {
 				if (curLeaseOpt.isSome) {
-					const leasePeriodIndex = currentLeasePeriodIndex + idx;
+					const leasePeriodIndex = currentLeasePeriodIndex
+						? currentLeasePeriodIndex.toNumber() + idx
+						: 0;
```
TarikGul, line 160: Not sure what this should be returning.

src/services/paras/ParasService.ts
```diff
@@ -330,14 +332,16 @@ export class ParasService extends AbstractService {
...
+			? leasePeriodIndex.mul(leasePeriod).add(leasePeriod)
+			: BN_ZERO;
 
 		return {
 			at: {
 				hash,
 				height: blockNumber.toString(10),
 			},
-			leasePeriodIndex,
+			leasePeriodIndex: leasePeriodIndex ? leasePeriodIndex : BN_ZERO,
```
emostov, line 344: IMO these values should be null, not 0

src/services/paras/ParasService.ts
emostov: I think it should be null

emostov (approved):

PR #740, opened 2021-10-28
fix: update blockWeights store for westend and kusama

TarikGul:
Updates the blockWeights Store for both westend and kusama with the most up to date runtimes.

### DIFF
--- a/src/chains-config/metadata-consts/kusamaConsts.ts
+++ b/src/chains-config/metadata-consts/kusamaConsts.ts
@@ -12,7 +12,7 @@ export const kusamaDefinitions: MetadataConsts[] = [
 	{
 		runtimeVersions: [
 			2027, 2028, 2029, 2030, 9000, 9010, 9030, 9040, 9050, 9070, 9080, 9090,
-			9100, 9110, 9111,
+			9100, 9110, 9111, 9122,
 		],
 		perClass,
 	},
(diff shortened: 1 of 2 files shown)


niklasad1 (commented):
why isn't polkadot bumped as well?

src/chains-config/metadata-consts/kusamaConsts.ts
```diff
@@ -12,7 +12,7 @@ export const kusamaDefinitions: MetadataConsts[] = [
 	{
 		runtimeVersions: [
 			2027, 2028, 2029, 2030, 9000, 9010, 9030, 9040, 9050, 9070, 9080, 9090,
-			9100, 9110, 9111,
+			9100, 9110, 9111, 9122,
```
niklasad1, line 15: 9120 is missing?

Included in v0.9.12 AFAIU

jsdw: DQ: How would I go about verifying that this list of versions is correct?

TarikGul: > 9120 is missing?

Kusama doesnt have 9120 actually (or atleast via polkadot-js types known).

> DQ: How would I go about verifying that this list of versions is correct?

https://github.com/polkadot-js/api/blob/master/packages/types-known/src/upgrades/kusama.ts

It is updated pretty religiously.

TarikGul:
> why isn't polkadot bumped as well?

@niklasad1 In this case I am just following Jaco. I didnt add it to the list in blocksStore, the idea being, what if polkadot doesnt have 9122 but instead they have to bump the runtime to v9123 before parachains are launched or something of the liking which ends up skipping v9122.

insipx (approved):

jsdw (approved):

Issue #743, opened 2021-11-01
Cannot read property 'validators' of undefined


tiendq:
**Description**

It has worked well for my local docker Moonbeam, but then fails with public testnet.

**Steps to Reproduce**

`NODE_ENV=production SAS_EXPRESS_BIND_HOST=0.0.0.0 SAS_EXPRESS_LOG_MODE=all SAS_LOG_LEVEL=http SAS_SUBSTRATE_WS_URL=wss://wss.testnet.moonbeam.network substrate-api-sidecar`

`curl http://localhost:8080/blocks/1053155 | jq` (this block here https://moonbase-blockscout.testnet.moonbeam.network/blocks/1053155/transactions)

**Expected vs. Actual Behavior**

I expected to get a block, instead got this error:

```
2021-10-29 11:28:21 error: Cannot read property 'validators' of undefined 
 TypeError: Cannot read property 'validators' of undefined
    at BlocksService.fetchBlock (/Users/tiendq/.nvm/versions/node/v14.17.0/lib/node_modules/@substrate/api-sidecar/build/src/services/blocks/BlocksService.js:45:39)
    at BlocksController.getBlockById (/Users/tiendq/.nvm/versions/node/v14.17.0/lib/node_modules/@substrate/api-sidecar/build/src/controllers/blocks/BlocksController.js:134:68)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async /Users/tiendq/.nvm/versions/node/v14.17.0/lib/node_modules/@substrate/api-sidecar/build/src/controllers/AbstractController.js:168:9
2021-10-29 11:28:21 error: GET /blocks/1053155 500 870ms
```

tiendq:
@TarikGul so how's about it this time? I can see the block in the explorer but failed to read it with sidecar.

TarikGul:
So the issue here is it's failing on not having the `session` module on moonbeam. Not super recently, but I had changed the way we query blocks from using api.derive to a stripped down version of the call. I query all the validators in order to get the author of the block. I am able to reproduce this so I will start looking for a sensible solution.

TarikGul:
Ahh okay so the reason it doesnt fail on the explorer is this line. https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/chain/getBlock.ts#L35

ill get a fix in asap. Thanks for reporting it

TarikGul:
@tiendq Fix is in, and will be in our next release on Wednesday.

PR #744, opened 2021-11-01
fix: confirm session module before block retrieval

TarikGul:
closes: [#743](https://github.com/paritytech/substrate-api-sidecar/issues/743)

Updates the blocks endpoint to not fail when the connected chain doesn't have the session module to gather validators.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -4,6 +4,7 @@ import { extractAuthor } from '@polkadot/api-derive/type/util';
 import { Compact, GenericCall, Struct, Vec } from '@polkadot/types';
 import { AbstractInt } from '@polkadot/types/codec/AbstractInt';
 import {
+	AccountId32,
 	Block,
 	BlockHash,
 	BlockNumber,
@@ -102,7 +103,7 @@ export class BlocksService extends AbstractService {
 
 		const [{ block }, validators, events, finalizedHead] = await Promise.all([
 			api.rpc.chain.getBlock(hash),
-			historicApi.query.session.validators(),
+			this.fetchValidators(historicApi),
 			this.fetchEvents(historicApi),
 			queryFinalizedHead
 				? api.rpc.chain.getFinalizedHead()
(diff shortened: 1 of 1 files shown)


jsdw (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -644,6 +644,20 @@ export class BlocksService extends AbstractService {
...
+	 * validators.
+	 *
+	 * @param historicApi ApiDecoration to use for the query
+	 */
+	private async fetchValidators(
+		historicApi: ApiDecoration<'promise'>
+	): Promise<Vec<AccountId32>> {
+		return historicApi.query.session
+			? historicApi.query.session.validators()
+			: ([] as unknown as Vec<AccountId32>);
```
insipx, line 658: I think this was brought up before but i forgot; why do we need to cast to unknown before Vec?

insipx (approved):

src/services/blocks/BlocksService.ts
TarikGul: So this comes down to the typescript compiler. Basically, the functions expects `Vec<AccountId32>` to be returned, and when we return [] its of type Array<any>. Now the compiler sees `Vec<AccountId32>` and notices there is no straight type conversion from `Vec<AccountId32>` -> `Array<string>`, therefore it throws a type conversion error (Even though we know the exact output and input for the values). To get around type conversion issue we have to set it to unknown so the typescript compiler sees that the type is convertable.

src/services/blocks/BlocksService.ts
```diff
@@ -644,6 +644,20 @@ export class BlocksService extends AbstractService {
...
+	 * Checks to see if the current chain has the session module, then retrieve all
+	 * validators.
+	 *
+	 * @param historicApi ApiDecoration to use for the query
+	 */
+	private async fetchValidators(
+		historicApi: ApiDecoration<'promise'>
+	): Promise<Vec<AccountId32>> {
+		return historicApi.query.session
+			? historicApi.query.session.validators()
```
TarikGul, line 657:
```suggestion
			? await historicApi.query.session.validators()
```

Issue #476, opened 2021-03-16
/transaction/material enhancement


vae520283995:
I hope that the metadata returned by /transaction/material is the decoded json data, which is the same as /runtime/metadata, otherwise I need to request the network twice to obtain the network information needed to construct the transaction

emostov:
@vae520283995 why do you need the decoded metadata? 
You should be able to use the encoded metadata to construct the transaction with something like txwrapper or just the @polkadot/metadata package in general.

If there is a compelling reason to add this feature we can probably put it behind a query param so it is not a breaking change and the default payload is smaller.

vae520283995:
@emostov Thank you, Because the transaction contains call_index, for example, Balances_transfer is 0500, and Balances_transfer_keep_alive is 0503. I am worried that this value will change according to the metadata update, so I get it from the decoded metadata, so I need the decoded metadata. By the way, this value will be Change?

emostov:
> By the way, this value will be Change?

Yes, in theory any value in the metadata can change so it is important to always use the latest metadata when constructing a transaction.

> Because the transaction contains call_index, for example, Balances_transfer is 0500, and Balances_transfer_keep_alive is 0503. 

The encoded metadata contains this info as well, but you need a scale codec library to decode it, such as txwrapper or polkadot.js. What are you using to create transactions?

vae520283995:
> The encoded metadata contains this info as well, but you need a scale codec library to decode it, such as txwrapper or polkadot.js. What are you using to create transactions?

I didn't use any package, I created the transaction according to the api of substrate-api-sidecar, so I hope the metadata returned by /transaction/material is the decoded data

emostov:
> I didn't use any package, I created the transaction according to the api of substrate-api-sidecar, so I hope the metadata returned by /transaction/material is the decoded data

Can you clarify why you need the decoded metadata? All the tooling I know of to make transactions can use the encoded metadata already provided in the transaction/material endpoint.

Additionally, since the decoded data is already available through another endpoint it is hard to justify adding this functionality.

vae520283995:
> > I didn't use any package, I created the transaction according to the api of substrate-api-sidecar, so I hope the metadata returned by /transaction/material is the decoded data
> 
> Can you clarify why you need the decoded metadata? All the tooling I know of to make transactions can use the encoded metadata already provided in the transaction/material endpoint.
> 
As I said before, call_index is required in the transaction, and call_index comes from the decoded metadata

[...]

PR #746, opened 2021-11-01
feat: add query param `metadata`

TarikGul:
closes: [#476](https://github.com/paritytech/substrate-api-sidecar/issues/476)

Pertains to `/transaction/material`

This adds the query param `metadata` that is acting as a replacement for the `noMeta` query param. 

`metadata` will take in a format for what type of data you want back. It accepts 'json' or 'scale'. 
json -> decoded metadata as json
scale -> SCALE encoded hex value of the metadata

`noMeta` will still be available for 2 more weekly releases then its subject to be removed. The currrent default behavior is that when there is no query param scale encoded metadata is returned. This will change when this is deprecated, and the default behavior will be no metadata unless the `metadata` field is specified.


src/services/transaction/TransactionMaterialService.ts
```diff
@@ -11,7 +11,8 @@ export class TransactionMaterialService extends AbstractService {
 	 */
 	async fetchTransactionMaterial(
 		hash: BlockHash,
-		noMeta: boolean
+		noMeta: boolean,
+		decodeMeta: boolean
```
jsdw, line 15: Just wondering; APi wise I guess we want to avoid breaking changes where possible, but I wonder about combining `decodeMeta` and  `noMeta` into one param that would be used like:

- `/?metadata=json` - return metadata decoded as JSON.
- `/?metadata=scale` - return the scale encoded blob.
- If `metadata` is no provided, don't return any metadata.

Just because at the mo you can provide combinations that don't make sense like `?noMeta&decodeMeta`.

Just a thought :)

jsdw (approved):

src/services/transaction/TransactionMaterialService.ts
TarikGul: This is a great idea, and an enhancement for sure. The one issue being that it would introduce a breaking change, but I think it's justifiable, and cleaner. 

@emostov Any thoughts on this?

TarikGul: So I discussed it in the integration channel, and I am going to go forward with your idea and implement `metadata` query param. The only thing being that we want to give users ample time to update because of how often we release sidecar. So I am going to have 2 params here. One is `noMeta`, and the other is `metadata`. Well allow users to migrate over something like 2 releases, then remove `noMeta` completely

emostov: this sounds good to me

emostov (approved):

docs/src/openapi-v1.yaml
```diff
@@ -685,6 +685,12 @@ paths:
           description: If true, does not return metadata hex. This is useful when
             metadata is not needed and response time is a concern. Defaults to false.
           default: false
+      - name: decodeMeta
+        in: query
+        schema:
+          type: boolean
+          description: If true, the metadata returned will be decoded and JSON formatted. 
+          default: false
```
jsdw, line 693: I think this needs removing again, then? :)

docs/src/openapi-v1.yaml
```diff
@@ -2707,7 +2713,8 @@ components:
             does not change `txVersion`.
         metadata:
           type: string
-          description: The chain's metadata in hex format.
+          description: The chain's metadata in hex format. When the query parameter
+            `decodeMeta` is true the metadata will returned a decoded object of the metadata.
```
jsdw, line 2717: And this wants an update

src/controllers/transaction/TransactionMaterialController.ts
```diff
@@ -51,16 +54,43 @@ export default class TransactionMaterialController extends AbstractController<Tr
...
+		if (metadata === 'json') return 'json';
+		if (!noMeta || noMeta === 'false' || metadata === 'scale') return 'scale';
+		if (noMeta === 'true') {
+			Log.logger.warn(
+				'`noMeta` query param will be deprecated in sidecar v13, and replaced with `metadata` please migrate'
+			);
+			return false;
+		}
+
+		return 'scale';
```
jsdw, line 94: I was pondering whether this could be simplified. The best I came up with was:

```
if (!noMeta && metadata === 'json')
    return 'json';
if (!noMeta && metadata === 'scale')
    return 'scale';
if (!noMeta && metadata)
    throw new Error('Invalid inputted value for the `metadata` query param.');

if (noMeta) {
    Log.logger.warn(
        '`noMeta` query param will be deprecated in sidecar v13, and replaced with `metadata` please migrate'
    );
}
return false;
```

src/services/transaction/TransactionMaterialService.spec.ts
```diff
@@ -9,15 +9,46 @@ const transactionMaterialService = new TransactionMaterialService(
...
 				sanitizeNumbers(
 					await transactionMaterialService.fetchTransactionMaterial(
 						blockHash789629,
-						false
+						'scale'
 					)
 				)
 			).toStrictEqual(response789629);
 		});
+
```
jsdw, line 22: Is it worth also haveing a test that regardless of `metadata`, if `noMeta` is set, nothing is returned?

docs/src/openapi-v1.yaml
TarikGul: Yea the docs still need to get updated :)

src/controllers/transaction/TransactionMaterialController.ts
TarikGul: Yea I agree, when I wrote this I knew it was really late, and super messy looking and was in need of a re-write 😅 . Thank you for working through that,  I had a lot of trouble grouping the arguments together. I agree the above is significantly better and ill update it!

src/services/transaction/TransactionMaterialService.spec.ts
TarikGul: Yea, I agree. I was thinking about that, and that will have to be tested using the controller. We send the service a argument `metadata` which can either be a string denoting the type of metadata we are going to return, or a value of false, meaning no metadata (You rewrote this section above). Little nit pick of my original logic, but do you think we should seperate those types and have two arguments passed in `metadata: boolean`, and `metadatFormat: string`, instead of just `metadata: string | false`.

src/controllers/transaction/TransactionMaterialController.ts
TarikGul: So the only issue with the logic above is that we need to account for when no query params are inputted the default behavior is to have scale metadata returned. 

```
/**
		 * Checks to see if the `metadata` query param is inputted, if it isnt,
		 * it will default to the old behavior. This is to be removed once after 
		 * the `noMeta` query param is fully deprecated.
		 */
		if (metadata) {
			switch (metadata) {
				case('json'):
					return 'json';
				case('scale'):
					return 'scale';
				default:
					throw new Error('Invalid inputted value for the `metadata` query param.');
			}
		}

		if (noMeta) {
			Log.logger.warn(
				'`noMeta` query param will be deprecated in sidecar v13, and replaced with `metadata` please migrate'
			);
			if (noMeta === 'true') return false;
			if (noMeta === 'false') return 'scale';
		}

		// default behavior until `noMeta` is deprecated, then false will be default
		return 'scale';
```

TarikGul: So i updated the logic a little, and kept the same format :). Thanks for making this easy for me to understand!

src/services/transaction/TransactionMaterialService.spec.ts
TarikGul: Considering the rework of the logic above, I think it's okay to have the current format. But tests to check are definitely needed.

src/controllers/transaction/TransactionMaterialController.ts
TarikGul: Also the idea here being that we want the metadata query param to override `noMeta` if its used

jsdw: Aaah overriding noMeta, got it! your new version looks clean :)

src/services/transaction/TransactionMaterialService.spec.ts
jsdw: I've no strong opinions on the input args really :) coming from the query params I guess things will always be a touch messy until they have been parsed into something nicer!

PR #749, opened 2021-11-04
fix: support calamari parachain fee calculations

grenade:


TarikGul:
Hey @grenade thanks for the PR. I'll have a run at this soon, just to verify, but from a quick glance it looks good. Youll just have to run `build.sh` file inside of `/<ROOT>/calc`

grenade:
> Hey @grenade thanks for the PR. I'll have a run at this soon, just to verify, but from a quick glance it looks good. Youll just have to run `build.sh` file inside of `/<ROOT>/calc`

thank you! i've actually not been able to prove this works yet (hence the `draft` pr).
https://api.serrano.calamari.systems/blocks/311056 is running the version of code in this pr but still shows "Fee calculation not supported for this network" so i'm still trying to work out why the fee calculation doesn't appear to work.

TarikGul:
Is there a public node I can test this against?

grenade:
> Is there a public node I can test this against?

yes. any of:
- wss://chilli.calamari.systems
- wss://falafel.calamari.systems
- wss://fritti.calamari.systems
- wss://smoothie.calamari.systems
- wss://jalapeno.calamari.systems

TarikGul:
Could you provide me with a block number/hash that contains an extrinisic that would require a fee as well?

TarikGul:
nvm found a block.

src/chains-config/calamariControllers.ts
```diff
@@ -0,0 +1,37 @@
...
+		'RuntimeMetadata',
+		'RuntimeSpec',
+		'TransactionDryRun',
+		'TransactionFeeEstimate',
+		'TransactionMaterial',
+		'TransactionSubmit',
+	],
+	options: {
+		finalizes: true,
+		minCalcFeeRuntime: null,
```
TarikGul, line 33: It's this line here that is causing it to error, set it to 1 assuming your minimum runtime version that supports fees is 1.

grenade: thank you! that makes good sense.

grenade: with that changed the error message now includes the runtime version. eg: "Fee calculation not supported for 4#calamari" from [/blocks/311056](https://api.serrano.calamari.systems/blocks/311056)

TarikGul: Hmm, I was able to get it working locally. To test it locally did you link the calc package?

To do that inside of the package.json you will replace the version of `@substrate/calc` to -> `"file:./calc/pkg"` in order to get the updated wasm generated code.

grenade: thank you again! i really appreciate your help with making this work. that does indeed fix it.

calc/pkg/package.json
```diff
@@ -18,4 +18,4 @@
   "main": "calc.js",
   "homepage": "https://github.com/paritytech/substrate-api-sidecar/tree/master/calc#readme",
   "types": "calc.d.ts"
-}
+}
```
TarikGul, line 21: One last nit could you add the space here. When the wasm compiles each time it trims this whitespace.. Sorry abou t that thanks :)

TarikGul (approved):
lgtm

calc/pkg/package.json
grenade: yes, thanks!

insipx (approved):

TarikGul:
After the tests, I will merge. Just as a heads up these changes will be reflected in next weeks release.

PR #754, opened 2021-11-09
fix: support calc fee for Karura and Acala

xlc:
I cannot test locally. I guess it is because we need to publish the calc package first before the wasm can be used.


src/chains-config/metadata-consts/karuraConsts.ts
```diff
@@ -0,0 +1,12 @@
+import { MetadataConsts } from '../../types/chains-config';
+import { extrinsicBaseWeight } from './substrateConsts';
+
+export const karuraDefinitions: MetadataConsts[] = [
+	{
+		runtimeVersions: [
+			1000, 1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, 1009, 1010, 1011,
```
xlc, line 7: Do we need to list all the runtime versions? Do we need to PR here if we did a runtime upgrade?

calc/src/calc_fee.rs
```diff
@@ -63,6 +63,9 @@ impl Multiplier {
 
             ("calamari", _v) => V2(new_u128(inner)),
 
+            ("karura", _v) => V2(new_u128(inner)),
```
xlc, line 66: should be a way to configure this without modify rust code. at least for common cases.

TarikGul: Yea we have had discussion about this, especially since parachains are going to want more and more support for fee calculations as time moves on. We thought about a plugin, and some different options. Still something to consider but I agree.

src/chains-config/metadata-consts/karuraConsts.ts
TarikGul: Yea and yes. If it doesnt have your new runtime version immediately thats alright, we just end up searching the decorated metadata for the weights for the missing runtime then cache it for the server instance.

TarikGul:
I will test it for you as well. But if you want to test it locally it's pretty simple. Just link the calc/pkg in the `package.json`

So it should look like this for the `@substrate/calc` dep:

"@substrate/calc": "file:./calc/pkg"

Then `yarn`, and it will have your changes reflected locally. Just make sure not to push it up :)

xlc:
> I will test it for you as well. But if you want to test it locally it's pretty simple. Just link the calc/pkg in the `package.json`
> 
> So it should look like this for the `@substrate/calc` dep:
> 
> "@substrate/calc": "file:./calc/pkg"
> 
> Then `yarn`, and it will have your changes reflected locally. Just make sure not to push it up :)

Thanks. I guess this needs to be part of README?

xlc:
Trie a few block and the number is matching values from Subscan.

TarikGul:
Sweet, ran it locally and it checks out. 

> I guess this needs to be part of README?

Good point. I write up some proper guide for it.

TarikGul (approved):

emostov (approved):

niklasad1 (approved):
clean, LGTM

TarikGul:
As a note, this will be reflected in the next release of sidecar. Merging.

Issue #379, opened 2021-01-06
Add test coverage `AccountsStakingPayoutsService`


emostov:
If possible, this service should have test coverage similar to the other services. This will likely require adding fake response to the `mockApi`.

PR #762, opened 2021-11-11
fix: staking payouts historicApi, and add tests

TarikGul:
closes: [#379](https://github.com/paritytech/substrate-api-sidecar/issues/379)

This updates staking payouts to use the historicApi and also adds tests.

TODO:
- [x] remove at, and add historicApi
- [x] tests


src/services/accounts/AccountsStakingPayoutsService.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Addresses and data below were taken from era 533 around block ~7,760,000,
+ * on runtime v9122 Polkadot.
+ *
+ * The real world data has been modified to scale to the unit tests to act as mock data.
+ * This test suite also uses polkadotRegistryV9122
+ */
+
+/**
+ * Acts as a placement variable for some tests where the era isnt an instrumntal
```
niklasad1, line 40:
```suggestion
 * Acts as a placement variable for some tests where the era isn't an instrumental
```

niklasad1 (approved):
LGTM

src/services/accounts/AccountsStakingPayoutsService.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+	mockLedger,
+} from '../test-helpers/mock/accounts';
+import stakingPayoutsResponse from '../test-helpers/responses/accounts/stakingPayout.json';
+import { AccountsStakingPayoutsService } from './AccountsStakingPayoutsService';
+
+/**
+ * Addresses and data below were taken from era 533 around block ~7,760,000,
+ * on runtime v9122 Polkadot.
+ *
+ * The real world data has been modified to scale to the unit tests to act as mock data.
```
dvdplm, line 35: How do you mean "modified to scale to the unit tests"?

src/services/accounts/AccountsStakingPayoutsService.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+ * This test suite also uses polkadotRegistryV9122
+ */
+
+/**
+ * Acts as a placement variable for some tests where the era isn't an instrumental
```
dvdplm, line 40: What do you mean by "placement variable"?

src/services/test-helpers/mock/accounts/eraRewardPoints.ts
```diff
@@ -0,0 +1,21 @@
...
+const rewardTwo = polkadotRegistryV9110.createType('u32', 4440);
+const mapToBe = new Map();
+mapToBe.set('12JZr1HgK8w6zsbBj6oAEVRkvisn8j3MrkXugqtvc4E8uwLo', rewardOne);
+mapToBe.set('1HDgY7vpDjafR5NM8dbwm1b3Rrs4zATuSCHHbe7YgpKUKFw', rewardTwo);
+const bTree = new BTreeMap(
+	polkadotRegistryV9110,
+	'AccountId32',
+	'u32',
+	mapToBe
+);
```
dvdplm, line 15: How does the `BTreeMap` type work? Do these 4 args become two key/value pairs or…?

TarikGul: So here the `BTreeMap`, is taking in a `Map` object and will convert it's key, values into the BTree. This is held in the `mapToBe` variable

TarikGul: Probably better being `mappedValues` or something

src/services/accounts/AccountsStakingPayoutsService.spec.ts
TarikGul: So it's a placement variable for some of the mock test calls. We call a few of them in the test suite, and even though it doesn't need the `era` param to check the integrity of the test, we still need it to call the mock test call.

src/services/accounts/AccountsStakingPayoutsService.spec.ts
TarikGul: The original data set had all the validators and nominators for polkadot. I abstracted only 2 validators and 2 nominators so that the data is small and manageable, but also easy to configure and change for tests.

src/services/accounts/AccountsStakingPayoutsService.spec.ts
dvdplm: So you mean it's a placeholder of sorts?

src/services/accounts/AccountsStakingPayoutsService.spec.ts
dvdplm: ah, got it. Maybe just say "reduced to fit the unit tests"?

src/services/accounts/AccountsStakingPayoutsService.spec.ts
TarikGul: Yes exactly, placeholder is the right term.

src/services/test-helpers/mock/accounts/eraRewardPoints.ts
TarikGul: Switched the naming to `mockEraRewardsMap`

src/services/accounts/AccountsStakingPayoutsService.spec.ts
TarikGul: Awesome switched it to "reduced to fit the unit tests", sounds better.

src/services/accounts/AccountsStakingPayoutsService.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Addresses and data below were taken from era 533 around block ~7,760,000,
+ * on runtime v9122 Polkadot.
+ *
+ * The real world data has been reduced to fit the unit tests and act as mock data.
+ * This test suite also uses polkadotRegistryV9122
+ */
+
+/**
+ * Acts as a placement variable for some tests where the era isn't an instrumental
```
dvdplm, line 40:
```suggestion
 * Acts as a placeholder variable for some tests where the era isn't an instrumental
```

dvdplm (approved):

PR #772, opened 2021-11-23
chore(release): 11.3.1

dvdplm:
Release v11.3.1


jsdw (approved):
LGTM

As mentioned, I don't think ths yarn update represents any user visible change, so the patch bump feels ok :)

TarikGul (approved):
👍 Thanks for the small updates and updating yarn.

PR #777, opened 2021-11-29
fix: update polkadot-js deps, and fix tests, and types for assets

TarikGul:
Updates the following deps:

    "@polkadot/api": "^6.10.1",
    "@polkadot/util-crypto": "^8.0.2",

Before our type `IAssetBalance` was representing `balance: TAssetBalance` (where `TAssetBalance` is `u64`). Now that polkadot-js represents the balance in `PalletAssetsAssetBalance`, the type for balance is `u128`. This is now reflected in the types, and tests.

Update `AssetDetails` -> `PalletAssetsAssetDetails `

### DIFF
--- a/src/services/accounts/AccountsAssetsService.spec.ts
+++ b/src/services/accounts/AccountsAssetsService.spec.ts
@@ -81,9 +81,9 @@ const assetBalanceObjThree = {
 };
 
 const assetBalanceFactory = {
-	'10': assetBalanceObjOne as PalletAssetsAssetBalance,
-	'20': assetBalanceObjTwo as PalletAssetsAssetBalance,
-	'30': assetBalanceObjThree as PalletAssetsAssetBalance,
+	'10': assetBalanceObjOne as unknown as PalletAssetsAssetBalance,
+	'20': assetBalanceObjTwo as unknown as PalletAssetsAssetBalance,
+	'30': assetBalanceObjThree as unknown as PalletAssetsAssetBalance,
 };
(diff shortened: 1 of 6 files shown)


jsdw (approved):
LGTM! Shame about the unknowns :)

src/services/accounts/AccountsAssetsService.spec.ts
```diff
@@ -81,9 +81,9 @@ const assetBalanceObjThree = {
 };
 
 const assetBalanceFactory = {
-	'10': assetBalanceObjOne as PalletAssetsAssetBalance,
-	'20': assetBalanceObjTwo as PalletAssetsAssetBalance,
-	'30': assetBalanceObjThree as PalletAssetsAssetBalance,
+	'10': assetBalanceObjOne as unknown as PalletAssetsAssetBalance,
```
niklasad1, line 84: can you explain why the `unknown` type casts are needed?

niklasad1 (approved):

insipx (approved):

src/services/accounts/AccountsAssetsService.spec.ts
TarikGul: Yea because the typescript compiler doesnt have the ability to cast `Obj A` as `Obj B` since their contents are not the exact same type structure. `PalletAssetsAssetBalance` has a `extra` field of type Null that we don't add for this specific test because it's not needed.

PR #780, opened 2021-11-30
fix: update websocket addresses used for e2e tests

TarikGul:
Currently the public nodes hosted by parity have been unreliable. 

When I run the e2e tests i find myself switching over all the addresses to onFinality to run the tests.

Below are some issues i find when querying the following public nodes. 
`wss://rpc.polkadot.io` -> A query can take up to 100,000 ms 
`wss://westend-rpc.polkadot.io` -> Same with polkadot, incredibly slow. 
`wss://kusama-statemine-rpc.paritytech.net` -> Connection issues, drops connection when initializing the api.

Pros for onFinality nodes:
Haven't ran into a single issue yet for any chain.

### DIFF
--- a/scripts/config.ts
+++ b/scripts/config.ts
@@ -26,7 +26,7 @@ export const defaultSasPackOpts = {
 
 export const config: Record<string, IChainConfig> = {
 	polkadot: {
-		wsUrl: 'wss://rpc.polkadot.io',
+		wsUrl: 'wss://polkadot.api.onfinality.io/public-ws',
 		JestProcOpts: {
 			...defaultJestOpts,
 			args: ['test:e2e-tests', '--chain', 'polkadot'],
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

jsdw (approved):
It does feel slightly damning that we are moving away from our own RPC nodes (but Polkadot.js seems to pick others by default as well, doesn't it, so there's precedent!)

TarikGul:
> It does feel slightly damning that we are moving away from our own RPC nodes (but Polkadot.js seems to pick others by default as well, doesn't it, so there's precedent!)

Yea I agree. The difference can be shocking though, parity nodes can make the e2e tests go as long as 20-30 minutes where as onfinality is 5-10 max. I hope to switch them back eventually, but the future goal is to have the e2e tests connect to private parity nodes. But it isnt so far up my list yet.

PR #784, opened 2021-12-02
feat: add POST `/transaction/watch`

### LINKED ISSUE #783: How do I track the success or failure of a trade after trading?
I will get the txhash after broadcasting the transaction, but the txhash is not unique, so what is the best way to get the transaction information and make sure the transaction is successful?

Is there support for the `author_submitAndWatchExtrinsic` method?

TarikGul:
This is just a draft prototype but this implements a `/transaction/watch` endpoint. It uses `author_submitAndWatch` as it's rpc call. 

There is still much discussion to be had on this design and implementation but I wanted to have a starting point. Once I am done finishing other important tasks, I will get back to this.

The current return response looks like:

```
{
  "hash": "0x52ffe0d68ad0cde2ded94da2b3a355522ec3a12992cb2df44ae2504a05317b2e",
  "index": 3,
  "isFinalized": false,
  "isInBlock": true,
  "blockHash": {
    "inBlock": "0xfeed6aa3bb2d77141a6932f4363241ad86bd31b8f5558344c3822a80688c194a"
  }
}
```

closes: [#783](https://github.com/paritytech/substrate-api-sidecar/issues/783)


dvdplm:
`watch` sounds a bit like it'd only be tracking a transaction, but in reality it both submits and then tracks it. So perhaps `POST transaction/submitAndWatch`?

Mitumaru (approved):

TarikGul:
Closing as this is not ideal for the API to have. Being that it is stateless, and the lack of client side information stored this is not ideal to have in Sidecar V1. Definitely a possibility when designing Sidecar V2 but out of scope for this API.

PR #786, opened 2021-12-06
fix: bump polkadot-js deps

TarikGul:
This introduced a problem when sanitizing leasePeriod via the tests. I found where the issue is happening and wrote a rough solution for it, but I still need to find the root of it in polkadot-js common, and what is causing this issue via BigInt. 

There is also a state tracing update too.

### DIFF
--- a/src/sanitize/sanitizeNumbers.ts
+++ b/src/sanitize/sanitizeNumbers.ts
@@ -147,9 +147,9 @@ export function sanitizeNumbers(data: unknown): AnyJson {
 	}
 
 	/**
-	 * With polkadot-js/api v6.11.1, the instanceOf `data` may not be a `BN`, 
+	 * With polkadot-js/api v6.11.1, the instanceOf `data` may not be a `BN`,
 	 * but the constructor.name represents its real value which is a BN.
-	 * 
+	 *
 	 * TODO: Find the root of the issue with BN and the bigInt updates to polkadot-js/common
 	 * try to reproduce if can too
 	 */
--- a/src/sanitize/sanitizeNumbers.ts
+++ b/src/sanitize/sanitizeNumbers.ts
@@ -147,8 +147,8 @@ export function sanitizeNumbers(data: unknown): AnyJson {
 	}
 
 	/**
-	 * When the given data has a superclass of `BN`, we need to check the 
-	 * constructor to see if the data is a subclass of `BN` since `instanceof` 
+	 * When the given data has a superclass of `BN`, we need to check the
+	 * constructor to see if the data is a subclass of `BN` since `instanceof`
 	 * will not check the complete prototype chain.
 	 */
 	if (typeof data === 'object' && data && data.constructor.name === 'BN') {
--- a/src/sanitize/sanitizeNumbers.ts
+++ b/src/sanitize/sanitizeNumbers.ts
@@ -146,21 +146,6 @@ export function sanitizeNumbers(data: unknown): AnyJson {
 		return mapTypeSanitizeKeyValue(data);
 	}
 
-	/**
-	 * With polkadot-js/api v6.11.1, the instanceOf `data` may not be a `BN`,
-	 * but the constructor.name represents its real value which is a BN.
-	 *
-	 * TODO: Find the root of the issue with BN and the bigInt updates to polkadot-js/common
-	 * try to reproduce if can too
-	 */
-	if (
-		typeof data === 'object' &&
-		data !== null &&
-		data.constructor.name === 'BN'
-	) {
-		return (data as BN).toString(10);
-	}
-
 	if (data instanceof BN || typeof data === 'number') {
 		return data.toString(10);
 	}
--- a/src/sanitize/sanitizeNumbers.ts
+++ b/src/sanitize/sanitizeNumbers.ts
@@ -146,15 +146,6 @@ export function sanitizeNumbers(data: unknown): AnyJson {
 		return mapTypeSanitizeKeyValue(data);
 	}
 
-	/**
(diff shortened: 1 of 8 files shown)


TarikGul:
I am not able to reproduce the issue via polkadot-js for `api.consts.slots.leasePeriod`, therefore the issue must be in Sidecar.

TarikGul:
`now.sub(offset).div(leasePeriod) instanceof BN` -> false; (To ensure the value returned I updated the following logic)
`new BN(now.sub(offset).div(leasePeriod).toNumber()) instanceof BN` -> true

This fixes the serializing issue with 6.11.1.

src/services/blocks/trace/Trace.ts
```diff
@@ -614,7 +614,7 @@ export class Trace {
 							field2: 'free',
 						},
 						amount: {
-							value: free,
+							value: free.toNumber(),
```
emostov, line 617: if we need to do this we should do something like `.toString(10)`, else risk overflowing the javascript number type since these represent u128

src/services/paras/ParasService.ts
```diff
@@ -413,7 +413,7 @@ export class ParasService extends AbstractService {
 			return null;
 		}
 
-		return now.sub(offset).div(leasePeriod);
+		return new BN(now.sub(offset).div(leasePeriod).toNumber());
```
emostov, line 416: although this practically will likely never overflow, I think our API makes some guarantees that the sanitizer will always spit our numbers as strings. Where exactly are the tests failing?

src/services/blocks/trace/Trace.ts
TarikGul: Yea totally agree!

src/services/paras/ParasService.ts
emostov: Oh I see the BN is the outer most thing. I would ideally avoid the toNumber here, but practically probably ok

emostov: actually disregard what I am saying, block number is u32 so no overflow risk

TarikGul: So this is where it's a little weird. The return value here for `now.sub(offset).div(leasePeriod)` is no longer an instance of BN when returned. Thus giving back a weird and incorrect wonky `leasePeriodIndex`, ex: We expect `39` back in our tests yet got `27`. 

The `.constructor.name` would return 'BN', but ` data instanceof BN || typeof data === 'number'` would return `false` when running through `sanitizeNumbers`. 

Currently looking for the root cause of it. It's between `@polkadot/common` 8.0.5 -> 8.1.2

emostov: can you do `.toBn()` on the return values?

TarikGul: > can you do .toBn() on the return values?

From what I tried, it doesn't exist on the value.

dvdplm: > The return value here for now.sub(offset).div(leasePeriod) is no longer an instance of BN when returned.

I would not expect a division of two `BN`s to return a `BN` so at least that part is not surprising?

maciejhirsz: Looking over [bn.js source](https://github.com/indutny/bn.js/blob/4187ca213e91b41acf72be046072f2dc1f06d0de/lib/bn.js#L2428) it _should_ be a BN, unless they have some wonky stuff happening there where if the output fits in 53 bits it returns a regular number instead?

Could we maybe require a newer version of node.js and switch to the native BigInt instead?

TarikGul: So we currently already support BigInt natively by requiring a minimum of Node v14. 

> switch to the native BigInt instead?

A lot of polkadot-js's infrastructure is returning BN values, but yea I agree we could switch to BigInt, but it would take some refactoring which should probably exist in its own PR. 

That being said, I isolated this logic in a sanity check, and when I run it I am getting the expoected behavior as oppose to the above:


```javascript
const main = async () => {
    const api = new ApiPromise({
        provider: new WsProvider('wss://kusama-rpc.polkadot.io')
    });

    await api.isReady
    const leasePeriod = await api.consts.slots.leasePeriod;

    console.log(leasePeriod instanceof BN); // -> True
    console.log(leasePeriod.constructor.name); // -> u32
    console.log(new BN(20000).sub(new BN(0)).div(leasePeriod) instanceof BN) -> true
}
```

src/sanitize/sanitizeNumbers.ts
```diff
@@ -146,6 +146,15 @@ export function sanitizeNumbers(data: unknown): AnyJson {
 		return mapTypeSanitizeKeyValue(data);
 	}
 
+	/**
+	 * When the given data has a superclass of `BN`, we need to check the 
+	 * constructor to see if the data is a subclass of `BN` since `instanceof` 
+	 * will not check the complete prototype chain.
+	 */
+	if (typeof data === 'object' && data && data.constructor.name === 'BN') {
```
maciejhirsz, line 154: Instead of comparing string name you could just do `data.constructor === BN`. I'm still not sure if this is the correct check here necessarily, but if it works in practice it works.

maciejhirsz (approved):

src/sanitize/sanitizeNumbers.ts
maciejhirsz: Weird, but oh well.

TarikGul: For documentation sake: Fixed via setting a resolution on `bn.js`

jsdw (approved):

PR #799, opened 2021-12-21
docs: update readme for e2e-tests

TarikGul:
This cleans up the e2e tests readme.

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -1,30 +1,22 @@
 ## Summary
 
-This is a helper library for Sidecar to run runtime tests against specific chains, at certain blocks. 
+This is a helper library for Sidecar to run e2e tests against specific chains, at certain blocks. 
 
 ## Testing
 
-The below instructions are specific to runtime-tests against one chain. 
-If you are looking to run the runtime-tests against all 3 chains (Polkadot, Kusama, Westend) then run `yarn test:init-runtime-tests` in 
+The below instructions are specific to running the e2e-tests against one chain. 
+If you are looking to run the e2e-tests against all chains (Polkadot, Kusama, Westend, Statemine) then run `yarn test:init-e2e-tests` in 
 the root directory of sidecar.
 
 ### Polkadot 
 
-Lets first get sidecar ready in a seperate terminal.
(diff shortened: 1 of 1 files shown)


e2e-tests/README.md
```diff
@@ -1,30 +1,22 @@
...
-$ export SAS_SUBSTRATE_WS_URL=<network archive node>
-$ yarn
-$ yarn build && yarn start
-```
+`yarn test:init-e2e-tests:polkadot`
 
-Sidecar should now be connected to the node and running successfully. If you find a bug file an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues).
-
-Now lets run our runtime tests against polkadot. Go to a separate terminal and run:
+Thats it!
```
niklasad1, line 17:
```suggestion
That's it! 
```

niklasad1 (approved):

e2e-tests/README.md
```diff
@@ -1,30 +1,22 @@
...
 
 ### Polkadot 
 
-Lets first get sidecar ready in a seperate terminal.
+To run the tests against a single chain, you may use the following below. For more examples, reference the `<ROOT>/package.json`
 
-```
-$ cd substrate-api-sidecar
-$ git checkout <your_branch>
-$ export SAS_SUBSTRATE_WS_URL=<network archive node>
```
vivekvpandya, line 18: I think setting up environment variable is required otherwise it will search on local machine. So better to mention those 2 points clearly in the readme.

TarikGul: So this is not required by the user because the script itself handles it. There is a [setWsUrl](https://github.com/paritytech/substrate-api-sidecar/blob/master/scripts/sidecarScriptApi.ts#L10) which the scriptsApi provides. The script itself uses a config from [config.ts](https://github.com/paritytech/substrate-api-sidecar/blob/master/scripts/config.ts#L27) in order to get the wsUrl. 

The whole idea being the user should only ever have to use `yarn test:init-e2e-tests` or for a specific chain `yarn test:init-e2e-tests:polkadot`

TarikGul: That being said, I think it will be useful to add a section that gives steps to users adding their own chains to the process.

jsdw (approved):

PR #802, opened 2021-12-24
fix(calc): add tests for impl Multplier

TarikGul:
This adds tests for `impl Multiplier`;

Todo:
- [x] Github actions test

### DIFF
--- a/calc/src/calc_fee.rs
+++ b/calc/src/calc_fee.rs
@@ -26,7 +26,7 @@ struct Coefficient {
     degree: u8,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum Multiplier {
     V0(Fixed128Legacy),
     V1((FixedI128, bool)),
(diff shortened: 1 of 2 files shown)


calc/src/calc_fee.rs
```diff
@@ -220,3 +220,52 @@ fn new_u128(inner: &str) -> FixedU128 {
...
+mod test_fees {
+    use super::Multiplier;
+    use super::Multiplier::{V2, V1, V0};
+    use super::new_u128;
+    use super::new_i128;
+    use super::new_legacy_128;
+
+    #[test]
+    fn multiplier_from_spec_name() {
+        let inner = "500000000";
```
niklasad1, line 233: I'm no fan of calling this `inner`, what is it multiplier (coefficient)? 

fine, it's used widely in this crate....

niklasad1 (approved):

emostov (approved):

insipx (approved):

calc/src/calc_fee.rs
TarikGul: Yea I agree that it's not the most informative naming, and I'd be happy to update it to something more verbose in a follow up PR, but for this PR as you mentioned its probably wise to keep the continuity in the crate.

PR #807, opened 2021-12-28
fix(e2e-tests): add statemint, and update with 9130 runtime tests

TarikGul:
This PR focus's on adding statemint to the e2e tests, it also adds 9130 runtime tests, and cleans up some logic in the scripts and configuration. 

When reviewing you can ignore any JSON files, as those are just responses the tests expects. 

This PR will be followed by another PR that will add the `paras` endpoint to the polkadot e2e tests.


src/chains-config/statemintControllers.ts
```diff
@@ -24,7 +24,7 @@ export const statemintControllers: ControllerConfig = {
 	],
 	options: {
 		finalizes: true,
-		minCalcFeeRuntime: 2,
+		minCalcFeeRuntime: 601,
```
TarikGul, line 27: This will be removed and changed back to `2` once #806 goes in.

TarikGul: This is just temporary so I can successfully run the tests.

niklasad1 (approved):

emostov (approved):

PR #809, opened 2022-01-03
fix: update polkadot-js api to 7.2.1

TarikGul:
This PR does a couple of things. 

1. It refactors the asset mock data to keep code cleaner and DRY. 

2. Updates polkadot-js api to 7.2.1 which introduces some new logic. 
- The `reasons` field is now added and replaces `sufficient` as of v9160. This is because the return type of `query.assets.account()` went from `PalletAssetsAssetBalance` too `Option<PalletAssetsAssetAccount>`. 

There are 3 cases we look for, and I wrote them in the comments of the code. 

It's good to note as well, once 9160 is released we can change the structure of the response to include all the contents of `reasons`. This PR is to make it compatible with Sidecar without breaking any code, and to allow us to catch up with p.js

ref: 
https://github.com/paritytech/substrate/pull/10382/files#diff-9acae09f48474b7f0b96e7a3d66644e0ce5179464cbb0e00671ad09aa3f73a5fR88

### DIFF
--- a/src/services/accounts/AccountsAssetsService.ts
+++ b/src/services/accounts/AccountsAssetsService.ts
@@ -17,7 +17,10 @@ import { AbstractService } from '../AbstractService';
  * https://github.com/paritytech/substrate/pull/10382/files#diff-9acae09f48474b7f0b96e7a3d66644e0ce5179464cbb0e00671ad09aa3f73a5fR88
  *
  * Both `PalletAssetsAssetBalance`, and `LegacyPalletAssetsAssetBalance` have either,
- * a `sufficient` or `isSufficient` field exposed instead.
+ * a `sufficient` or `isSufficient` field exposed instead and are the types used
+ * pre v9160.
+ *
+ * For v9160 and future runtimes, the returned type is `PalletAssetsAssetAccount`.
  */
 interface PalletAssetsAssetBalance extends Struct {
 	readonly balance: u128;
--- a/src/services/accounts/AccountsAssetsService.ts
+++ b/src/services/accounts/AccountsAssetsService.ts
@@ -198,8 +198,8 @@ export class AccountsAssetsService extends AbstractService {
 				/**
 				 * This return value wont ever be reached as polkadot-js defaults the
 				 * `balance` value to `0`, `isFrozen` to false, and `isSufficient` to false.
-				 * This ensures the the typescript compiler is happy, but we also follow with
-				 * its convention.
+				 * This ensures that the typescript compiler is happy, but we also follow along
+				 * with polkadot-js/substrate convention.
 				 */
 				return {
 					assetId,
--- a/src/services/accounts/AccountsAssetsService.ts
+++ b/src/services/accounts/AccountsAssetsService.ts
@@ -125,8 +125,8 @@ export class AccountsAssetsService extends AbstractService {
 					address
 				);
 
-				let balance = null, 
-					isFrozen =  null,
+				let balance = null,
+					isFrozen = null,
 					reason = null;
 				if (assetBalance.isSome) {
 					({ balance, isFrozen, reason } = assetBalance.unwrap());
--- a/src/services/accounts/AccountsAssetsService.ts
+++ b/src/services/accounts/AccountsAssetsService.ts
@@ -12,13 +12,16 @@ import {
 import { AbstractService } from '../AbstractService';
 
 /**
- * This is a type that is necessary for any runtime pre 9160. It excludes the
- * `reason` field which 9160 introduces via the following PR.
+ * These two types (`PalletAssetsAssetBalance, LegacyPalletAssetsAssetBalance`) are necessary for any
+ * runtime pre 9160. It excludes the `reason` field which v9160 introduces via the following PR.
  * https://github.com/paritytech/substrate/pull/10382/files#diff-9acae09f48474b7f0b96e7a3d66644e0ce5179464cbb0e00671ad09aa3f73a5fR88
  *
- * Both `PalletAssetsAssetBalance`, and `LegacyPalletAssetsAssetBalance` have either,
- * a `sufficient` or `isSufficient` field exposed instead and are the types used
- * pre v9160.
+ * `LegacyPalletAssetsAssetBalance` which is the oldest historic type here had a `isSufficient`
+ * key. It was then updated to be `sufficient` which we represent here within `PalletAssetsAssetBalance`.
+ *
+ * v9160 removes the `sufficient` key typed as a boolean, and instead
+ * replaces it with a `reason` key. `reason` is an enum and has the following values
+ * in polkadot-js: (`isConsumer`, `isSufficient`, `isDepositHeld`, `asDepositHeld`, `isDepositRefunded`, `type`).
  *
  * For v9160 and future runtimes, the returned type is `PalletAssetsAssetAccount`.
  */
--- a/src/services/accounts/AccountsAssetsService.ts
+++ b/src/services/accounts/AccountsAssetsService.ts
@@ -1,4 +1,5 @@
 import { ApiDecoration } from '@polkadot/api/types';
+import { bool, Null, Struct, u128 } from '@polkadot/types';
 import { StorageKey } from '@polkadot/types';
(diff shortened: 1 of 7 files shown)


src/services/test-helpers/mock/assets/mockAssetData.ts
```diff
@@ -0,0 +1,157 @@
...
+			accounts: rococoRegistry.createType('u32', 10),
+			sufficients: rococoRegistry.createType('u32', 15),
+			approvals: rococoRegistry.createType('u32', 20),
+			isFrozen: falseBool,
+		};
+
+		return rococoRegistry.createType('AssetDetails', responseObj);
+	});
+
+export const assetsInfoKeysInjected = (): (() => Promise<AssetDetails>) => {
```
TarikGul, line 103: To highlight a change for reviewers. I use to directly inject the `keys` field into the `assetsInfo` in order to replicate the stucture of the object polkadot-js. I  used Object.assign() directly in the top layer of the file without enclosing it within its own function. I removed that by adding the exported `assetsInfoKeysInjected`. This way we don't tamper with `assetsInfo` ever within testing, but instead can have the keys if it's necessary with a shallow copy of the original object with the keys injected.

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -125,12 +147,47 @@ export class AccountsAssetsService extends AbstractService {
...
+				 * `sufficient` as a key.
+				 *
+				 * 3. The older legacy type of `PalletAssetsAssetBalance` has a key of `isSufficient` instead
+				 * of `sufficient`.
+				 *
+				 */
+				let balance = null,
+					isFrozen = null,
+					isSufficient = null;
+				if (assetBalance.isSome) {
```
TarikGul, line 166: This section feels unreadable to me just based off the linter alone. I propose we increase the characters per line for some more readability in a general sense.

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -10,6 +11,27 @@ import {
...
 import { AbstractService } from '../AbstractService';
 
+/**
+ * This is a type that is necessary for any runtime pre 9160. It excludes the
+ * `reason` field which 9160 introduces via the following PR.
+ * https://github.com/paritytech/substrate/pull/10382/files#diff-9acae09f48474b7f0b96e7a3d66644e0ce5179464cbb0e00671ad09aa3f73a5fR88
+ *
+ * Both `PalletAssetsAssetBalance`, and `LegacyPalletAssetsAssetBalance` have either,
+ * a `sufficient` or `isSufficient` field exposed instead.
+ */
```
jsdw, line 21: BEcause this talks about being the type needed pre-9160, just checking; should it be on the `LegacyPalletAssetsAssetBalance` type?

TarikGul: So the issue is they are both Legacy types, one is just way older than the other. 

Not sure what the best approach is for naming, but since `PalletAssetsAssetBalance` is not technically legacy yet because 0.9.16 is not active I chose this naming

jsdw: Oh, sorry, I didn't mean to imply that the type names were wrong; it just sounded like the comment related more to the other one, but I'm easy either way :)

src/services/accounts/AccountsAssetsService.ts
jsdw: I found this code a bit harder to read, and was sortof wondering whether duplicating a little more might make it easier to grok; like moving any variables needed into each branch and returning the resolt from each?

I had a quick go at it; something like this perhaps:

```
const assetBalance = await historicApi.query.assets.account(
	assetId,
	address
);

/**
 * The following checks for three different cases:
 */

	// 1. Via runtime v9160 the updated storage introduces a `reason` field,
	// and polkadot-js wraps the newly returned `PalletAssetsAssetAccount` in an `Option`.
if (assetBalance.isSome) {
	const balanceProps = assetBalance.unwrap();

	return {
		assetId,
		balance: balanceProps.balance,
		isFrozen: balanceProps.isFrozen,
		isSufficient: balanceProps.reason.isSufficient
	}
}

// 2. `query.assets.account()` return `PalletAssetsAssetBalance` which exludes `reasons` but has
// `sufficient` as a key.
if ((assetBalance as unknown as PalletAssetsAssetBalance).sufficient) {
	const balanceProps = assetBalance as unknown as PalletAssetsAssetBalance;

	return {
		assetId,
		balance: balanceProps.balance,
		isFrozen: balanceProps.isFrozen,
		isSufficient: balanceProps.sufficient
	}
}

// 3. The older legacy type of `PalletAssetsAssetBalance` has a key of `isSufficient` instead
// of `sufficient`.
if (assetBalance['isSufficient'] as bool) {
	const balanceProps = assetBalance as unknown as LegacyPalletAssetsAssetBalance;

	return {
		assetId,
		balance: balanceProps.balance,
		isFrozen: balanceProps.isFrozen,
		isSufficient: balanceProps.isSufficient
	}
}

// Fallback; should we throw if we get here?
return {
	assetId,
	balance: null,
	isFrozen: null,
	isSufficient: null,
};
```

Not a big deal either way; mainly I was interested to see if it could be cleaned up at all :)

src/services/test-helpers/mock/assets/mockAssetData.ts
jsdw: Cool, I like this change!

Since JS is so awful at preventing things from being mutated I do like the "safe" option of returning objects and such from functions for tests; that way you know nothing else can ever mess with them!

jsdw (approved):
Left a couple of comments, and I didn't go too deep, but lgtm :) Do as you wish with the comments!

src/services/accounts/AccountsAssetsService.ts
TarikGul: You make a great point though, and I think I will update the docs for sure. It really does sound like one is a legacy over the other.

src/services/accounts/AccountsAssetsService.ts
TarikGul: Updated to your snippet above. Thanks again, greatly improved readability.

TarikGul: So for the last part, I was on the same though train where throwing an error would make sense, but just remembered that polkadot-js always returns the values back as 0<u128>, and false<bool> if they don't exist. So I changes it slightly to reflect that

src/services/accounts/AccountsAssetsService.ts
TarikGul: Updated :)

src/services/test-helpers/mock/assets/mockAssetData.ts
TarikGul: 😺 Yea I totally agree. Glad I changed that in the refactoring in this PR, very thankful for the TS compiler in moments like this as well. Makes it super easy to cleanup.

dvdplm (approved):

src/services/accounts/AccountsAssetsService.ts
```diff
@@ -10,6 +11,30 @@ import {
...
 import { AbstractService } from '../AbstractService';
 
+/**
+ * This is a type that is necessary for any runtime pre 9160. It excludes the
+ * `reason` field which 9160 introduces via the following PR.
+ * https://github.com/paritytech/substrate/pull/10382/files#diff-9acae09f48474b7f0b96e7a3d66644e0ce5179464cbb0e00671ad09aa3f73a5fR88
+ *
+ * Both `PalletAssetsAssetBalance`, and `LegacyPalletAssetsAssetBalance` have either,
+ * a `sufficient` or `isSufficient` field exposed instead and are the types used
+ * pre v9160.
```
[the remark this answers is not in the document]
TarikGul: Yea it can be a little confusing. Thanks for the feedback. Ill go it and make it clearer.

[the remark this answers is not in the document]
TarikGul: So I cleaned up the wording completely, I think its way more clear (a bit longer) and doesn't feel clunky when reading it. 

I think the last wording I had made assumptions on how much the reader knew which I removed, and instead explained it more from the ground up.

PR #813, opened 2022-01-11
fix: error handling on account balances

### LINKED ISSUE #812: Return code for invalid balance-info query returns HTTP status500
**Description**

A query to `/sidecar/accounts/<account-id>/balance-info` with a non-substrate/invalid account-id returns HTTP status code 500, indicating a server error. 

Expected Behavior: The query returns code 4xx
Actual Behavior: The query returns code 500

**Proposed solution**
HTTP status code 400 (bad request) is more appropriate for incorrect input over 500 (server error).

TarikGul:
closes: [#812](https://github.com/paritytech/substrate-api-sidecar/issues/812)

This adds some more verbose error handling in the account Balances endpoint. Since we allow ethereum addresses to pass through our validate address middleware we need to also make sure that when the ethereum address is used against a substrate network that doesnt have support for it, it will error properly with the right code, and message.

### DIFF
--- a/src/services/accounts/AccountsBalanceInfoService.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.ts
@@ -174,8 +174,8 @@ export class AccountsBalanceInfoService extends AbstractService {
 	}
 
 	/**
-	 * Returns HttpError with the correct err message for querying accounts balances. 
-	 * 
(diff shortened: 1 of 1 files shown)


jsdw (approved):
Looks sane to me :)

emostov (approved):

PR #821, opened 2022-01-18
[ci] add benchmarks to pipeline

alvicsam:
Added benchmarks to pipeline

Part of https://github.com/paritytech/ci_cd/issues/174


.gitlab-ci.yml
```diff
@@ -96,6 +96,22 @@ build:
...
+  variables:
+    POLKADOT_ADDRESS:              "ws://polkadot-node.nodes:9944"
+  image:                           "alvicsam/sidecar-bench:3.0"
+  script:
+    - export SAS_SUBSTRATE_WS_URL=$POLKADOT_ADDRESS
+    - yarn
+    - yarn build
+    - yarn run main &
+    - sleep 10
+    - wrk -d10m -t4 -c12 --timeout 120s --latency -s ./ci/benchmarks/lightweight-bench.lua "http://127.0.0.1:8080"
```
TarikGul, line 111: This all seems incredibly flexible, thanks so much again! So if I understand this correctly we can actively update our benchmarking scripts through this directory, and fix any connecting calls through this yml, and it should proactively run the updated code?

alvicsam: Yes, that's correct. I'm still experimenting and it seems that 10 minutes shows more accurate result.
This job runs on each commit. If we merge it with current settings it'll run on each commit in each pr and on master/tag as well.

alvicsam:
@TarikGul 
I suggest merging this PR if it looks good for you.
Currently it's working like this:
- build and deploy logic is the same
- added jobs `benchmark`, `check-benchmark` and `push-benchmark`. They run directly after build and are not related to deploy jobs. They run only on master or on tag. 
- added `build-pr` and `benchmark-pr` jobs to test that docker image can be built during PR and run benchmarks manually if needed

`check-benchmark`  doesn't work because I need to adjust the script compares benchmark results. So while I'm doing this we can at least collect results

radupopa2010 (approved):

.gitlab-ci.yml
```diff
@@ -96,20 +137,118 @@ build:
...
+      if [[ $BUILD_LATEST_IMAGE ]]; then
+        buildah bud \
+        --format=docker \
+        --tag "$CONTAINER_REPO:$DOCKER_IMAGE_TAG" \
+        --tag "$CONTAINER_REPO:latest" "$DOCKERFILE_DIRECTORY"
+      else
+        buildah bud \
+        --format=docker \
+        --tag "$CONTAINER_REPO:$DOCKER_IMAGE_TAG" "$DOCKERFILE_DIRECTORY"
+      fi
```
TriplEight, line 156: without `|-` notation you could skip using `\` at the end of each line

.gitlab-ci.yml
```diff
@@ -96,20 +137,118 @@ build:
...
+  stage: check-benchmark
+  <<:                              *publish-refs
+  <<:                              *kubernetes-env
+  needs:
+    - benchmark
+  variables:
+    PROMETHEUS_URL:                "http://vm-longterm.parity-build.parity.io"
+    CI_IMAGE:                      "paritytech/benchmarks:latest"
+  script:
+    - echo "TBD"
```
TriplEight, line 229: should it be in this PR?

TriplEight (changes requested):
minor grumbles

.gitlab-ci.yml
alvicsam: Nope, just made a stub job

TriplEight (approved):

PR #824, opened 2022-01-21
fix(bug): address `?at` bug and `endingOffset` conditionals

### LINKED ISSUE #814: /paras/auction/current
**Description**

Version used : 11.3.5 (running on a Docker container)

When accessing /paras/auctions/current?at=10935634 the response is
` {"at":{"hash":"0xf131bf645d5860b103c7aa1e00866a78a2b1890c35bb90e395c3b43951d43970","height":"10935634"},"beginEnd":"10914000","finishEnd":"10986000","phase":"endPeriod","auctionIndex":"21","leasePeriods":["19","20","21","22","23","24","25","26"],"winning":null} `

However there is a winning bidder (with the paraId 2015), so I was not expecting a null value for that field in the response. But rather some object as described in the doc : 

`
  "winning": [
    {
      "bid": {
        "accountId": "string",
        "paraId": "string",
        "amount": "string"
      },
      "leaseSet": [
        "string"
      ]
    }
  ]`

**Steps to Reproduce**

Run substrate-api-sidecar on docker

Access /paras/auctions/current?at=10935634 url (or with no parameter, or any parameter which is a block that was created during any auction) 

See Description section.

TarikGul:
closes: #814 

Fixes `/paras/auctions/current`

## Bug 1

This bug is straight forward. The `at` query param was broken because the appropriate `hash` wasn't being passed into the header query. Therefore always retrieving the latest head. 

## Bug 2

We were calculating the endingOffset incorrectly with the updated `vrfDelay` code. A few things to reflect on.

1. During an endingPeriod we need to retrieve the [`sample`](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/auctions.rs#L336) as the offset which is passed into [`Winning::<T>::get(offset)`](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/auctions.rs#L450).

2. Similar for startingPeriod we should be returning `BN_ZERO` as noted [here](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/auctions.rs#L433).

3. For vrfDelay we return no winning information, as that should be queried by using that `at` query paramter for the last block in the `endingPeriod`.

^ cc: @emostov 

NOTE: I updated the tests to reflect better real worlds scenarios (In relation to the block numbers, not the test logic). 

TODO: 
- [X] Add unit tests


src/services/paras/ParasService.ts
```diff
@@ -223,18 +223,45 @@ export class ParasService extends AbstractService {
 			phase: IOption<AuctionPhase>,
 			winning;
 		if (auctionInfoOpt.isSome) {
+			/**
+			 * AuctionInfo:::<T>:get() when Some, it returns a tuple where the first item is the
+			 * lease period index that the first of the four contiguous lease periods
+			 * on auction is for. The second is the block number when the auction will
```
emostov, line 229:
```suggestion
			 * an auction is for. The second is the block number when the auction will
```
```suggestion
			 * on auction is for. The second is the block number when the auction will
```

src/services/paras/ParasService.ts
```diff
@@ -223,18 +223,45 @@ export class ParasService extends AbstractService {
 			phase: IOption<AuctionPhase>,
 			winning;
 		if (auctionInfoOpt.isSome) {
+			/**
+			 * AuctionInfo:::<T>:get() when Some, it returns a tuple where the first item is the
```
emostov, line 227: If `AuctionInfo:::<T>:get()` is Some  it returns a tuple where the first item is the

src/services/paras/ParasService.ts
```diff
@@ -223,18 +223,45 @@ export class ParasService extends AbstractService {
...
+				blockNumber,
+				beginEnd,
+				phase,
+				historicApi
+			);
 
 			const winningOpt = endingOffset
 				? await historicApi.query.auctions.winning<Option<WinningData>>(
 						endingOffset
 				  )
```
emostov, line 263:
```suggestion
```

src/services/paras/ParasService.ts
```diff
@@ -223,18 +223,45 @@ export class ParasService extends AbstractService {
...
-				: await historicApi.query.auctions.winning<Option<WinningData>>(
-						// when we are not in the ending phase of the auction winning bids are stored at 0
-						0
-				  );
-
+				: { isSome: undefined };
```
emostov, line 264:
```suggestion
```

src/services/paras/ParasService.ts
```diff
@@ -223,18 +223,45 @@ export class ParasService extends AbstractService {
...
 			if (winningOpt.isSome) {
```
emostov, line 265:
```suggestion
			if (endingOffset) {
			    let winningOffset = await historicApi.query.auctions.winning<Option<WinningData>>(
						endingOffset
				  );
			
```

src/services/paras/ParasService.ts
```diff
@@ -437,7 +463,26 @@ export class ParasService extends AbstractService {
...
+		const sampleLength = historicApi.consts.auctions
+			.sampleLength as BlockNumber;
+
+		switch (phase) {
+			case 'startPeriod':
+				return BN_ZERO;
+			case 'endPeriod':
+				return afterEarlyEnd.div(sampleLength).sub(BN_ONE);
+			case 'vrfDelay':
+				return BN_ZERO;
```
emostov, line 482: Whats the justification for returning zero in this case?

emostov: I would think it would be more intuitive to return the offset into the last ending sample. In other words if we are in the vrf delay phase, we display who was winning at the last block of the auction

src/services/paras/ParasService.ts
emostov: If understand correctly, these should be the equivalent logic of what you have, but a little simpler

src/services/paras/ParasService.spec.ts
```diff
@@ -485,9 +485,9 @@ describe('ParasService', () => {
 
 			const expectedResponse = {
 				at: expectedAt,
-				beginEnd: '1000',
-				finishEnd: '21000',
-				phase: 'vrfDelay',
+				beginEnd: '780000',
+				finishEnd: '800000',
+				phase: 'endPeriod',
```
emostov, line 490: Would be good to have a test for vrfDelay, startPeriod, endPeriod, and auction over

src/services/paras/ParasService.ts
TarikGul: So I am actually split between returning null in a vrfDelay phase (no winning information), and returning zero for an offset. 

The polkadot code doesn't return a zero, but instead errors when retrieving the offset during a `vrfDelay` phase, but winning data is still available during the vrfDelay when querying with zero via polkadot-js. Therefore I had it in. Whether it's right to do I am not totally sure tbh.

Probably safer to just result to the default `return null` in the case of `vrfDelay`?

src/services/paras/ParasService.ts
TarikGul: Nice refactor! :), yea like the change a lot

src/services/paras/ParasService.ts
emostov: In polkadot, during the vrf delay there is not really notion of "currently winning" because we are waiting for the offset into the ending period. As a user of this API though it makes the most sense to either return:

1)  who was winning for the last ending sample, as this makes the most sense when query in order. For example, say during the last sample para 5 was winning, but during first sample para 0 was winning. If we are returning 5 as winning for all the queries during the last sample, but then all of a sudden start returning 0 as winning during the VRF delay it is confusing

2) not return any winners, and clearly add to the docs they should query the last block of the ending period to see who was winning at the end of the election

> The polkadot code doesn't return a zero, but instead errors when retrieving the offset during a vrfDelay 

This is because you can't bid in an auction that has ended (the linked code is for `handle_bid`) (sorry if I am just pointing out the obvious but wanted to clarify :P)

emostov: The more I think about it, the more I think option 2 is "more correct" in terms of a consistent API

TarikGul: > we display who was winning at the last block of the auction

Ahh okay, yea I can see that making sense. 

Basically instead of calculating the `afterEarlyEnd` via `now.sub(beginEnd)` it would be `(finishEnd - 1).sub(beginEnd)` I believe the minus one because bid arent accepted on the last block. (Ill have to double check).

TarikGul: Weird how it's not updating the comments as were both typing haha!

> The more I think about it, the more I think option 2 is "more correct" in terms of a consistent API

Cool, yea at first I figured this was the best way to go about it too. So for now as you mentioned, ill actually just remove the whole `vrfDelay` check and have it default to null. Ill write some inline docs explaining how we handle the vrfDelay, then also update the user docs, explaining they should use the finishEnd as a block reference to query the winner. 

Also thanks for the explanation above super helpful.

src/services/paras/ParasService.ts
```diff
@@ -468,6 +468,10 @@ export class ParasService extends AbstractService {
 		 * A sample can be represented by `afterEarlyEnd` / `sampleLength`.
 		 * When we are in the endingPeriod, the offset is represented by:
 		 * `sample - 1`.
+		 *
+		 * If the current phase is in `vrfDelay`, and you are interested in
+		 * querying the winners of the auction that just finished, it is advised
+		 * to query the last block of the `endingPeriod`.
```
emostov, line 474: This should go in open api docs, if it isn't already

TarikGul: Yea I have it locally, just havent commited it yet.

TarikGul: its up now.

src/services/paras/ParasService.spec.ts
TarikGul: We were already testing for `endPeriod`, and when there is no current auctionInfo available. I added vrfDelay, and startPeriod and cleaned it all up so it makes sense together.

TarikGul:
So this is ready for a final review. 

Few takeaway: Currently queries for any block that has a runtime version of 9090 or before will not work, until the next release of polkadot-js. Specific to [this PR](https://github.com/polkadot-js/api/pull/4495) which introduces the patch. 

When finalized and review this should go in and be followed by a PR that updates the deps in order to introduce the fix, and then a PR to update the e2e tests.

src/services/paras/ParasService.ts
```diff
@@ -437,7 +462,28 @@ export class ParasService extends AbstractService {
 			return null;
 		}
 
-		return now.sub(new BN(beginEnd)).iadd(BN_ONE);
+		/**
+		 * The length of each sample to take during the ending period.
+		 *
+		 * A sample can be represented by `afterEarlyEnd` / `sampleLength`.
+		 * When we are in the endingPeriod, the offset is represented by:
+		 * `sample - 1`.
```
TarikGul, line 470:
```suggestion
		 * `sample`.
```

TarikGul:
https://github.com/paritytech/substrate-api-sidecar/pull/828 should fix the isssue with historical auctions

TarikGul:
Now works for all historical, and current runtimes

docs/src/openapi-v1.yaml
```diff
@@ -1045,7 +1045,10 @@ paths:
         Get the status of the current auction.
       description: |
         Returns an overview of the current of auction. There is only one auction
```
jsdw, line 1047:
```suggestion
        Returns an overview of the current auction. There is only one auction
```

docs/src/openapi-v1.yaml
```diff
@@ -1045,7 +1045,10 @@ paths:
...
-        at a time. If there is no auction most fields will be `null`.
+        at a time. If there is no auction most fields will be `null`. If the current
+        auction phase is in `vrfDelay` and you are looking to retrieve the winners,
+        it is advised to  query the last block of the `endingPeriod` phase for that 
+        auction.
```
emostov, line 1051:
```suggestion
        auction as there technically are no winners during the `vrfDelay` and thus the field is `null`.
```

emostov (approved):
lgtm - just some minor recs for open api docs

docs/src/openapi-v1.yaml
```diff
@@ -1045,7 +1045,10 @@ paths:
         Get the status of the current auction.
       description: |
         Returns an overview of the current of auction. There is only one auction
-        at a time. If there is no auction most fields will be `null`.
+        at a time. If there is no auction most fields will be `null`. If the current
+        auction phase is in `vrfDelay` and you are looking to retrieve the winners,
```
jsdw, line 1049: Just to check based on our chat; I guess you can retrieve the highest bidders, but perhaps you don't know the actual "winners" until after the vrfDelay? (I need to read up on auctions more!)

src/services/paras/ParasService.spec.ts
```diff
@@ -201,7 +200,7 @@ const slotsLeasesEntriesAt = () =>
  */
 export const auctionsInfoAt = (): Promise<Option<Vec<BlockNumber>>> =>
 	Promise.resolve().then(() => {
-		const beingEnd = rococoRegistry.createType('BlockNumber', 1000);
+		const beingEnd = rococoRegistry.createType('BlockNumber', 780000);
```
jsdw, line 203: Typo for `beginEnd`?

src/services/paras/ParasService.spec.ts
```diff
@@ -516,33 +538,60 @@ describe('ParasService', () => {
...
-				extrinsicsRoot:
-					'0x2d5d3fdb96b487d480b08b64ed69a65433c1713ae3579dd23704cb790aa3b2ae',
-				digest: {},
+		it('Should return the correct data during a startPeriod phase', async () => {
+			const overrideHeader = generateOverrideHeader(770000);
+			const header = polkadotRegistry.createType('Header', overrideHeader);
+
+			// Override the mockApi
+			(mockApi.rpc.chain.getHeader as unknown) = () =>
+				Promise.resolve().then(() => header);
```
jsdw, line 547: Just a general thought; I wonder whether there's a nice way to avoid needing to remember to "reset" this sort of thing at the end of the call. Eg with immer you might do something like:

```
const thisMockApi = immer(mockApi, mockApi => {
    (mockApi.rpc.chain.getHeader as unknown) = () =>
        Promise.resolve().then(() => header);
});
```

And then the original `mockApi` would be untouched. Just thinking out loud really :)

src/services/paras/ParasService.ts
```diff
@@ -223,21 +223,47 @@ export class ParasService extends AbstractService {
 			phase: IOption<AuctionPhase>,
 			winning;
 		if (auctionInfoOpt.isSome) {
+			/**
+			 * If `AuctionInfo:::<T>:get()` is `Some`, it returns a tuple where the first item is the
+			 * lease period index that the first of the four contiguous lease periods
+			 * an auction is for. The second is the block number when the auction will
```
jsdw, line 229:
```suggestion
			 * of an auction is for. The second is the block number when the auction will
```

jsdw (approved):
Looks good to me offhand; thanks for the walkthrough!

src/services/paras/ParasService.spec.ts
TarikGul: Yes, great catch 😅

src/services/paras/ParasService.ts
```diff
@@ -223,21 +223,47 @@ export class ParasService extends AbstractService {
...
+			 */
+			finishEnd = beginEnd.add(endingPeriod);
+			/**
+			 * We determine what our phase is so we can decide how to calculate our
+			 * ending offset.
+			 */
+			if (finishEnd.lte(blockNumber)) {
+				phase = 'vrfDelay';
+			} else {
+				phase = beginEnd.gt(blockNumber) ? 'startPeriod' : 'endPeriod';
```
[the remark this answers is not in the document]
TarikGul: So for the `vrfDelay` I actually took this into account and test for it, but this is a great catch, because I actually tested off by 2 instead of 1, so I will fix that!

docs/src/openapi-v1.yaml
TarikGul: So yea the actual "winners" are reflected by slots that are leased out. When `manage_auctions_end` is called in the auctions pallet, it [calculates the finals winners](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/auctions.rs#L568) then it will [lease out the slots](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/auctions.rs#L577). 

So Options for finding the current winners: Use the `AuctionClosed` event as an block indicator to query `/leases/current`, 

Query the last block before the `finishEnd` block in order to get the latest winning bids.

TarikGul: Going to update it to "latest winning bids" to be more accurate

TarikGul: fixed

src/services/paras/ParasService.spec.ts
TarikGul: Yea I am definitely open to this, as it's an unfortunate pattern we have in the testing and I think this could definitely help and improve it. If we do use this I think it would be useful to do a seperate PR that overhauls all the times where we change the mockApi then reestablish it.  This is super apparent in `/blocks` tests.

src/services/paras/ParasService.ts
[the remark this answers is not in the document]
TarikGul: fixed

src/services/paras/ParasService.spec.ts
TarikGul: TIL about `immer` btw. Thank you :)

jsdw: Yep agreed; a separate PR would def make more sense!

(Ah I love immer :D)

src/services/paras/ParasService.ts
jsdw: Just looked at this again and I'm confused at why I suggested adding `of`; clearly my brain wasn't working :D

PR #829, opened 2022-01-31
fix: Retrieving public endpoints with `createWsEndpoints` function

### LINKED ISSUE #782: Sidecar should parse all the public endpoints from apps-config, and replace our hardcoded values
These [endpoints](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/main.ts#L106) should be updated to take in a generated array of all the values from [`@polkadot/apps-config`](https://github.com/polkadot-js/apps/tree/master/packages/apps-config). With the substrate space rapidly growing, the amount of public endpoints are also increasing, and they should no longer be hardcoded in sidecars codebase. 

This is a great first issue for anyone looking to contribute.

Imod7:
### Related Issue
This PR is an attempt to solve the https://github.com/paritytech/substrate-api-sidecar/issues/782 issue.

### Suggested Solution
I used `createWsEndpoints` function to get the endpoints from [@polkadot/apps-config/endpoints](https://github.com/polkadot-js/apps/tree/master/packages/apps-config/src/endpoints) and then generated the new `publicWsUrls` array. 

Note : I have not used the `specName` as suggested in https://github.com/paritytech/substrate-api-sidecar/issues/782

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -104,15 +104,16 @@ function startUpPrompt(wsUrl: string, chainName: string, implName: string) {
 	const { logger } = Log;
 	const { config } = SidecarConfig;
 
-  // Retrieving public endpoints from @polkadot/apps-config/endpoints 
-  const publicWsUrls: string[] = [];
-  const endpoints = createWsEndpoints(<T = string>(): T => ('' as unknown as T));
-  let count = 0;
-  while (count < endpoints.length) {
-    if (endpoints[count].value && endpoints[count].info != 'local')
-      publicWsUrls.push(endpoints[count].value);
-    count += 1;
-  }
+	/**
+	 * Retrieving public endpoints from @polkadot/apps-config/endpoints
+	 */
+	const endpoints = createWsEndpoints(<T = string>(): T => '' as unknown as T);
+	const publicWsUrls: string[] = endpoints
+		.map(function (endpoint) {
+			if (endpoint.value && endpoint.info != 'local') return endpoint.value;
+			else return '';
+		})
+		.filter((x) => x !== '');
 
 	logger.info(
 		`Connected to chain ${chainName} on the ${implName} client at ${config.SUBSTRATE.WS_URL}`
--- a/src/main.ts
+++ b/src/main.ts
@@ -21,6 +21,7 @@ import '@polkadot/api-augment';
 
 import { ApiPromise } from '@polkadot/api';
 import * as apps from '@polkadot/apps-config/api';
+import { createWsEndpoints } from '@polkadot/apps-config/endpoints';
 import { WsProvider } from '@polkadot/rpc-provider';
 import { OverrideBundleType, RegistryTypes } from '@polkadot/types/types';
 import { json } from 'express';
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -102,26 +103,16 @@ async function main() {
...
-		'wss://mainnet-rpc.stafi.io',
-		'wss://rpc.subsocial.network',
-		'wss://full-nodes-lb.kilt.io:443',
-	];
+
+  // Retrieving public endpoints from @polkadot/apps-config/endpoints 
+  const publicWsUrls: string[] = [];
+  const endpoints = createWsEndpoints(<T = string>(): T => ('' as unknown as T));
+  let count = 0;
+  while (count < endpoints.length) {
```
niklasad1, line 111: I would prefer to iterate over the endpoints via an iterator to avoid the `count variable` and indexing into the array a few time such as

```
for (var endpoint of endpoints) {
 ...code goes here
}
```

why not do it or am I missing something?

src/main.ts
```diff
@@ -102,26 +103,17 @@ async function main() {
...
-		'wss://rpc.kulupu.corepaper.org/ws',
-		'wss://main1.nodleprotocol.io',
-		'wss://rpc.plasmnet.io/',
-		'wss://mainnet-rpc.stafi.io',
-		'wss://rpc.subsocial.network',
-		'wss://full-nodes-lb.kilt.io:443',
-	];
+	const endpoints = createWsEndpoints(<T = string>(): T => '' as unknown as T);
+	const publicWsUrls: string[] = endpoints
+		.map(function (endpoint) {
```
Imod7, line 112: Hey @niklasad1! 
Thank you for your review! I replaced the `while` loop with `map` to iterate through `endpoints`.

src/main.ts
```diff
@@ -102,26 +103,17 @@ async function main() {
...
+			if (endpoint.value && endpoint.info != 'local') return endpoint.value;
+			else return '';
```
Imod7, line 114: I had to add the `else` statement here otherwise I get the following error :
`error TS7030: Not all code paths return a value.`
So then I had some empty strings in my array which I removed by chaining a `filter` iterator after it.

niklasad1: If you swap the order of `map` and `filter` I think we can get rid of that.

```ts
const publicWsUrls: string[] = endpoints
                .filter(e => e.value && e.info != 'local');
		.map(e => e.value)
```

However, I'm no `TS/JS` expert but I still think I would prefer

```ts
 const publicWsUrls: string[] = [];
 const endpoints = createWsEndpoints(<T = string>(): T => ('' as unknown as T));
 for (var endpoint of endpoints) {
         if (endpoint.value && endpoint.info != 'local') {
              publicWsUrls.push(endpoint.value);
         }
 }
```

but I have no strong opinions but let's wait for @TarikGul to take a look so I don't lie to you :)

TarikGul: Yea I agree with @niklasad1 here that both these examples are improvements. I would also say i prefer the `for` loop over filter and map. 

1. As mentioned in a chat with me and @Imod7 there are performance gains (even if they are incredibly small here being that the max size of this vector would be probably no more than a 1000).

2. I think it's easier to read / debug, or modify.

niklasad1 (approved):
nice, LGTM

TarikGul:
LGTM, great job, and thank you for the PR.

TarikGul (approved):

src/main.ts
Imod7: Hello @TarikGul @niklasad1 ! Thank you both for taking the time to review and your super valuable feedback! 
Two more questions : 
1. Do I need to update the changelog file or it is not necessary for this type of change ?
2. Would it be useful to add a test for this change ? eg. a test that checks that in the generated array (`publicWsUrls`) there are at least some standard public urls we always expect to have there (polkadot / kusama)?

PR #830, opened 2022-02-01
fix(deps): update polkadot-js/api, and rxjs

TarikGul:
Updates deps to the following:

    "@polkadot/api": "^7.6.1",
    "rxjs": "^7.5.2",

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,10 +50,9 @@
     "test:test-release": "yarn start:test-release"
   },
   "dependencies": {
-    "@polkadot/api": "^7.5.2-2",
+    "@polkadot/api": "^7.6.1",
     "@polkadot/apps-config": "0.103.1",
     "@polkadot/util-crypto": "^8.3.3",
-    "@polkadot/x-rxjs": "^6.11.1",
     "@substrate/calc": "^0.2.7",
     "argparse": "^2.0.1",
     "confmgr": "1.0.7",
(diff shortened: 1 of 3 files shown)


TarikGul:
This introduces a warning every time a new runtime is queried but wont appear again once its cached:

```
2022-01-31 23:12:45 warn: ABLEREGISTRY: Unable to determine runtime Call type, cannot inspect sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic
2022-01-31 23:12:45 warn: ABLEREGISTRY: Unable to determine runtime Event type, cannot inspect frame_system::EventRecord
```

niklasad1:
> This introduces a warning every time a new runtime is queried but wont appear again once its cached:

Have you filed an issue upstream? However, it's just a warn shouldn't be a big deal I suppose.

niklasad1 (approved):

TarikGul:
> > This introduces a warning every time a new runtime is queried but wont appear again once its cached:
> 
> Have you filed an issue upstream? However, it's just a warn shouldn't be a big deal I suppose.

So I believe this is already known, as it's checking parachains metadata which explains why it only warns for relay chains. But that being said, it should be logging `PORTABLEREGISTRY` instead of `ABLEREGISTRY` which i'll report.

jsdw (approved):

TarikGul:
The warning's are still going to be there, but after doing some extensive testing, and reading the code, this is a non issue so I am merging. I will follow up in polkadot-js api as well.

PR #834, opened 2022-02-04
fix: update block weight store

TarikGul:
Update block weight store with updated runtimes


emostov (approved):

niklasad1 (approved):

src/chains-config/metadata-consts/polkadotConsts.ts
```diff
@@ -11,7 +11,7 @@ export const polkadotDefinitions: MetadataConsts[] = [
 	{
 		runtimeVersions: [
 			27, 28, 29, 30, 9000, 9010, 9030, 9033, 9050, 9070, 9080, 9090, 9100,
-			9110, 9122, 9130, 9140,
```
jsdw, line 14: I hope I haven't asked before but; is there an easy way for me to verify these versions?

jsdw (approved):

src/chains-config/metadata-consts/polkadotConsts.ts
TarikGul: An easy way would be subscan. 

The next best thing: https://github.com/polkadot-js/api/tree/master/packages/types-known/src/upgrades

PR #840, opened 2022-02-10
fix: correctly serialize a multsig call, then parse WrapperKeepOpaque, and WrapperOpaque

### LINKED ISSUE #835: v11.3.8 not parse call parameters
v11.3.8 return block info, not parse call parameters
image
This is the return information of the previous version
image
Mind if you can parse all call parameters, such as proposal preimage

TarikGul:
closes: [#835](https://github.com/paritytech/substrate-api-sidecar/issues/835)

This PR resolves serializing a MultiSig call and correctly parsing the args with the added `WrapperKeepOpaque`, and `WrapperOpaque` types. The code is a little repetitive to keep it readable, and understandable, also `parseGenericCalls` is heavily leveraging recursion so best to not mess with the logic much. 

I added an e2e test for the block mentioned in the issue to ensure we are tracking a `WrapperKeepOpaque` call. 

This is via polkadot-js/api v7.6.1

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -701,7 +701,8 @@ export class BlocksService extends AbstractService {
 				} else if (argument instanceof GenericCall) {
 					newArgs[paramName] = this.parseGenericCall(argument, registry);
 				} else if (paramName === 'call' && argument?.toRawType() === 'Bytes') {
-					// multiSig.asMulti.args.call is an OpaqueCall (Vec<u8>) that we
+					// multiSig.asMulti.args.call is either an OpaqueCall (Vec<u8>),
+					// WrapperKeepOpaque<Call>, or WrapperOpaque<Call> that we
 					// serialize to a polkadot-js Call and parse so it is not a hex blob.
 					try {
 						const call = registry.createType('Call', argument.toHex());
@@ -713,8 +714,6 @@ export class BlocksService extends AbstractService {
 					paramName === 'call' &&
 					argument?.toRawType() === 'WrapperKeepOpaque<Call>'
 				) {
-					// multiSig.asMulti.args.call is also an WrapperKeepOpaque<Call> (Vec<u8>) that we
-					// serialize to a polkadot-js Call and parse so it is not a hex blob.
 					try {
 						const call = registry.createType('Call', argument.toHex());
 						newArgs[paramName] = this.parseGenericCall(call, registry);
@@ -725,8 +724,6 @@ export class BlocksService extends AbstractService {
 					paramName === 'call' &&
 					argument?.toRawType() === 'WrapperOpaque<Call>'
 				) {
-					// multiSig.asMulti.args.call is also an WrapperKeepOpaque<Call> (Vec<u8>) that we
-					// serialize to a polkadot-js Call and parse so it is not a hex blob.
 					try {
 						const call = registry.createType('Call', argument.toHex());
 						newArgs[paramName] = this.parseGenericCall(call, registry);
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -700,30 +700,16 @@ export class BlocksService extends AbstractService {
 					newArgs[paramName] = this.parseArrayGenericCalls(argument, registry);
 				} else if (argument instanceof GenericCall) {
 					newArgs[paramName] = this.parseGenericCall(argument, registry);
-				} else if (paramName === 'call' && argument?.toRawType() === 'Bytes') {
+				} else if (
+					argument &&
+					paramName === 'call' &&
+					['Bytes', 'WrapperKeepOpaque<Call>', 'WrapperOpaque<Call>'].includes(
+						argument?.toRawType()
+					)
+				) {
 					// multiSig.asMulti.args.call is either an OpaqueCall (Vec<u8>),
 					// WrapperKeepOpaque<Call>, or WrapperOpaque<Call> that we
 					// serialize to a polkadot-js Call and parse so it is not a hex blob.
-					try {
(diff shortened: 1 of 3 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -709,6 +709,30 @@ export class BlocksService extends AbstractService {
 					} catch {
 						newArgs[paramName] = argument;
 					}
+				} else if (
+					paramName === 'call' &&
+					argument?.toRawType() === 'WrapperKeepOpaque<Call>'
+				) {
+					// multiSig.asMulti.args.call is also an WrapperKeepOpaque<Call> (Vec<u8>) that we
+					// serialize to a polkadot-js Call and parse so it is not a hex blob.
```
jarcodallo, line 717: Why not remove this comments?

src/services/blocks/BlocksService.ts
```diff
@@ -709,6 +709,30 @@ export class BlocksService extends AbstractService {
...
+						newArgs[paramName] = this.parseGenericCall(call, registry);
+					} catch {
+						newArgs[paramName] = argument;
+					}
+				} else if (
+					paramName === 'call' &&
+					argument?.toRawType() === 'WrapperOpaque<Call>'
+				) {
+					// multiSig.asMulti.args.call is also an WrapperKeepOpaque<Call> (Vec<u8>) that we
+					// serialize to a polkadot-js Call and parse so it is not a hex blob.
```
jarcodallo, line 729: same

src/services/blocks/BlocksService.ts
```diff
@@ -701,14 +701,35 @@ export class BlocksService extends AbstractService {
...
+				} else if (
+					paramName === 'call' &&
+					argument?.toRawType() === 'WrapperOpaque<Call>'
+				) {
+					try {
+						const call = registry.createType('Call', argument.toHex());
+						newArgs[paramName] = this.parseGenericCall(call, registry);
+					} catch {
+						newArgs[paramName] = argument;
+					}
```
jsdw, line 732: Ah ok, so basically these bits are the same as something like the following then I guess?:

```
} else if (argument && paramName === 'call' && ['Bytes', 'WrapperKeepOpaque<Call>', 'WrapperOpaque<Call>'].includes(argument.toRawType())) {
	// multiSig.asMulti.args.call is either an OpaqueCall (Vec<u8>),
	// WrapperKeepOpaque<Call>, or WrapperOpaque<Call> that we
	// serialize to a polkadot-js Call and parse so it is not a hex blob.
	try {
		const call = registry.createType('Call', argument.toHex());
		newArgs[paramName] = this.parseGenericCall(call, registry);
	} catch {
		newArgs[paramName] = argument;
	}
} else {
	newArgs[paramName] = argument;
}
```

jsdw (approved):
LGTM, not a big fan of the repetition since it is identical even at a type level but up to you :)

src/services/blocks/BlocksService.ts
TarikGul: Yea that is way cleaner, ill change it to reflect that. Thanks!

TarikGul: fixed

TarikGul:
@jsdw I made the changes to stop the repetition, i totally agree that the solution you provided was way better.

src/services/blocks/BlocksService.ts
```diff
@@ -700,8 +700,15 @@ export class BlocksService extends AbstractService {
...
-					// multiSig.asMulti.args.call is an OpaqueCall (Vec<u8>) that we
+				} else if (
+					argument &&
+					paramName === 'call' &&
+					['Bytes', 'WrapperKeepOpaque<Call>', 'WrapperOpaque<Call>'].includes(
+						argument?.toRawType()
+					)
+				) {
+					// multiSig.asMulti.args.call is either an OpaqueCall (Vec<u8>),
+					// WrapperKeepOpaque<Call>, or WrapperOpaque<Call> that we
```
dvdplm, line 711: It's a bit hard to map the code in the `try…catch` block to each of the three cases. What do you think about something like this:

```ts
			} else if (
				argument &&
				paramName === 'call'
			) {
				switch (argument?.toRawType()) {
					case 'Bytes':
					case 'WrapperKeepOpaque<Call>':	
						newArgs[paramName] = argument;
					case 'WrapperOpaque<Call>':
						const call = registry.createType('Call', argument.toHex());
						newArgs[paramName] = this.parseGenericCall(call, registry);
	
					default:
						// log or otherwise report error?;
				}
```

(Not sure I got the cases right, which is illustrating the point I guess)

niklasad1: I like that if that works :)

niklasad1 (approved):

src/services/blocks/BlocksService.ts
TarikGul: So actually each of the 3 cases are going to map out to the same output i.e.
```
					try {
						const call = registry.createType('Call', argument.toHex());
						newArgs[paramName] = this.parseGenericCall(call, registry);
					} catch {
						newArgs[paramName] = argument;
					}
```

therefore the above wont work. I filed an issue [here](https://github.com/paritytech/substrate-api-sidecar/issues/845) regarding the `try catch` in order to potentially optimize it. For this PR i think it is out of scope as we are just extending the conditional to include `WrapperKeepOpaque<Call>`, and `WrapperOpaque`.

TarikGul:
Merging. @dvdplm I think your comment has some merit above, and should be given some extra investigation and potential optimization. That being said, for the scope of this PR, lets just patch the bug (since the solution itself is pretty simple and doesnt change any of the core logic when parsing the generic calls). But I will follow this PR up with the issue I filed and see what we can do :)

PR #842, opened 2022-02-10
Abstract error handling for addresses into AbstractService

### LINKED ISSUE #823: Abstract error handling for addresses into AbstractService
Extension of #813 

Now that we correctly handle an error when ethereum addresses are passed into `AccountBalanceInfoService` (ie: [`createHttpError`](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/accounts/AccountsBalanceInfoService.ts#L182)), lets abstract that out and put it into the `AbstractService` as `createHttpErrorForAddr`. 

Let's then use that in each accounts service within a `catch` method. Well then throw the correct error inside of that `catch` method using the above function we just created. It would look like the following [line of code](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/accounts/AccountsBalanceInfoService.ts#L49);

This will also fix the following potential error that arises in #813 for the other endpoints that have address as a param. 

cc: @CurlyBracketEffect

0xslipk:


TarikGul:
Looks like the following test is failing because [this mock data](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/test-helpers/mock/assets/mockAssetData.ts#L131) used in the historicApi should be returning a resolved promise i.e `Promise<Foo>` (similar to the other mock functions in the file). If you could update that, that would be awesome, should fix the tests as well.

TarikGul:
closes: [#823](https://github.com/paritytech/substrate-api-sidecar/issues/823)

TarikGul (approved):
Overall, lgtm 👍 .

1. the Docs need to be built using `yarn build:docs`,

2. The one test needs to pass (solution written above)

TarikGul:
Nice updates 👍

alvicsam (approved):
GHA looks fine, other things I can't review :)

PR #849, opened 2022-02-22
Add benchmark checker

alvicsam:
PR adds check of the result in the `check-benchmark` job.
Currently, the average benchmark result is ~9100ms, so I suggest to set threshold on 13000ms. If this threshold exceeds then the issue will be created in the [tracker](https://github.com/paritytech/substrate-api-sidecar/issues)

Closes https://github.com/paritytech/ci_cd/issues/174


TarikGul (approved):
👍

PR #853, opened 2022-02-28
fix: support fee calculation for Bifrost

ark930:


TarikGul:
Looks like it needs a `yarn lint --fix`

ark930:
@TarikGul My mistake, already fixed.

TarikGul (approved):
> @TarikGul My mistake, already fixed.

No Worries at all :)

lgtm!

emostov (approved):

emostov:
@TarikGul  - should we have a CI job that generates `calc/pkg/calc_bg.wasm` to make sure its inline with exactly what gets merged?

TarikGul:
> @TarikGul - should we have a CI job that generates `calc/pkg/calc_bg.wasm` to make sure its inline with exactly what gets merged?

Yea I was just thinking the same thing since I always recompile it after these PR's are merged. I agree though, it would enforce consistency on the build.

PR #868, opened 2022-03-10
fix: enable experimental trace endpoints on polkadot

TarikGul:
cc:/ @emostov 

Enables the tracing endpoints for polkadot.

### DIFF
--- a/src/chains-config/polkadotControllers.ts
+++ b/src/chains-config/polkadotControllers.ts
@@ -14,6 +14,7 @@ export const polkadotControllers: ControllerConfig = {
 		'AccountsVestingInfo',
 		'Blocks',
 		'BlocksExtrinsics',
+		'BlocksTrace',
 		'NodeNetwork',
 		'NodeTransactionPool',
 		'NodeVersion',


emostov:
What is the motivation? Can we just instruct people to add this on their own?

Imod7:
I did the change locally (added the `BlocksTrace` controller for polkadot), run a polkadot node with flag `--unsafe-rpc-external` but still I cannot see the `trace` endpoints on sidecar. So I guess there is also another change that I need to do in the node level ?

TarikGul:
> I did the change locally (added the `BlocksTrace` controller for polkadot), run a polkadot node with flag `--unsafe-rpc-external` but still I cannot see the `trace` endpoints on sidecar. So I guess there is also another change that I need to do in the node level ?

I sanity checked this PR before pushing it up and the traces endpoints worked. Did you use `yarn build` to recompile the code, or are you using `yarn dev`?

Imod7:
I run a polkadot node with this command : 
`./target/release/polkadot --chain=polkadot-dev --tmp  --rpc-cors=all --alice --unsafe-rpc-external`

Then in sidecar I run
`yarn build`

TarikGul:
> 

So it should actually be 
`yarn build`
`yarn start`

emostov (approved):

0xslipk (approved):
LGTM

PR #872, opened 2022-03-15
docs: Updated README with `trace` endpoints & requirements note

Imod7:
### Description
Additions in the README related to : 
- the experimental `trace` endpoints and 
- hardware requirements of sidecar. 

The hardware requirements was a question from one of the teams we support in the integration side. So I thought it might be useful to have in the README the fact that sidecar is stateless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ NOTE: You must be a member of the `@substrate` NPM org and must belong to the `D
 Sidecar is a stateless program and thus should not use any disk space.
 
 ### Memory
-Memory wise the requirements follow the default of node.js processes which is an upper bound in HEAP memory of a little less than 2GB thus 4GB of memory should be sufficient.
+The requirements follow the default of node.js processes which is an upper bound in HEAP memory of a little less than 2GB thus 4GB of memory should be sufficient.
 
 ### Running sidecar and a node
 Please note that if you run sidecar next to a substrate node in a single machine then your system specifications should improve significantly. 
--- a/README.md
+++ b/README.md
@@ -389,4 +389,34 @@ NOTE: You must be a member of the `@substrate` NPM org and must belong to the `D
 
 ## Hardware requirements
 
-Sidecar is a stateless program and thus has no specific hardware requirements to run (disk space, memory, etc.).
+### Disk Space
+Sidecar is a stateless program and thus should not use any disk space.
+
+### Memory
+Memory wise the requirements follow the default of node.js processes which is an upper bound in HEAP memory of a little less than 2GB thus 4GB of memory should be sufficient.
+
+### Running sidecar and a node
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -373,3 +386,7 @@ NOTE: You must be a member of the `@substrate` NPM org and must belong to the `D
 #### Publish Calc Package
 
 1. `cd` into `calc/pkg` and `npm login`, then `npm publish`.
+
+## Hardware requirements
+
+Sidecar is a stateless program and thus has no specific hardware requirements to run (disk space, memory, etc.).
```
emostov, line 392: FWIW Sidecar shouldn't really use any disk space. Memory wise, I think the node.js processes default to having an upper bound heap of a little less than 2gb, so 4gb of memory should be sufficient. However many users might run sidecar next to a substrate node in a single machine -- in this case they need to significantly improved specs. We have official validator specs, but I am unsure about archive node specs. If we have recommended archive specs we should link to that. The main variable for a nodes hardware will be disk space, which depends on the chain.

README.md
```diff
@@ -93,6 +94,18 @@ node_modules/.bin/substrate-api-sidecar
 
 [Click here for full endpoint docs.](https://paritytech.github.io/substrate-api-sidecar/dist/)
 
+In the full endpoints doc, you will also find the following `trace` related endpoints : 
+- /experimental/blocks/{blockId}/traces/operations?actions=false
+- /experimental/blocks/head/traces/operations?actions=false
+- /experimental/blocks/{blockId}/traces
+- /experimental/blocks/head/traces
```
emostov, line 101:
```suggestion
- `/experimental/blocks/{blockId}/traces/operations?actions=false`
- `/experimental/blocks/head/traces/operations?actions=false`
- `/experimental/blocks/{blockId}/traces`
- `/experimental/blocks/head/traces`
```

README.md
TarikGul: It might be useful for me to aggregate the benchmarks we have currently, and post some of the results in conjunction with the hardware specs in order to give people an idea of how sidecar runs. So far all of that is internal, as I don't update the benchmarks manually anymore.

TarikGul: @Imod7 I think what you referenced above is perfect, and given the info Zeke provided. During my benchmarks of sidecar using 4 threads over 12 open http connections, I found that sidecar would use a max of 1.1GB of RSS mem. The benchmarks were overloading the cache with every runtime possible on polkadot, so this would be a max use case. The hardware in which i tested the benchmarks on were: 

```
Machine type:
n2-standard-4 (4 vCPUs, 16 GB memory)

CPU Platform:
Intel Cascade Lake

Hard-Disk:
500GB
```

TarikGul (approved):
👍

emostov (approved):

PR #890, opened 2022-04-04
fix: update polkadot-js deps, and adjust tests for getWeight

TarikGul:
Updates the following deps:
    @polkadot/api: ^7.15.1,
    @polkadot/apps-config: ^0.111.1,
    
This PR also fixes a test regarding `getWeights` in `BlocksService`. It removes the `createApiWithAugmentation` and replaces it with a set of known weights without having to generate it through injected metadata (which has been deprecated for what its worth ie `api.injectMetadata`).

### DIFF
--- a/src/services/blocks/BlocksService.spec.ts
+++ b/src/services/blocks/BlocksService.spec.ts
@@ -10,12 +10,10 @@ import LRU from 'lru-cache';
 
 import { sanitizeNumbers } from '../../sanitize/sanitizeNumbers';
 import { createCall } from '../../test-helpers/createCall';
-import { polkadotMetadataV29 } from '../../test-helpers/metadata/metadata';
 import {
 	kusamaRegistry,
 	polkadotRegistry,
 } from '../../test-helpers/registries';
-import { createApiWithAugmentations } from '../../test-helpers/typeFactory';
 import { ExtBaseWeightValue, PerClassValue } from '../../types/chains-config';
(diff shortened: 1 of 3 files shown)


src/services/blocks/BlocksService.spec.ts
```diff
@@ -309,23 +307,43 @@ describe('BlocksService', () => {
...
 			cache.reset();
 
-			const historicApi = createApiWithAugmentations(
-				polkadotMetadataV29.toHex()
-			) as ApiDecoration<'promise'>;
+			/**
+			 * This is the mockApi adjusted to mock a runtime that uses
+			 * consts.system.blockWeights for its weight nomination.
+			 */
+			const mockHistoricApiAdjusted = {
```
0xslipk, line 314: I not sure what are you trying to do here. is this a unit test or an integration test? I am confused by the casting concatenation. Why not create a mockClass of `ApiDecoration` with the data that you need? We have `jest` in place we can take advantage of it

emostov (approved):
lgtm

src/services/blocks/BlocksService.spec.ts
TarikGul: Via our conversation going to leave as is. But this is a good point to think about potentially switching over all the mock apis to a jest mockApi.

niklasad1 (approved):

PR #891, opened 2022-04-05
fix: remove blockWeightsStore

TarikGul:
Part 2 via https://github.com/paritytech/substrate-api-sidecar/pull/890

This will remove the blockWeightsStore as it is not needed anymore. `api.at` does all the decoration for us so we can retrieve the weights via the `historicApi.consts` every time when applicable. 

Todo's:
- [x] Update all the docs


src/services/blocks/BlocksService.ts
```diff
@@ -537,7 +534,7 @@ export class BlocksService extends AbstractService {
 
 		// Now that we know the exact runtime supports fee calcs, make sure we have
 		// the weights in the store
```
jsdw, line 536: Nit: I think this comment is now out of date :)

jsdw (approved):
Loving how much code was removed, and it feels simpler now :)

0xslipk (approved):
LGTM

src/services/blocks/BlocksService.spec.ts
```diff
@@ -309,23 +281,43 @@ describe('BlocksService', () => {
...
 			cache.reset();
 
-			const historicApi = createApiWithAugmentations(
-				polkadotMetadataV29.toHex()
-			) as ApiDecoration<'promise'>;
+			/**
+			 * This is the mockApi adjusted to mock a runtime that uses
+			 * consts.system.blockWeights for its weight nomination.
+			 */
+			const mockHistoricApiAdjusted = {
```
emostov, line 288: any reason for the term "Adjusted"?

emostov (approved):
nice optimization!

src/services/blocks/BlocksService.ts
TarikGul: Yea will get removed/updated. Good catch!

src/services/blocks/BlocksService.spec.ts
TarikGul: No particular reason, any suggestions are definitely welcome. I found the naming a little awkward so i stuck with adjusted

niklasad1 (approved):

PR #907, opened 2022-04-25
fix: decode `Bytes` types for `/runtime/metadata` correctly

### LINKED ISSUE #893: Decode hex values returned from /runtime/metadata
It would be useful if the value returned by the `/runtime/metadata` endpoint was decoded to decimal.

Currently, this is what is returned when querying Polkadot:

> "constants": [
            {
              "name": "ExistentialDeposit",
              "type": "6",
              "value": "0x00e40b54020000000000000000000000",
              "docs": [
                " The minimum amount required to keep an account open."
              ]
            },

The alternative to this would be to have the user convert it themselves, but being that this is SCALE encoded, they would need another library and the know-how to do so. Since the goal for Sidecar is to be an abstraction layer between PJS API and the user, I think this would be a good quality of life upgrade.

TarikGul:
closes: [#893](https://github.com/paritytech/substrate-api-sidecar/issues/893)

There are certain values given via the metadata that exist as type `Bytes` or `Raw`. Since technically the bytes can be any value, we need to find the lookup type definitions and then sanitize them accordingly. To do this I added an options param to `sanitizeNumbers` which allows us to specify we are sanitizing metadata, and to pass in a registry in order to correctly build the corresponding types these `Bytes` point to then decode them to a correct (a.) JSON value (b.) correct integer.


CurlyBracketEffect (approved):
LGTM

src/controllers/runtime/RuntimeMetadataController.ts
```diff
@@ -40,9 +40,17 @@ export default class RuntimeMetadataController extends AbstractController<Runtim
...
+			historicApi = await this.api.at(hash);
+		}
+
+		const registry = historicApi ? historicApi.registry : this.api.registry;
+
 		RuntimeMetadataController.sanitizedSend(
 			res,
-			await this.service.fetchMetadata(hash)
+			await this.service.fetchMetadata(hash),
+			{ isMetadata: true, registry: registry }
```
jsdw, line 53: Just trying to understand the significance of this; so now, we are passing in a potentially different type registry from the one that will come along with the fetched metadata being provided?

src/sanitize/sanitizeNumbers.spec.ts
```diff
@@ -1157,4 +1161,76 @@ describe('sanitizeNumbers', () => {
...
+		const sanitizeOptions = {
+			isMetadata: true,
+			registry: polkadotRegistryV9190,
+		};
+
+		it('handles unknown struct bytes correctly for u128', () => {
+			const struct = new Struct(
+				polkadotRegistryV9190,
+				{ type: 'Text', value: 'Bytes' },
+				{ type: '6', value: '0x00e87648170000000000000000000000' }
```
jsdw, line 1175: Is type 6 a sequencew of `u8`?

jsdw: ah or is it more like `[u8; 16]`?

src/sanitize/sanitizeNumbers.ts
```diff
@@ -124,47 +138,50 @@ function sanitizeCodec(value: Codec): AnyJson {
  *
  * @param data - any arbitrary data that Sidecar might send
```
jsdw, line 139: Add another `@param` for the options?

src/sanitize/sanitizeNumbers.ts
```diff
@@ -185,19 +202,71 @@ export function sanitizeNumbers(data: unknown): AnyJson {
  *
  * @param map Map | CodecMap
```
jsdw, line 203: Another @param for options?

src/sanitize/sanitizeNumbers.ts
```diff
@@ -185,19 +202,71 @@ export function sanitizeNumbers(data: unknown): AnyJson {
...
 	return jsonMap;
 }
+
+/**
+ * When v14 metadata is being sanitized, we ensure arbitrary exceptions are sanitized
+ * properly.
+ *
+ * @param key Current key of an object
+ * @param struct Current struct being sanitized
+ * @param property Current value of the inputted key
```
jsdw, line 231: another @param for registry?

src/sanitize/sanitizeNumbers.ts
```diff
@@ -23,23 +31,24 @@ import {
 	isCodec,
 	isToJSONable,
 } from '../types/polkadot-js';
+import { ISanitizeOptions } from '../types/sanitize';
 
 /**
  * Forcibly serialize all instances of AbstractInt to base 10. With Codec
  * based types we can provide a strong guarantee that the output will be of AnyJson
  *
  * @param value a type that implements polkadot-js Codec
```
jsdw, line 40: another @param for options?

src/sanitize/sanitizeNumbers.ts
```diff
@@ -49,7 +58,12 @@ function sanitizeCodec(value: Codec): AnyJson {
 			if (!property) {
 				return jsonStruct;
 			}
-			jsonStruct[key] = sanitizeNumbers(property);
+
+			jsonStruct[key] = sanitizeNumbers(property, options);
+
+			if (options?.isMetadata && options?.registry) {
+				sanitizeMetadataExceptions(key, jsonStruct, property, options.registry);
+			}
```
jsdw, line 66: I'm not currently sure what this bit does; worth a comment?

src/controllers/runtime/RuntimeMetadataController.ts
TarikGul: So for this, we actually are passing in the correct registry depending on whether the `at` query param is passed in. If there is no `at` we use the currently generated registry via `this.api`, and if `at` is used, then we retrieve the historic registry used at that block.

src/sanitize/sanitizeNumbers.spec.ts
TarikGul: So type 6 for polkadot as defined is:

```
{
  "id": "6",
    "type": {
    "path": [],
      "params": [],
        "def": {
      "primitive": "U128"
    },
    "docs": []
  }
},
```

src/sanitize/sanitizeNumbers.ts
```diff
@@ -49,7 +59,16 @@ function sanitizeCodec(value: Codec): AnyJson {
 			if (!property) {
 				return jsonStruct;
 			}
-			jsonStruct[key] = sanitizeNumbers(property);
+
+			jsonStruct[key] = sanitizeNumbers(property, options);
```
jsdw, line 63: Is it guaranteed that, by the time you're looking at the "value" key in `sanitizeMetaDataExceptions` below, the "type" key has definitely already been sanitized and placed in the `jsonStruct`?

(if `options.isMetadata`, do we know that the shape of the struct is something like `{ type, value }`? I wonder whether it makes any sense to duplicate a little more logic so that it's something like:

```
if (value instanceof Struct) {
   if (options?.isMetadata && options?.registry) {
      return sanitizeMetadataStruct(value);
   } else {
      return value.defKeys.reduce((jsonStruct, key) => { ... });
   }
}
```

Just so that all of the important logic/information for "handle this random struct" or "handle this specific type of struct" is in one place and neatly separated? (I might be misunderstanding though and this might make no sense!)

jsdw: (I guess in any case, a commentalong the lines of "we expect the metadata struct to look like `{ key: string, value: Bytes }`, and we want to convert that value into something more meaningful like a u8/u16.." would help me here because then I'd be like "ah ok, I see, we're looking for something of that shape, gotcha!)

jsdw (approved):

src/sanitize/sanitizeNumbers.spec.ts
TarikGul: Yea I can add comments, it might be a little redundant because the title of each test refers to what its looking for, but it doesnt hurt to be more explicit.

src/sanitize/sanitizeNumbers.ts
TarikGul: Based on our conversation yesterday, resolving this. I also added some inline notes for clarifications, and another test for `Option<u128>`. This is a good test because it will break if the creating of the type is ever done directly from the hex value (we need to create each type by its Uint8Array).

TarikGul:
Before we merge I am actually going to switch the structure up a bit to accommodate different metadata versions.

PR #914, opened 2022-05-06
feat(balance-info): add query param to convert free balance to human 

TarikGul:
## Feature: `denominated` queryParam for `/accounts/{accountId}/balance-info`

Summary: Returns the atomic balance as a human readable number based on the given chains token decimal. Currently the query param is a boolean so you would just pass in `?denominated=true`.


src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,9 +25,11 @@ export class AccountsBalanceInfoService extends AbstractService {
 		hash: BlockHash,
 		historicApi: ApiDecoration<'promise'>,
 		address: string,
-		token: string
+		token: string,
+		balanceToHuman: boolean
 	): Promise<IAccountBalanceInfo> {
 		const { api } = this;
+		const decimal = api.registry.chainDecimals[0];
```
TarikGul, line 32: Do we know that this value will always be there for a substrate based chain, or is it safe to say we should check and throw an error if it doesnt exist?

TarikGul:
Ill add the docs once we are happy with the naming, and what the goal of this query param should be.

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -171,4 +173,19 @@ export class AccountsBalanceInfoService extends AbstractService {
 			throw new BadRequest('Account not found');
 		}
 	}
+
+	private applyDenomination(balance: Balance, dec: number): string {
```
TarikGul, line 177: Should have a test, and check for when values are `0`

TarikGul: Also this should be applied for all Balance types.

src/types/responses/AccountBalanceInfo.ts
```diff
@@ -7,9 +7,9 @@ export interface IAccountBalanceInfo {
...
 	nonce: Index;
-	free: Balance;
-	reserved: Balance;
-	miscFrozen: Balance;
-	feeFrozen: Balance;
+	free: Balance | string;
+	reserved: Balance | string;
+	miscFrozen: Balance | string;
+	feeFrozen: Balance | string;
 	locks: Vec<BalanceLock>;
```
TarikGul, line 14: We also want to denominate the balance value in each `BalanceLock` and will need a seperate class method to parse the locks and apply the denomination accordingly

src/services/accounts/AccountsBalanceInfoService.ts
niklasad1: if a substrate base chain is using a SS58 address it will always exist AFAIU the list of addresses can be found https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json

however it's possible for a substrate based chain to use some other format.

TarikGul: > https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json

This is extremely helpful! Thanks so much. Yea based on seeing some of them having no decimal value within the array, I am going to throw an error for those who use the query param against a chain that doesnt have the value available. One of them in the list for example: subsocial.

jsdw: (I wonder why the symbol and decimals are in an array; it sortof implies that there could be more than 1, too, in some cases?)

src/services/accounts/AccountsBalanceInfoService.spec.ts
```diff
@@ -217,4 +221,64 @@ describe('AccountsBalanceInfoService', () => {
...
+
+		it('Should correctly denominate a balance when balance.length <= decimal', () => {
+			const ltValue = accountsBalanceInfoService['applyDenomination'](
+				balance,
+				7
+			);
+			const etValue = accountsBalanceInfoService['applyDenomination'](
+				balance,
+				5
+			);
```
jsdw, line 236: Just a silly edge case; what about `0` as the second arg?

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -25,9 +25,18 @@ export class AccountsBalanceInfoService extends AbstractService {
...
 	): Promise<IAccountBalanceInfo> {
 		const { api } = this;
+
+		if (withDenomination && api.registry.chainDecimals.length === 0) {
+			throw new BadRequest(
+				"Invalid use of the query parameter `denominated`. This chain doesn't have a valid chain decimal to denominate a value."
+			);
+		}
+
+		const decimal = api.registry.chainDecimals[0];
```
jsdw, line 39: Not important but I wonder whether it's worth a teeny helper function here like:

```js
function inDenomination(value) {
    withDenomination ? this.applyDenomination(value, decimal) : value
}
```

Just so that we don't need to repeat the ternary below?

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -161,14 +186,65 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+	 */
+	private applyDenomination(balance: Balance, dec: number): string {
+		const strBalance = balance.toString();
+
+		// We dont want to denominate a zero balance
+		if (strBalance === '0') {
+			return strBalance;
+		}
+		// If the denominated value will be less then zero, pad it correctly
+		if (strBalance.length <= dec) {
```
jsdw, line 220: I'm struggling to parse exactly how thse logic for this is working, but I can see that it's being tested so it's not a big deal, just curious really!

jsdw (approved):

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: Well the idea was how can we properly denominate the decimal places without ever having the string (number represented as a string) be converted to a number. This is because the atomic value given can be too large for javascript to interpret. So in the logic above I wanted to make sure we also never have to BN, or convert to a bigInt. We can easily fix the value to a decimal with some simple math, but because the values might be to big thats not possible.

So basically what the logic does takes care of three cases.

1. If the value is zero just return zero

2. If the decimal value is larger than the length of the atomic value (it's a string), the value denominated is going to be less than zero. Which means it needs the correct amount of padding (zeros) after the decimal. 

3. The value is greater than zero, and therefore handle the decimal places by subtracting the length of the value, by the decimal places, and using the value to determine where we put the decimal.

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: Yea I agree this is cleaner.

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: Yea you are correct. After doing some research though, the order of the values in the array correspond to the values in the tokens array: So... If the available tokens in the chain are `['MOVR', 'GMBR']` -> then the chainsDecimals array would look like `[18, 18]`. I haven't seen a single case where the decimals change, BUT doesnt mean it cant. I should probably check for this.

src/services/accounts/AccountsBalanceInfoService.spec.ts
TarikGul: Added the test in and fixed the logic a bit

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: fixed

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -161,14 +174,89 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+	 *
+	 * @param balance free balance available encoded as Balance
+	 * @param dec The chains given decimal value
+	 */
+	private applyDenominationBalance(balance: Balance, dec: number): string {
+		const strBalance = balance.toString();
+
+		// We dont want to denominate a zero balance or zero decimal
+		if (strBalance === '0' || dec === 0) {
+			return strBalance;
```
niklasad1, line 199: hmm,

this means that `0`will be => `0` and not `0.00` correct?
perhaps it would be better to add the decimals for zero to for consistency

src/services/accounts/AccountsBalanceInfoService.ts
niklasad1: > If the decimal value is larger than the length of the atomic value (it's a string), the value denominated is going to be less than zero. Which means it needs the correct amount of padding (zeros) after the decimal.

I don't follow that but maybe I'm just slow such you might have a number like `123.9999` (decimal is bigger than the number) why do you need to pad it?

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -161,14 +174,89 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+				feeFrozen: this.inDenominationBal(denominate, feeFrozen, decimal),
+				locks: this.inDenominationLocks(denominate, locks, decimal),
 			};
 		} else {
 			throw new BadRequest('Account not found');
 		}
 	}
+
+	/**
+	 * Apply a denomination to a balance depending on the chains decimal value.
```
niklasad1, line 189: so I don't really understand why we can't just do `balance.toString() || "." || dec.toString()`?

src/services/accounts/AccountsBalanceInfoService.ts
jsdw: I also had non-integers in mind, but makes sense to me after both of your comments; thanks!

TarikGul: > ah, right the decimals is just some number such as 10 that the number should be multiplied with not an actual decimal point. Sorry I got confused by dec.

Yup exactly!

Important to note as well, that the balance passed in will always be represented as an atomic balance via the `Balance` type.

src/services/accounts/AccountsBalanceInfoService.ts
```diff
@@ -161,14 +174,89 @@ export class AccountsBalanceInfoService extends AbstractService {
...
+	 *
+	 * @param balance free balance available encoded as Balance
+	 * @param dec The chains given decimal value
```
niklasad1, line 192: If you could explain that `dec` is decimal tokens in (ss58 context) which is something that applied to `base unit` then I'm good with this.

AFAIU the `decimal tokens` must > 0

niklasad1 (approved):

src/services/accounts/AccountsBalanceInfoService.ts
TarikGul: added

PR #926, opened 2022-05-23
feat: Added the `network` and `accountId` to the response of `validate` endpoint

Imod7:
## Description
Added the `network` and `accountId` in the response from the `/accounts/:address/validate`endpoint when the given address `isValid`. If it is an invalid address (`isValid: false`) then `network` and `accountId` have `null` values. 

## Functionality of the endpoint
Below you can find a description of how this endpoint behaves and why.

This endpoint is meant to receive only valid `addresses` and these can be in the following formats :
1. ss58 format
    - example address from [tests](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/accounts/AccountsValidateService.spec.ts) 1xN1Q5eKQmS5AzASdjt6R6sHF76611vKR4PFpFjy1kXau4m
2. hex (from a u8 array) format (of valid length)
    - example address from [tests](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/accounts/AccountsValidateService.spec.ts) 0x002a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc592312
3. for anything else, this endpoint will/should return `null` which means `isValid` field will be `false` and all other fields will be `null`. 

### Comparing with `subkey` functionality
If we compare this endpoint's functionality with how [subkey](https://github.com/paritytech/substrate/tree/master/bin/utils/subkey) works then we have the following behaviours :
1. For the first format : `subkey` and `validate` endpoint will return a result : 
    -  hence will consider this address valid and 
    - the `AccountId` returned will be the same as the one returned from the `validate` endpoint
3. For the second format : 
     - `subkey` : 
	   1. considers this address invalid
	   2. but if we correct manually the address and we remove prefix+checksum then it will return a result and the expected `AccountId` 
     - `validate` endpoint :
	   1. considers this address valid. Why ? because it is possible that a user stores their keys in this format.
	   2. then strips the prefix+checksum and then returns the correct `AccountId`

### Important Note
If we give as input the address (stripped from prefix+checksum hence `AccountId`) : 
- to subkey : it will return a valid result
	- why ? because accepts addresses + account ids.
- to `validate` endpoint : it will return `null`
	- why? because this is an `AccountId` and this endpoint is meant to validate only addresses. For account ids we intend to create another endpoint.


src/services/accounts/AccountsValidateService.ts
```diff
@@ -26,16 +30,21 @@ export class AccountsValidateService extends AbstractService {
 
 		if (defaults.allowedEncodedLengths.includes(u8Address.length)) {
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
+			const networkName = allNetworks.filter(
```
TarikGul, line 33: Yea for loops will inherently be more performant since no callbacks are in the call stack. But that being said I don't think it's something that will be noticeable here. But is always nice to have. I will leave this up to you, with what you think is cleaner.

joepetrowski:
> For the moonbeam address I will need to look into that, and play with it. Not exactly sure whats happening there.

For Ethereum style addresses they are not SS58 formatted, they are just 20 raw bytes. It should be a variant of `enum MultiAddress` from Substrate.

TarikGul:
> For Ethereum style addresses they are not SS58 formatted, they are just 20 raw bytes. It should be a variant of enum MultiAddress from Substrate.

Ahh okay cool, I assumed they were leveraging ethereums aaddress format. Thanks for the info.

src/types/responses/ValidateAddress.ts
```diff
@@ -1,4 +1,5 @@
 export interface IValidateAddrResponse {
 	isValid: boolean;
 	ss58Prefix: number | null;
+	networkName: string | null;
```
joepetrowski, line 4: I would also add `AccountId` that would return the account stripped of any formatting. In the case of the Relay Chain, it's just a 32 byte value.

Perhaps it is a separate endpoint, but there could be a prefix converter that would take args like `?public=true` for giving an `AccountId` as input and `?prefix=<int>` for defining the prefix.

TarikGul:
As a note as well, when its out of [WIP] to change the squashed commit name (title of the PR) to a conventional commit structure.

src/services/accounts/AccountsValidateService.ts
```diff
@@ -26,16 +32,28 @@ export class AccountsValidateService extends AbstractService {
...
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
-
+			const networkName = allNetworks.filter(
+				(network) => network.prefix === ss58Prefix
+			)[0].network;
+			if (isHex(address)) {
+				accountId = address;
+			} else {
+				const registry = new TypeRegistry();
+				accountId = registry.createType('AccountId', address).toHex();
```
Imod7, line 42: @TarikGul Not sure if this is the way to retrieve the `accountId` but still I pushed the change to get some feedback. 
The logic is described in the last commit message and I added below also : 
- If the address given is a hex value then it returns the same as the `accountId`
- If the address given is not a hex value then it returns the hex value of the related registry type

src/types/responses/ValidateAddress.ts
Imod7: @joepetrowski @TarikGul So in practice that would be something like this ?
- If the given address is a substrate address like 
  ```
  5Gv8YYFu8H1btvmrJy9FjjAWfb99wrhV3uhPFoNEr918utyR
  ``` 
  then the request would look like this : 
  ```
  http://127.0.0.1:8080/accounts/5Gv8YYFu8H1btvmrJy9FjjAWfb99wrhV3uhPFoNEr918utyR/validate
  ```
  and the response will look like :  
  ```
  isValid : true
  ss58Prefix : "42"
  networkName : "substrate"
  accountId : "0x2ad6a3105d6768e956e9e5d41050ac29843f98561410d3a47f9dd5b3b227ab87"
  ```
  and that would mean that args `public=false` 

- If what is given is a substrate `accountId` like 
  ```
  0x2ad6a3105d6768e956e9e5d41050ac29843f98561410d3a47f9dd5b3b227ab87
  ``` 
  then the request would look like : 
  ```
  http://127.0.0.1:8080/accounts/0x2ad6a3105d6768e956e9e5d41050ac29843f98561410d3a47f9dd5b3b227ab87?public=true&prefix=42/validate
  ```
  and the response will look like :  
  ```
  isValid : true
  ss58Prefix : "42"
  networkName : "substrate"
  address : "5Gv8YYFu8H1btvmrJy9FjjAWfb99wrhV3uhPFoNEr918utyR"
  ```
 
My apologies if i misunderstood the requirement and thank you so much for any feedback! :))

Imod7 (commented):
- Comments on the added `accountId` in the response of the endpoint and 
- Thoughts on the implementation related to the conversion of the `accountId`

src/types/responses/ValidateAddress.ts
joepetrowski: Yes, exactly! I'd also add in conversion, so:

```
http://127.0.0.1:8080/accounts/5Gv8YYFu8H1btvmrJy9FjjAWfb99wrhV3uhPFoNEr918utyR/validate?prefix=0
```

Would return:

```
isValid : true
ss58Prefix : "0"
networkName : "polkadot"
address : "15rRgsWxz4H5LTnNGcCFsszfXD8oeAFd8QRsR6MbQE2f6XFF"
accountId : "0xd6a3105d6768e956e9e5d41050ac29843f98561410d3a47f9dd5b3b227ab8746"
```

Also starting to realize that this might be getting more complicated than expected. To do this, you also need a `?scheme` param that takes `{sr25519, ed25519, ecdsa}`, and `if ecdsa && public` then the account ID is the hash of the public key.

src/services/accounts/AccountsValidateService.ts
```diff
@@ -26,16 +32,28 @@ export class AccountsValidateService extends AbstractService {
...
 		if (defaults.allowedEncodedLengths.includes(u8Address.length)) {
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
-
+			const networkName = allNetworks.filter(
+				(network) => network.prefix === ss58Prefix
+			)[0].network;
+			if (isHex(address)) {
+				accountId = address;
+			} else {
+				const registry = new TypeRegistry();
```
TarikGul, line 41: To handle the registry correctly we can leverage the decorated api. See [ValidateAddressController](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/controllers/accounts/AccountsValidateController.ts#L9). The controller passes an api down to the service. Therefore you can use `this.api.registry`. The reason we dont want to use `new TypeRegistry` is because it doesn't have any metadata attached to it specific to the connected chain. Instead it probably has static substrate metadata.

src/services/accounts/AccountsValidateService.ts
TarikGul: See above comment on how we can properly generate the accountId with the correct registry. 

To address the above logic though. Seems fine to me 👍

src/types/responses/ValidateAddress.ts
TarikGul: @joepetrowski I think this should be a separate endpoint so we don't couple the logic between validation and conversion. My thought would be:

`/accounts/{accountId}/convert`

It would take in 3 query params: `scheme` = sr25519, `prefix` = 42, `public` = true. Each value to the right of the equal sign would be the defaults. This way it's clear to users what the REST endpoint is providing, and would also be easier to maintain.

joepetrowski: Yeah that's reasonable.

src/services/accounts/AccountsValidateService.ts
```diff
@@ -26,16 +34,31 @@ export class AccountsValidateService extends AbstractService {
...
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
-
+			network = null;
+			for (const networkParams of allNetworks) {
+				if (networkParams['prefix'] === ss58Prefix) {
+					network = networkParams['network'];
+					break;
+				}
+			}
+			if (isHex(address)) {
```
TarikGul, line 44:
```
account = isHex(address) ? address : this.api.registry.createType('AccountId', address).toHex();
```

docs/src/openapi-v1.yaml
```diff
@@ -1451,10 +1452,16 @@ components:
           type: boolean
           description: Whether the given address is valid ss58 formatted.
         ss58Prefix:
-          type: string
+          type: number
```
TarikGul, line 1455:
```suggestion
          type: string
```

So this is actually a string, since the sanitization process will serialize it as a string. 
That being said, you can add a `format` field beneath the description with `unsignedInteger` to denote it will be a number.

TarikGul (approved):
Few nits to resolve, but overall great job! 👍 

You will also need to run `yarn build:docs` to regenerate the docs.

docs/src/openapi-v1.yaml
```diff
@@ -344,15 +344,16 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  /accounts/{accountId}/validate:
+  /accounts/{address}/validate:
```
TarikGul, line 347: Good catch on this btw.

joepetrowski (commented):
For the `Should give the correct response for a {network} hex value` tests, I don't understand how the sanitization should know to which network the hex belongs? Wouldn't you have to pass it a network name or prefix as an arg? The hex value should generally speaking be the same across most networks (exceptions being those that don't use `AccountId32: [u8; 32]` as their `AccountId` type).

TarikGul:
@joepetrowski Our process for converting a hex address goes as such:

1. convert it to a u8a
2. If the encodedlLength satisfies one of the acceptable [lengths](https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/defaults.ts#L9) then we can check the addresses checksum.
3. We then specifically use [checkAddresssChecksum](https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/checksum.ts#L6) which will return the ss58Prefix for us. 
4. Using the given information we can then match that ss58Prefix with its known network.

src/services/accounts/AccountsValidateService.ts
```diff
@@ -26,16 +34,29 @@ export class AccountsValidateService extends AbstractService {
...
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
-
+			network = null;
+			for (const networkParams of allNetworks) {
+				if (networkParams['prefix'] === ss58Prefix) {
+					network = networkParams['network'];
+					break;
+				}
+			}
+			accountId = isHex(address)
```
TarikGul, line 44: No need for the `let accoundId` either anymore, we can save this directly to a const

src/services/accounts/AccountsValidateService.spec.ts
```diff
@@ -75,6 +90,9 @@ describe('Validate addresses', () => {
 		const expectedResponse = {
 			isValid: true,
 			ss58Prefix: '0',
+			network: 'polkadot',
+			accountId:
+				'0x002a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc592312',
```
joepetrowski, line 95: OK, so I looked into it more, and all of these are incorrect. What you have here is `(prefix, AccountId, checksum)`, not `AccountId`. This breaks down to:

- Prefix: `00`
- AccountId: `2a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc59`
- Checksum: `2312`

For example, in subkey:

```bash
> subkey inspect --public 0x002a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc592312
Error: Input("Failed to construct public key from given hex")
```

But strip away the prefix and checksum:

```bash
> subkey inspect --public 0x2a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc59 -n polkadot
Network ID/version: polkadot
  Public key (hex):   0x2a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc59
  Account ID:         0x2a39366f6620a6c2e2fed5990a3d419e6a19dd127fc7a50b515cf17e2dc5cc59
  Public key (SS58):  1xN1Q5eKQmS5AzASdjt6R6sHF76611vKR4PFpFjy1kXau4m
  SS58 Address:       1xN1Q5eKQmS5AzASdjt6R6sHF76611vKR4PFpFjy1kXau4m
```

joepetrowski: This also aligns with the `AccountId`s in the previous tests.

Imod7: Thank you @joepetrowski for the super valuable feedback!

I changed the code so that if the user calls this endpoint with an address of : 
- valid length and
- in hex (from a u8 array) format
now I print an account id stripped from the prefix + checksum. 

Since I strip characters from specific indexes, I was wondering if there is any case where the user can input an address of another hex format of valid length ? If yes and if this format does not have the prefix + checksum in the same indexes then this solution might not work.

joepetrowski: Abstractly, yes, it's possible. If you look at [MultiAddress](https://github.com/paritytech/substrate/blob/6a59cc3/primitives/runtime/src/multiaddress.rs#L26) it does support `Raw(Vec<u8>),`. Practically speaking, most `AccountId` stuff uses this [32 byte](https://github.com/paritytech/substrate/blob/6a59cc3/primitives/core/src/crypto.rs#L483) object and converts [pub keys into it](https://github.com/paritytech/substrate/blob/6a59cc3/primitives/runtime/src/lib.rs#L328L335).

TarikGul (approved):
Great updates, and changes. Good catch @joepetrowski .

Just seems like there is a conflict with the docs. 
```
git pull origin master
yarn && yarn build:docs
```

That should sort out the merge conflict

Imod7:
Thank you @TarikGul and @joepetrowski for all the comments and reviews on this PR. Very much appreciated!!!

src/services/accounts/AccountsValidateService.spec.ts
```diff
@@ -51,6 +60,9 @@ describe('Validate addresses', () => {
 		const expectedResponse = {
 			isValid: true,
 			ss58Prefix: '16',
+			network: 'kulupu',
```
jsdw, line 63: Just wondering how many third party chains we support here; any which has claimed an ss58 prefix?

src/services/accounts/AccountsValidateService.ts
```diff
@@ -42,16 +49,29 @@ export class AccountsValidateService extends AbstractService {
...
+			network = null;
+			for (const networkParams of allNetworks) {
+				if (networkParams['prefix'] === ss58Prefix) {
+					network = networkParams['network'];
+					break;
+				}
+			}
+			const accountId = isHex(address)
+				? '0x' + address.slice(4, -4)
+				: this.api.registry.createType('AccountId', address).toHex();
```
jsdw, line 61: 1. Should the result of `isHex(address)` (which I assume has to loop over and check each character?) be cached at the top of the function to avoid calling it again here)?

2. Just trying to wrap my head around what's happening. So, if I provide arbitrary hex that happens to be one of the allowed lengths, `network` would still be `null`, `isValid` would be false, and the accountId would be some arbitrary bytes from my hex?

I guess since we return `isValid`, this is all good, but it makes me wonder whether the result type should really be either `{ isValid: true, ss58Prefix: 12, network: 'foo', accountId: '0x1234'}` or `{ isValid: false, ss58Prefix: null, network: null, accountId: null }`? Since if it's not valid, the other details make no sense?

src/services/accounts/AccountsValidateService.spec.ts
TarikGul: Yup exactly, all the chains that have an ss58 prefix, but I also believe it's constrained to what `allNetworks` has logged in polkadot-js.

src/services/accounts/AccountsValidateService.ts
Imod7: Thank you for the feedback @jsdw !
1. I am not sure if we need to cache the result of `isHex(address)` since from what I see [here](https://github.com/polkadot-js/common/blob/3597a4eb4e0cf9985104568058c5cd23701bbaed/packages/util/src/is/hex.ts#L25) it does not loop over every character. But if needed I will move it at the top.
2. You are right, it is more clear and correct to give all `null` values when `isValid` is `false`. I just changed it in my last commits. Very good point!

jsdw: 1. Ah I see; I guess the current `isHex` would see something like `0xZZZZ` as valid hex, then! I'd probably cache the result to avoid the extra call to it but it's not a big deal given the current implementation!

src/services/accounts/AccountsValidateService.ts
```diff
@@ -15,21 +15,29 @@
...
+	 * a hex (from a u8 array) formatted address,
+	 * what the ss58 prefix for that address is,
+	 * what is the network address format,
+	 * and what the account ID is for this address.
 	 *
 	 * @param address ss58 or hex address to validate
 	 */
 	validateAddress(address: string): IValidateAddrResponse {
 		let u8Address;
+		let network;
```
jsdw, line 40: No need to have `let network` at the top level by the looks of it; just have `let network = null` on line 54, so it's more obvious at a glance where the variable is used :)

src/services/accounts/AccountsValidateService.ts
```diff
@@ -42,16 +50,31 @@ export class AccountsValidateService extends AbstractService {
 
 		if (defaults.allowedEncodedLengths.includes(u8Address.length)) {
 			const [isValid, , , ss58Prefix] = checkAddressChecksum(u8Address);
-
-			return {
-				isValid: isValid,
-				ss58Prefix,
-			};
+			if (isValid == true) {
```
jsdw, line 53: Nit:

```suggestion
			if (isValid) {
```

jsdw (approved):
Couple of wee comments but LGTM!

lexnv (approved):

TarikGul (approved):
Awesome LGTM 👍  Thank you!

src/services/accounts/AccountsValidateService.ts
Imod7: done!

src/services/accounts/AccountsValidateService.ts
Imod7: done! thank you!

src/services/accounts/AccountsValidateService.ts
Imod7: oh! Now I think I understand what you meant, to avoid calling twice the `isHex`. Yes, ofc, very good point. So I cache it in L40 and then I just check what boolean value the `addressIsHex` has.

Issue #925, opened 2022-05-20
[Feature] Add Tip Estimation Functionality for Transaction Priority


IkerAlus:
**Proposed Feature**

It would be nice to have a tip estimation feature available to increase the transaction priority prior submitting a transaction. This way, API-Sidecar users would have a guidance if they need fast transaction inclusion or they are ok with slow transaction inclusion.

In fact , users have already asked around [this functionality](https://substrate.stackexchange.com/questions/2716/is-there-a-way-to-find-out-the-appropriate-fee-when-generate-tx).

**Additional Information**

In general, this would imply to fetch up to n past blocks from the node and compute a tip estimation for a given transaction as a function of past transaction tips, past block weight usage and weight of the transaction to be send. There is extensive work done in this direction for Bitcoin and Ethereum and most of the wallets for those networks offer this functionality.

**Alternatives**

If we think that tip estimation is out of the scope of API-Sidecar, at least it would be nice to have an endpoint to fetch n blocks from the node (As it was already proposed [here](https://github.com/paritytech/substrate-api-sidecar/issues/460)).

TarikGul:
Okay, so I definitely like this and I think it holds a lot of merit. On the side of computation, this could be added to calc fee and I can write it in rust to save speed in sidecar. 

As to querying the past n blocks in regards to tip estimation, id be interested to see how we would want to implement this since it would take up some resources. That being said though we would only need to parse the extrinsics of each block and nothing else, and maybe cache any particular information that would be useful for future query's when it comes to tips. Curious how we would do that analysis, and happy to put together a call to discuss the design of something like this.  

> If we think that tip estimation is out of the scope of API-Sidecar, at least it would be nice to have an endpoint to fetch n blocks from the node (As it was already proposed https://github.com/paritytech/substrate-api-sidecar/issues/460).

I am happy to implement this very soon, as its on the roadmap already, and I think it can be done fairly quickly.

vae520283995:
Hope this feature will be implemented soon

IkerAlus:
>  That being said though we would only need to parse the extrinsics of each block and nothing else...

Actually, considering a similar approach to the typical fee estimator implemented [for Bitcoin core](https://blog.iany.me/2020/08/bitcoin-core-fee-estimate-algorithm/), the application would need to track also the (node's) transaction pool and carry several estimated values to be updated with every new block. This would imply to do automated queries to the node every 5-10 seconds and persist some kind of local store for the estimations...

[...]

PR #931, opened 2022-05-25
feat: add query param to add a `tip`, `partialFee`, and `priority` field to each extrinsic for `/node/transaction-pool`

TarikGul:
## Feature

**Summary**: Allow `/node/transaction-pool` to accept a query parameter called `includeFee`. When true it will attach a decoded `tip`, `partialFee`, and `priority` to each extrinsic. 

### Query Param `includeFee`

- Default: It will be `false`.
- If true: It will attach a `tip`, `partialFee`, and `priority` to each extrinsic within the "pool" array. 

rel: https://github.com/paritytech/substrate-api-sidecar/issues/925


src/controllers/node/NodeTransactionPoolController.ts
```diff
@@ -29,12 +30,14 @@ export default class NodeTransactionPoolController extends AbstractController<No
 	 * @param res Express Response
```
CurlyBracketEffect, line 30: Just a little note to adjust the comment describing the function params.

CurlyBracketEffect (approved):
Logic looks good...Just the note about the comment.

src/controllers/node/NodeTransactionPoolController.ts
CurlyBracketEffect: This comment was supposed to reference line 28

TarikGul: Good catch.

lexnv (approved):

docs/src/openapi-v1.yaml
```diff
@@ -2765,6 +2773,10 @@ components:
                   type: string
                   format: hex
                   description: Scale encoded extrinsic.
+                tip:
+                  type: string
+                  format: number
```
TarikGul, line 2778:
```suggestion
                  format: unsignedInteger
```

src/services/node/NodeTransactionPoolService.ts
```diff
@@ -3,20 +3,32 @@ import { INodeTransactionPool } from 'src/types/responses';
...
-	async fetchTransactionPool(): Promise<INodeTransactionPool> {
+	async fetchTransactionPool(
+		includeTip: boolean
+	): Promise<INodeTransactionPool> {
 		const { api } = this;
 
 		const extrinsics = await api.rpc.author.pendingExtrinsics();
 
-		const pool = extrinsics.map((ext) => {
+		if (includeTip) {
```
jsdw, line 13: Wondering whether you can avoid the duplication and get away with something like:

```ts
return {
    pool: extrinsics.map((ext) => {
        return {
            hash: ext.hash.toHex(),
            encodedExtrinsic: ext.toHex(),
            tip: includeTip ? ext.tip.toString() : undefined,
        };
    }),
};
```

(afaik undefined fields disappear when serialized to JSON anyway)

jsdw (approved):

jsdw:
Just to relay @joepetrowski's comment for consideration here:

Wouldn't you want to return the total fee including tip?
Since that's what determines priority. So if a transfer has a fee of 0.1, and a remark has a fee of 0.001, and the pool you query is loaded with transfers with a tip of 0, you could then think, "I'll put a tip of 0.01 just to make sure I'm ahead of all the stingy non-tippers", but with a total fee of 0.011 you'll still be last in line.

TarikGul:
@jsdw Yea I am going to rewrite this a bit to include the patialFee, and totalFee. Will request reviews again after.

src/services/node/NodeTransactionPoolService.ts
```diff
@@ -14,24 +14,103 @@
...
+					'This runtime does not have the sufficient materials to calculate fees.'
+				);
+			}
+
+			return {
+				pool: extrinsics.map((ext) => {
+					const { hash, encodedLength, tip } = ext;
+					const partialFee = calcFeeWithParams.calc_fee(
+						// TODO This value comes from the success, failure events.
+						BigInt(0),
```
TarikGul, line 47: This is the last piece of this calculation to figure out. Originally this value comes from the weight provided by a failed or successful event.

TarikGul: Might make more sense to use `api.rpc.payment.queryInfo` to find fee estimate

TarikGul: Okay so this is the right move, I am going to remove all `@substrate/calc` logic here and replace it with `queryInfo`.

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -28,8 +28,10 @@ import {
 	SessionIndex,
 	StakingLedger,
 } from '@polkadot/types/interfaces';
+import BN from 'bn.js';
 
 import { polkadotMetadata } from '../../../test-helpers/metadata/metadata';
+// import { TypeFactory } from '../../../test-helpers/typeFactory';
```
niklasad1, line 34:
```suggestion
```

niklasad1, line 34:
```suggestion
```

src/services/node/NodeTransactionPoolService.ts
```diff
@@ -14,24 +14,172 @@
...
+		ext: Extrinsic,
+		dispatchClass: DispatchClass,
+		weight: Weight
+	): Promise<string> {
+		const { api } = this;
+		const { tip, encodedLength: len } = ext;
+		const BN_ONE = new BN(1);
+		const sanitizedClass = this.defineDispatchClassType(dispatchClass);
+
+		const { maxBlock } = this.api.consts.system.blockWeights;
```
niklasad1, line 106:
```suggestion
		const { maxBlockWeight } = this.api.consts.system.blockWeights;
```

niklasad1, line 106:
```suggestion
		const { maxBlockWeight } = this.api.consts.system.blockWeights;
```

src/services/node/NodeTransactionPoolService.ts
```diff
@@ -14,24 +14,172 @@
...
+		const maxLength: u32 =
+			this.api.consts.system.blockLength.max[sanitizedClass];
+		const boundedWeight = BN.min(BN.max(weight.toBn(), BN_ONE), maxBlock);
+		const boundedLength = BN.min(BN.max(new BN(len), BN_ONE), maxLength);
+		const maxTxPerBlockWeight = maxBlock.div(boundedWeight);
+		const maxTxPerBlockLength = maxLength.div(boundedLength);
+
+		const maxTxPerBlock = BN.min(maxTxPerBlockWeight, maxTxPerBlockLength);
+		const saturatedTip = tip.toBn().add(BN_ONE);
+		const scaledTip = this.maxReward(saturatedTip, maxTxPerBlock);
```
niklasad1, line 116: nice, totally missed that that was a closure in substrate

niklasad1, line 116: nice, totally missed that that was a closure in substrate

niklasad1 (approved):
gucci :)

src/services/node/NodeTransactionPoolService.ts
TarikGul: So we can't actually rename this using this syntax as it's destructuring the key `maxBlock` from the object `blockWeights`, but I do agree that some explicit naming would be better here so I'll give it a change.

TarikGul: changed

PR #944, opened 2022-06-06
fix(scripts): jest pass-fail bug

TarikGul:
Currently there was an issue when running the e2e tests where it wouldn't properly catch whether or not a test failed. It would always assume it passed and resolve a success code. This PR fixes that and allows the script to properly handle the error code on a failed test.

### DIFF
--- a/scripts/sidecarScriptApi.ts
+++ b/scripts/sidecarScriptApi.ts
@@ -83,12 +83,12 @@ export const killAll = (procs: ProcsType): void => {
 export const launchProcess = (
 	cmd: string,
 	procs: ProcsType,
-	{ proc, resolver, resolverStartupErr, args }: IProcOpts
+	{ proc, resolver, resolverJestErr, resolverStartupErr, args }: IProcOpts
 ): Promise<StatusCode> => {
 	return new Promise<StatusCode>((resolve, reject) => {
 		const { Success, Failed } = StatusCode;
-
 		const command = cmd || 'yarn';
+		let jestStatus = Success;
 
 		procs[proc] = spawn(command, args, { detached: true });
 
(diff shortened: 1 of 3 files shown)


scripts/sidecarScriptApi.ts
```diff
@@ -103,12 +103,17 @@ export const launchProcess = (
 		procs[proc].stderr.on('data', (data: Buffer) => {
 			console.error(data.toString().trim());
 
+			if (resolverJestErr && data.toString().trim().includes(resolverJestErr)) {
+				jestStatus = Failed;
+			}
+
 			if (resolverStartupErr && data.toString().includes(resolverStartupErr)) {
```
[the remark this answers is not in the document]
TarikGul: Good catch.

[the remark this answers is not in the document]
TarikGul: fixed

scripts/sidecarScriptApi.ts
```diff
@@ -103,12 +103,17 @@ export const launchProcess = (
...
 			console.error(data.toString().trim());
 
-			if (resolverStartupErr && data.toString().includes(resolverStartupErr)) {
+			if (resolverJestErr && data.toString().trim().includes(resolverJestErr)) {
+				jestStatus = Failed;
+			}
+
+			if (resolverStartupErr && data.toString().trim().includes(resolverStartupErr)) {
 				resolve(Failed);
 			}
```
jsdw, line 112: Just want to understand this bit; if there's a startup error and `resolverStartupErr` is true then we bail immedaitely with Failed, but in the `resolverJestErr` case we just set it to failed and let the "close" handle it?

jsdw, line 112: Just want to understand this bit; if there's a startup error and `resolverStartupErr` is true then we bail immedaitely with Failed, but in the `resolverJestErr` case we just set it to failed and let the "close" handle it?

jsdw (approved):

scripts/sidecarScriptApi.ts
jsdw: (I guess I'm wondering why we choose to immediately resolve in some cases, and wait for close in others?)

TarikGul: Yea the difference here being when we know Jest fails, we want it to finish resolving the suite, but acknowledge it's failure. Whereas if there is a startup error, we just resolve the process. This makes sure the Jest finishes doing whatever it needs before we log its failed tests.

jsdw: Ok gotcha, makes sense!

niklasad1 (approved):

PR #954, opened 2022-06-09
feat: add `/blocks` that enforces `range` query param.

### LINKED ISSUE #460: /blocks should allow fetching of multiple blocks
Right now the `blocks` endpoint (somewhat misleadingly) fetches info for only a single block. It would be good if it supported a comma or plus separated list of blocks, and return them all in an array.

```
/blocks/500001+500002
or
/blocks/500001,500002
```

The BC-approach would be to only return as array when multiple blocks are passed in. The *right* approach would be a mid version bump due to a change in return values which should now ALWAYS return arrays, even when looking for a simple blocks. The single-element return should be renamed to `/block` and should under the hood just call `/blocks/xxxx[0]`

TarikGul:
### Feature

Endpoint:  `/blocks`

Query params: `range` (required) ex. `/blocks?range=0-999`
  - I implemented our own range parser because it's not perfectly clear yet if we want to provide support only for a static range such as `0-999`, or for something more dynamic such as `1,2,3,10-100,123,1234`
  
Response: 
```
[
{<IBlock>},
{<IBlock>},
...
]
```

co-authored by @jsdw (Implemented the PromiseQueue)

closes: [#460](https://github.com/paritytech/substrate-api-sidecar/issues/460)


TarikGul:
Some benchmarks for querying 1000 blocks at a time against a polkadot archive node:

```
GET /blocks?range=0-1000 200 4242ms
GET /blocks?range=10000-11000 200 4074ms
GET /blocks?range=100000-101000 200 4578ms
GET /blocks?range=1546264-1547264 200 7568ms
GET /blocks?range=2416152-2417152 200 5778ms
GET /blocks?range=3000000-3001000 200 6084ms
```

src/controllers/AbstractController.ts
```diff
@@ -196,6 +217,10 @@ export default abstract class AbstractController<T extends AbstractService> {
 		);
 	}
 
+	private verifyInt(num: Number): boolean {
+		return !Number.isInteger(num) || num < 0;
+	}
```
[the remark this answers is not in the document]
TarikGul: ahh yes, good call.

src/controllers/AbstractController.ts
```diff
@@ -180,13 +182,32 @@ export default abstract class AbstractController<T extends AbstractService> {
...
+	protected parseRangeOfNumbersOrThrow(n: string): number[] {
+		const splitRange = n.split('-');
+		if (splitRange.length !== 2) {
+			throw new BadRequest('Incorrect range format. Expected example: 0-999');
+		}
+
+		const min = Number(splitRange[0]);
+		const max = Number(splitRange[1]);
+
+		if (!this.verifyInt(min) || !this.verifyInt(max)) {
```
niklasad1, line 204: I find this logic a little bit hard to understand.

As it will accept both `0-100` and `100-0`, I kind assumed that that "reverse range" should not work.

src/controllers/AbstractController.ts
```diff
@@ -180,13 +182,32 @@ export default abstract class AbstractController<T extends AbstractService> {
...
+			throw new BadRequest('Inputted range contains non integers.');
+		}
+
+		return [...Array(max - min + 1).keys()].map((i) => i + min);
```
niklasad1, line 208: I think you should check that `max >= min`

src/controllers/AbstractController.ts
niklasad1: i.e, 

```
if !(this.verifyUInt(min) {
  throw new BadRequest('Inputted `min` invalid UInt.');
) 

if !(this.verifyNonZeroUInt(min) {
  throw new BadRequest('Inputted `max` invalid NonZeroUInt.');
) 

if (min >= max ) {
   throw new BadRequest('Invalid range: max must be bigger than min');
}
```

src/controllers/AbstractControllers.spec.ts
```diff
@@ -316,4 +316,34 @@ describe('AbstractController', () => {
...
+			const badFormatRequest = new BadRequest(
+				'Incorrect range format. Expected example: 0-999'
+			);
+			const badIntRequest = new BadRequest(
+				'Inputted range contains non integers.'
+			);
+			expect(() => controller['parseRangeOfNumbersOrThrow']('100')).toThrow(
+				badFormatRequest
+			);
+			expect(() => controller['parseRangeOfNumbersOrThrow']('100-h')).toThrow(
```
niklasad1, line 337: perhaps a test for `100-0` and `1-1`

src/controllers/blocks/BlocksController.ts
```diff
@@ -233,4 +236,71 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+			queryFinalizedHead,
+			omitFinalizedTag,
+		};
+
+		/**
+		 * Given a range of hashes, generate an array containing callbacks for
+		 * each query.
+		 */
+		const tasks: Array<() => Promise<IBlock>> = [];
+		for (let i = 0; i < rangeOfNumsToHash.length; i++) {
```
niklasad1, line 279: does iterators work here? would be nice to get rid of the `c-style for loops` if possible.

src/controllers/blocks/BlocksController.ts
```diff
@@ -233,4 +236,71 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+					historicApi,
+					options
+				);
+			});
+		}
+
+		const blocksPromise: Promise<unknown>[] = [];
+
+		const pQueue = new PromiseQueue(4);
+		for (let i = 0; i < tasks.length; i++) {
```
niklasad1, line 293: does iterators work here? would be nice to get rid of the c-style for loops if possible.

src/util/PromiseQueue.ts
```diff
@@ -0,0 +1,102 @@
...
+ * A PromiseQueue, enforcing that no more than `maxTasks` number of tasks
+ * are running at a given time.
+ */
+export class PromiseQueue<T> {
+	// How many tasks are allowed to run concurrently?
+	#maxTasks: number;
+	// How many tasks are currently running concurrently?
+	#runningTasks = 0;
+	// The queued tasks waiting to run
+	#tasks: LinkedList<() => void>;
```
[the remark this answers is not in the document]
jsdw: that's true, the linkedlist was just easier to implement I guess, and with a double ended queue you probably want to make sure it shrinks down again when items are removed, which the linked list thing just handles :)

docs/src/openapi-v1.yaml
```diff
@@ -1589,6 +1616,10 @@ components:
           - Fee = 0
           - Misc = 1
           - All = 2
+    BlockRange:
```
jsdw, line 1619: I wonder whether this should just be `Blocks`?

src/controllers/AbstractController.ts
```diff
@@ -196,6 +217,10 @@ export default abstract class AbstractController<T extends AbstractService> {
 		);
 	}
 
+	private verifyInt(num: Number): boolean {
+		return Number.isInteger(num) && num > 0;
+	}
```
jsdw, line 222: Just a random thought; I wonder how many of these sorts of functions could just be standalone utility functions and not be associated with a class in some way (once something is in a class, it has access to everything under `this` and so in my mind can be a little harder to reason about what they might do, whereas with standlone functions in a utility file you know exactly all of the data they will have access to)

src/controllers/AbstractController.ts
jsdw: If somebody asks for the range 1-10000000, this sort of thing will blow up quite quickly I guess?

It might be better to represent the range like some ` class Range { start: number, end: number }` that can be turned into an iterator to iterate over each number?

TarikGul: So I did actually think of this. But not sure what the limit should be. On a local archive node I was able to query 10,000 blocks within 30 seconds. So my guess is to put a 10k max on it, with an overrideMax query param?

docs/src/openapi-v1.yaml
TarikGul: So we have `Block` already and I didnt want to confuse others. But maybe the plural `Blocks` is ideal.

src/controllers/AbstractController.ts
TarikGul: Yea this is totally correct, will add that check.

src/controllers/AbstractControllers.spec.ts
TarikGul: Yes, good call.

src/controllers/blocks/BlocksController.ts
TarikGul: So iterators aren't the best choice here. They are not as clean as iterators in rust for example, but performance is relevant here and c-style for loops are the best for performance in javascript. 

Iterators are great in JS for anything where we want to wait for execution until the next iteration (especially when we want to call async calls within a for loop), but because we are just initializing an array with callbacks it's ideal here for the c-style for loop.

src/controllers/AbstractController.ts
jsdw: if there was an `overrideMax` query param, wouldn't that allowa request to blow up the server still? Or do you mean ENV var perhaps?

How many blocks is something likely to realistically want to act for at once, I wonder? A response containing 10k blocks would still be quite large and be quite an easy DoS vector for any public sidecar instance (though that might be true of other requests, and if sidecar is mainly used internally maybe it's not a huge deal?)

I'd go for something like a 100-200ish max offhand, without knowing what people actually would be requesting back?

docs/src/openapi-v1.yaml
jsdw: Yeah; I mean it's just a nit, so I'm not bothered either way, but if I saw `Blocks` I'd instantly understand it's relationship with `Block`, whereas `BlockRange` could be something totally different.

src/controllers/blocks/BlocksController.ts
niklasad1: ok, ty for the explaination.

my JS skillz :see_no_evil: :)

src/controllers/AbstractController.ts
TarikGul: Yea I think in general, users shouldn't ever really be using sidecar against public nodes and just their own archive nodes. Or atleast thats what all the exchanges, external teams use it for. I would propose maybe a 500 max with no override at all. Considering the requests are generally pretty fast all the way up to 1k blocks.

docs/src/openapi-v1.yaml
TarikGul: Yea makes sense, i'll change it up. I was kinda conflicted when writing these up anyways. I dont like the name `BlockRange` anyways.

src/controllers/AbstractController.ts
TarikGul: Yea I am open to pushing this in our util folder. It's newly created with the `PromiseQueue` being added to sidecar so it would be nice to have these simpler verification helper functions in there.

TarikGul: It's a start to definitely clean things up.

src/controllers/AbstractController.ts
jsdw: I'd be up for 500! I'd also have no issue with an env var or other config parameter so that exchanges etc can customise it if they want.

docs/src/openapi-v1.yaml
TarikGul: Fixed

src/controllers/AbstractController.ts
TarikGul: fixed

src/util/PromiseQueue.ts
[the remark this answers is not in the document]
TarikGul: Given this notion, do we want to implement a double ended queue, or is a linked list sufficient?

niklasad1 (approved):

src/controllers/AbstractController.ts
TarikGul: You mean with verifying UInt and a NonZeroUInt correct?

src/util/PromiseQueue.ts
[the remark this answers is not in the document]
niklasad1: leave it out this PR, I think the networking is the bottleneck anyway.

but maybe worth investigating :)

src/controllers/AbstractController.ts
niklasad1: yeah, or something even better.

TarikGul: Yea i agree, it will always for concise behavior between the min and max as well. Will put that in now.

src/controllers/AbstractController.ts
```diff
@@ -180,13 +182,44 @@ export default abstract class AbstractController<T extends AbstractService> {
...
+		const min = Number(splitRange[0]);
+		const max = Number(splitRange[1]);
+
+		if (!this.verifyInt(min) || !this.verifyInt(max)) {
+			throw new BadRequest('Inputted range contains non integers.');
+		}
+
+		if (min >= max) {
+			throw new BadRequest(
+				'Inputted min value cannot be greater than the max value.'
```
jsdw, line 210: Nit:

```suggestion
				'Inputted min value cannot be greater than or equal to the max value.'
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -233,4 +236,72 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+	> = async (
+		{ query: { range, eventDocs, extrinsicDocs } },
+		res
+	): Promise<void> => {
+		if (!range) throw new BadRequest('range query parameter must be inputted.');
+
+		// We set a max range to 500 blocks.
+		const rangeOfNums = this.parseRangeOfNumbersOrThrow(range, 500);
+		const rangeOfNumsToHash = await Promise.all(
+			rangeOfNums.map(async (n) => await this.getHashForBlock(n.toString()))
```
jsdw, line 260:
```suggestion
			rangeOfNums.map((n) => this.getHashForBlock(n.toString()))
```

I guess this is identical anyway?

jsdw, line 260: Should this be queued at all too or happy to do all of the number -> hash requests at once, as opposed to block details which are queued to do 4 at a time?

src/controllers/AbstractController.ts
TarikGul: Okay cool, so the verifyInt has all been removed and added to its own integers util folder.

src/controllers/AbstractController.ts
TarikGul: Okay yea i cleaned it up to handle Uint correctly and Non zero uints correctly. I think its way cleaner and concise about what the error messages will return.

src/controllers/blocks/BlocksController.ts
jsdw: I almost wonder whether you could combine a bunch of steps into something like:

```
const pQueue = new PromiseQueue(4);
const blocksPromise: Promise<Unknown>[] = [];

for (let i = 0; i < rangeOfNums.length; i++) {
    let result = pQueue.push(async () => {
        let hash = await this.getHashForBlock(n.toString());
	const historicApi = await this.api.at(hash);
	return await this.service.fetchBlock(
		hash,
		historicApi,
		options
	);
    });
    blocksPromise.push(result);
}
```

That way there would be no intermediate array for tasks and no intermediate array for block hashes, and also everything is queued, rather than hashes being all obtained at once. WDYT?

TarikGul: yea this is correct we dont need the async await

TarikGul: Yea something like this would be cleaner. Having it all in one iteration is way more sound IMO. I agree. Will update it right now to work, and run some quick benchmarks.

TarikGul: Thanks for the suggestion.

jsdw (approved):

jsdw:
Looks great!

PR #962, opened 2022-06-15
ci: increase benchmark threshold

TarikGul:
This increases the benchmark threshold to 35000. Some calls may take some time to resolve since the benchmarks focus on stressing the server. This will also stop us from overloading our issues with benchmark regressions. If its over 35000 then I can take the time to investigate the current version.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -224,7 +224,7 @@ check-benchmark:
   variables:
     GITHUB_REPO:                   "paritytech/substrate-api-sidecar"
     CI_IMAGE:                      "paritytech/benchmarks:latest"
-    THRESHOLD:                     13000
+    THRESHOLD:                     35000
     GITHUB_TOKEN:                  $GITHUB_PR_TOKEN
   script:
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):
Do we really need the benchmark threshold at all?

Fine for me, I assume this is tightly coupled to the RPC node. 
IIRC, the sidecar RPC node was switched out to a shared one which may decrease the throughput...

sergejparity (approved):

TarikGul:
> Do we really need the benchmark threshold at all?
> 
> Fine for me, I assume this is tightly coupled to the RPC node. IIRC, the sidecar RPC node was switched out to a shared one which may decrease the throughput...

Yea I see where you are coming from. I think the threshold holds one merit where if there is an egregious request that takes too much time, it might be worth investigating. Hence the large increase. I think everything we have seen so far is kinda normal behavior, but there was times in the past where requests would take 40000ms and we were able to find regressions. (This had to do with the metadatav14 transition). So all in all I think it might be nice to have this be a last barrier of sanity.

PR #966, opened 2022-06-17
fix: set minCalcFee runtime to 1058 for kusama

### LINKED ISSUE #429: Investigate supporting fee calculations in all Kusama runtimes
Currently we only support Kusama runtime spec versions >= 1062

https://github.com/paritytech/substrate-api-sidecar/blob/081b8b3c9712528f64c8a3486d71ca85076a5dc3/calc/src/calc_fee.rs#L43-L47

With recent increase interest in Kusama it is worth seriously exploring how we can indefinitely support fee calculations for older Kusama runtimes.

TarikGul:
closes: [#429](https://github.com/paritytech/substrate-api-sidecar/issues/429)

The lowest runtime which supports fee calculation given payment::queryInfo is 1058 for kusama. 
This updates that, and adds an e2e test for it.

### DIFF
--- a/e2e-tests/endpoints/kusama/blocks/2101439.json
+++ b/e2e-tests/endpoints/kusama/blocks/2101439.json
@@ -1,0 +1,212 @@
+{
+	"number": "2101439",
+	"hash": "0x8b0fbe687c87fc7c18c21fd255f0010cffcf34a35cb67187d88d110612c50d65",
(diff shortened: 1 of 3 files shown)


jsdw (approved):
How did you work out what the lowest runtime version was to support it out of interest?

niklasad1 (approved):

Issue #756, opened 2021-11-10
Fetch tx fee from onchain events


xlc:
It will be useful if there is a mode to query tx fee info from onchain events data instead of calculating it.

But need to consider about https://github.com/paritytech/polkadot/issues/4230

joepetrowski:
We explicitly decided against this because there is no guarantee that the `balances.Deposit` and `treasury.Deposit` events are related to transaction fees. For example, if you just send a transfer to the Treasury, there will be a `treasury.Deposit` event.

We also discussed adding an event to the runtime for the fee, but decided against that as well because events are not free and the information can be deterministically reproduced by...something like Sidecar.

joepetrowski:
Although I suppose if a parachain wanted to add a runtime event for this purpose, then sure, we could send Sidecar down that logic road. So we could make certain chain configs look for an event instead of providing a `calc_fee` package.

xlc:
I see. That make sense. I know Subscan is using events to fetch fees and they had some issues with new runtime.

Also the last treasury deposit should always be the fee deposit right?

joepetrowski:
> Also the last treasury deposit should always be the fee deposit right?

I thought of that too when we first designed Sidecar, but decided it was not an assumption we could rely on.

xlc:
I see. While that may not be true universally, it could be true for a particular parachain so could be a useful opt-in option.

TarikGul:
I wanted to bring this back to light. 

I think it could be useful to have a query parameter such like `?calcFeeByEvent=treasury_deposit`, this way parachains can opt in to the events they feel might be useful for calculating fees, but by default we would stick to the calcFee package. This could be a good opportunity to abstract the fee calculation logic for `/blocks`, and seperate it into a cleaner more "plug-in style" container. 

1. Could give more extensibility to Parachains, and users for how they would like their fees to be represented or fetched (Makes sense in regards to what Joe said about a chain potentially having their own runtime logic with an event like this attached).

2.  `BlocksService` is starting to get large, and having the ability to pull fee calculation logic out, and create a clean way for parachains to build on top of, or contribute to will help a lot. 

3. Potential for performance increases for Parachains if sidecar doesn't have to calculate the fees, since some unique runtime event will have it stored. 

Just some thoughts I had.

xlc:
Now transaction payment pallet is emitting an event to include the fee amount, sidecar should support it https://github.com/paritytech/substrate/pull/11618

[...]

PR #970, opened 2022-06-24
feat: support for `feeByEvent` query param which will abstract the fees by events

### LINKED ISSUE #969: Mismatched fee amount for Acala
Reported by an exchange. For this transaction: https://acala.subscan.io/extrinsic/0x31d946cd2e6939d394d9160383e552a3c46a38fb8b7cb5e8e2c96d089e531990

The fee from sidecar is `2490128142` but the actual fee is `2490128143`

TarikGul:
## Feature:

Query Parameter: `feeByEvent` -> Defaults to false. When true it will collect the `partialFee` by the events of each extrinsic. 

### How?

1. Call `rpc::payment::query_info` in order to retrieve an estimated partialFee
2. We search through the events per extrinsic looking for 3 specific events (This list of events are subject to grow as `Event::TransactionPaidFee` is added).
      a) Look through the Balances::Event::Withdraw event. We than compare the value given in the data to the estimated partialFee to confirm that there is some commonality. In this case we look for a less than `00.00001` difference in atomic value. (This is a very arguable approach, and I don't think it holds any real mathematical integrity, i Just used that abritrary value as a way to get this working). 
      b) If the above isn't satisfied we do the same for `Treasury::Event::Deposit`, and do the same comparison. 
      c) If the above isnt satisfied we then add up all the `Balances::Event::Deposit` values and compare that to the esimated partialFee. 
      d) if none of those values are satisfied we return an error to log into the `info` field for the extrinsic.

### Todo

- [x] Documentation
- [x] Tests 

closes: #969 #756 
    
Note: This PR will be followed by another PR to add support for `Event::TransactionPaidFee`


src/services/blocks/BlocksService.ts
```diff
@@ -450,6 +469,123 @@ export class BlocksService extends AbstractService {
...
+				method.pallet === 'balances'
+		);
+		if (withdrawEvent && withdrawEvent.data) {
+			const eventArr = withdrawEvent.data.toJSON();
+			if (Array.isArray(eventArr)) {
+				const fee = (eventArr as Array<number>)[eventArr.length - 1];
+				const diff = subIntegers(new BN(fee), partialFee).toString();
+
+				// The difference between values is 00.00001% or less so they are alike.
+				if (
```
TarikGul, line 519: It's this comparison right here that I would like some input on. What is a better way to ensure that the given value is indeed an inclusion. I do think comparing the estimated value from `rpc::payment::queryInfo` is a good approach but mathematically what constraints do we think are viable or accurate enough?

@joepetrowski @IkerAlus @jsdw

TarikGul: Important to note here too, this is not regarding parity based chains such as polkadot, kusama, and westend. But external chains and parachains.

jsdw: Aah, I think I'd assumed that `feeByEvent` would be a query param that accepted something like `?feeByEvent=balances.withdraw.4` to point at the event/property-in-event to pull a fee out of (for when parachains know what events they want to be inspected to find the actual fee).

But I guess the point here is that when a fee is taken, we probably expect one of these specific events to signal that some token has been taken for the fee? But this might vary based on the parachain and how they implement these things?

Code wise, it feels a bit "messy" to have to guess like this; we're essentialy saying "hmm this number looks about right so we'll use that!". But I appreciate that there may be no better way until that more standard TransactionFee event is a thing? What about making them point at the specific event they want us to use?

We could cut down on some repetition too (a `function findEvent(events, pallet, method)` and a `function areNumbersCloseEnough(a, b)` sort of thing for instance) :)

TarikGul: > Aah, I think I'd assumed that feeByEvent would be a query param that accepted something like ?feeByEvent=balances.withdraw.4 to point at the event/property-in-event to pull a fee out of (for when parachains know what events they want to be inspected to find the actual fee).

Per our conversation offline, it makes sense to keep the current implementation. As the guesswork is not perfectly ideal, it will allow for support for historic blocks. Once `TransactionPaidEvent` is in, then we can set that as the default event to retrieve fee's from and keep this implementation as a fallback for historic blocks as mentioned above. 

> We could cut down on some repetition too (a function findEvent(events, pallet, method) and a function areNumbersCloseEnough(a, b) sort of thing for instance) :)

Yea I will try to clean it up but each event has the `Balance` value potentially in a different location in the data array, but there might be a good way to organize it more.

TarikGul: @jsdw So I actually took your suggestions and cleaned up quite a bit. I think it's way better now. As soon as this PR goes in, I will add the `TransactionPaidFee` event as well.

src/services/blocks/BlocksService.ts
```diff
@@ -450,6 +469,153 @@ export class BlocksService extends AbstractService {
...
+			if (this.areFeesSimilar(sumOfFees, partialFee)) {
+				return {
+					partialFee: sumOfFees.toString(),
+				};
+			}
+		}
+
+		return {
+			partialFee: partialFee.toString(),
+			error: 'Could not find a reliable fee within the events data.',
```
jsdw, line 577: I know I suggested returning an error at some point, but I am wondering whether, especially called from the `/blocks` endpoint, it makes sense or not? I'm not sure how often we'll find the fee in events vs having to use the queryInfo result?

We could choose to acknowledge that we'll use events if possible, but if no events are available, it's not an error. We should tell the user whether we managed to use an event or not, so I wonder about a return type here that is just `string | undefined` (either we found the fee in events or not), and then in `getPartialFeeInfo` a return type like:

```
{
    partialFee: string,
    partialFeeFromEvent?: boolean
    error,
    dispatchClass
}
```

Just a thought tho!

src/services/blocks/BlocksService.ts
```diff
@@ -450,6 +469,153 @@ export class BlocksService extends AbstractService {
...
+				const fee = (dataArr as Array<number>)[dataArr.length - 1];
+
+				// The difference between values is 00.00001% or less so they are alike.
+				if (this.areFeesSimilar(new BN(fee), partialFee)) {
+					return {
+						partialFee: fee.toString(),
+					};
+				}
+			}
+		}
```
jsdw, line 540: Def looks tidier now; I like it!

jsdw (approved):
LGTM; just a thought about returning an error vs some `partialFeefromEvent` for when we get the fee from events vs don't manage to :)

lexnv (approved):

LXHLeaner:
Is it now reading the estimated fee amount instead of the actual fee？

PR #977, opened 2022-07-04
chore(dev): update @substrate/dev

TarikGul:


lexnv (approved):

package.json
```diff
@@ -36,7 +36,6 @@
     "dev": "tsc-watch --onSuccess \"yarn run main\"",
     "test": "NODE_ENV=test substrate-exec-jest --detectOpenHandles",
     "test:watch": "NODE_ENV=test substrate-exec-jest --watch",
-    "test:cov": "NODE_ENV=test substrate-exec-jest --detectOpenHandles --coverage",
```
niklasad1, line 39: why was this removed?

TarikGul: Good question. We dont actually use it anywhere, but also with Jest updated, it requires us to set the `coverageProvider` to `v8` and off of the default which is babel. `v8` is also experimental at the moment and causes some side-affects for us in the CI, so I removed it.

jsdw (approved):

PR #990, opened 2022-07-20
fix(feeByEvent): sanitize fee for hex values

### LINKED ISSUE #989: get Shiden 1500094 fee failed
get the dataArr[0] is WuSTANhkhikbhfsifpeTW3dHvb4UotErkCCs7sLxxGfg57z
get the dataArr[1] is 0x000000000000000000119e2433bf11f3.

and new BN(fee) is err 

// The difference between values is 00.00001% or less so they are alike.
				if (this.areFeesSimilar(new BN(fee), partialFee)) {
					return {
						partialFee: fee.toString(),
					};
				}

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/989

When a fee is represented as a hex value, sanitize the fee correctly.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -30,7 +30,7 @@ import {
 } from '@polkadot/types/interfaces';
 import { AnyJson, Codec, Registry } from '@polkadot/types/types';
 import { ICompact, INumber } from '@polkadot/types-codec/types/interfaces';
-import { u8aToHex } from '@polkadot/util';
+import { hexToNumber, isHex, u8aToHex } from '@polkadot/util';
 import { blake2AsU8a } from '@polkadot/util-crypto';
 import BN from 'bn.js';
 import { BadRequest, InternalServerError } from 'http-errors';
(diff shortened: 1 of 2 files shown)


jsdw (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -532,7 +532,9 @@ export class BlocksService extends AbstractService {
 		if (withdrawEvent.length > 0 && withdrawEvent[0].data) {
 			const dataArr = withdrawEvent[0].data.toJSON();
 			if (Array.isArray(dataArr)) {
-				const fee = (dataArr as Array<number>)[dataArr.length - 1];
+				const fee = this.sanitizeFee(
+					(dataArr as Array<string>)[dataArr.length - 1]
```
Xanewok, line 536: Did the shape change here or do we want to have it seen as `string` for the `sanitizeFee`/`isHex` call?

TarikGul: So the shape was inaccurately type casted for fees (technically number or string work here for BN.js but I changed it to its proper type to be more accurate). The type returned from `.toJSON()` should be a string, because any u128/u64 value interpreted by polkadot-js will be returned as a string.

lexnv (approved):

PR #1001, opened 2022-07-31
feat!: add support for http provider using SAS_SUBSTRATE_URL

### LINKED ISSUE #999: [Feature] Support the HTTP provider for connection to a chain
## Summary

rel: https://github.com/paritytech/substrate-api-sidecar/issues/326

The HTTPProvider is similar to the WS provider but will allow access to a http connection to a chain. See the above issue for an example on implementation. 

There will be a couple of things that will need to be added to the Specs for Sidecar. 
1. A new env var called `SAS_SUBSTRATE_HTTP_URL`. If the http url is inputted than it would override the default WS url. 

2. Logic to the Provider field will need to be updated to handle the above in `main.ts`.

3. This will need to be added to the logging.

TarikGul:
## Summary (BREAKING CHANGE)

Add support for connections to http addresses. The ENV var `SAS_SUBSTRATE_WS_URL` is no longer available and we are migrating to `SAS_SUBSTRATE_URL`. It can now accept both a http and ws URL's. 

closes: https://github.com/paritytech/substrate-api-sidecar/issues/999


README.md
```diff
@@ -163,6 +163,8 @@ For more information on our configuration manager visit its readme [here](https:
 
 - `SAS_SUBSTRATE_WS_URL`: WebSocket URL to which the RPC proxy will attempt to connect to, defaults to
     `ws://127.0.0.1:9944`.
+- `SAS_SUBSTRATE_HTTP_URL`: Http URL to which the RPC proxy will attemp to connec to. This is optional, and the default
```
lexnv, line 166:
```suggestion
- `SAS_SUBSTRATE_HTTP_URL`: Http URL to which the RPC proxy will attempt to connect to. This is optional, and the default
```

src/Specs.ts
```diff
@@ -97,6 +97,14 @@ export class Specs {
 			})
 		);
 
+		// HTTP
+		this._specs.appendSpec(
+			MODULES.SUBSTRATE,
+			this._specs.getSpec(CONFIG.HTTP_URL, 'Http URL', {
+				mandatory: false,
```
lexnv, line 104: Nit: Should this also have a `regexp` to properly validate the HTTP URL? Similar to `WS` to ensure we can't pass arbitrary URLs without port numbers?

lexnv (approved):
LGTM! Wondering if we could use regex to ensure the HTTP URL has a valid port number, but no strong feelings about it 😄

src/main.ts
```diff
@@ -44,7 +44,9 @@ async function main() {
 	const { TYPES_BUNDLE, TYPES_SPEC, TYPES_CHAIN, TYPES } = config.SUBSTRATE;
 	// Instantiate a web socket connection to the node and load types
 	const api = await ApiPromise.create({
-		provider: new WsProvider(config.SUBSTRATE.WS_URL),
+		provider: config.SUBSTRATE.HTTP_URL
```
jsdw, line 47: Since we have to check and decide on which env var URL we're using here, how about passing the "for sure" url into `startUpPrompt` rather than having it also do the check?

src/types/sidecar-config/CONFIG.ts
```diff
@@ -21,6 +21,7 @@ export enum CONFIG {
 	BIND_HOST = 'BIND_HOST',
 	PORT = 'PORT',
 	WS_URL = 'WS_URL',
+	HTTP_URL = 'HTTP_URL',
```
jsdw, line 24: This works! Another approach would be to just accept a URL and check whether it was a WS or HTTP one, rather than having a separate env var for each; that way, there's no need to wonder which is the default or handle both being provided or whatever :)

niklasad1: I really like that idea :)

src/Specs.ts
TarikGul: Yea I agree, I had it before but it would fail everytime because if there was no default addr it would see an empty value and fail. But given the suggestions below I think it will be much easier :)

src/types/sidecar-config/CONFIG.ts
TarikGul: Yea at first I wanted to do the same thing, but I didn't do it to avoid a breaking change. But that being said I think this holds merit and is a lot better in the long run! I'll switch it over

TarikGul:
@jsdw The PR is updated to accept `SAS_SUBSTRATE_URL`.

src/Specs.ts
```diff
@@ -87,13 +87,13 @@ export class Specs {
...
-		// WS
+		// WS OR HTTP
 		this._specs.appendSpec(
 			MODULES.SUBSTRATE,
-			this._specs.getSpec(CONFIG.WS_URL, 'Websocket URL', {
+			this._specs.getSpec(CONFIG.URL, 'Websocket or HTTP URL', {
 				default: 'ws://127.0.0.1:9944',
 				mandatory: true,
-				regexp: /^wss?:\/\/.*(:\d{4,5})?$/,
+				regexp: /^(ws|wss|http)?:\/\/.*(:\d{4,5})?$/,
```
sergejparity, line 96:
```suggestion
				regexp: /^(ws|wss|http)?:\/\/.*$/,
```

Just for sake of clarity. This optional part of regexp `(:\d{4,5})?` totally have no meaning after `.*`

niklasad1: good catch, thanks.

@TarikGul is it possible to get rid of this regex and use a URI/URL parser instead?

src/main.ts
```diff
@@ -97,26 +99,33 @@ async function main() {
  * @param wsUrl websocket url of the node Sidecar is connected to
  * @param chainName chain name of the network Sidecar is connected to
  * @param implName implementation name of the node Sidecar is connected to
+ * @param httpUrl http url of the node Sidecar is connected to. This is optional,
+ * but when used will override the ws url.
  */
-function startUpPrompt(wsUrl: string, chainName: string, implName: string) {
+function startUpPrompt(
+	wsUrl: string,
```
niklasad1, line 106: I reckon that it might be cleaner just to use `url` here and get rid of the ´wsUrl and httpUrl`

niklasad1 (approved):

src/Specs.ts
TarikGul: @sergejparity Thanks for that catch.

@niklasad1 I think it's probably best to leave the regex since the support is built in directly into `confmgr`, it also gives some solid logging associated with it, and removing it and doing our own check would complicate the SpecConfig. 

If it had an option to accept a callback to do custom validation I would be down, but it currently doesn't. Heres a link to the options available: https://gitlab.com/chevdor/confmgr/-/blob/master/src/types/optionTypes.ts#L9-15

jsdw (approved):

PR #1003, opened 2022-08-01
fix(feeByEvent): handle tip for partialFee

TarikGul:
rel: https://github.com/paritytech/substrate-api-sidecar/issues/969#issuecomment-1201214583

A partialFee needs to not include the tip. When using feeByEvent, if there is a tip we need to make sure we subtract it from the fee data provided by the event.


jsdw (approved):
LGTM!

lexnv (approved):
LGTM 👍

PR #1007, opened 2022-08-03
feat: add /accounts/:accountId/convert endpoint

Imod7:
## Description

This endpoint will receive as an input an Account ID or a Public key (hex) that is given by the user as the path parameter and convert it into an `SS58` address. The conversion will depend on the values of the 3 query parameters that are mentioned below. 

This implementation was motivated by this [comment](https://github.com/paritytech/substrate-api-sidecar/pull/926#discussion_r879451992) while making changes in the `validate` endpoint.

**From SS58 Address to Account ID also ?**
Does this endpoint also implements the reverse conversion, meaning from SS58 Address to Account ID ? No because this functionality is already covered by the `/accounts/:address/validate` endpoint after this [PR](https://github.com/paritytech/substrate-api-sidecar/pull/926) was merged.

## Query Parameters

This endpoint can take the following 3 query parameters :

- `scheme`
- `prefix`
- `publicKey`

### Possible values

The possible values for these parameters are the following :

- `scheme` can take one of those 3 values `{sr25519, ed25519, ecdsa}`
- `prefix` can be one of the prefices found [here](https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json)
- `publicKey` is a boolean value and can be `true` or `false`

### Default values
The default values for these query params are the following :

- `scheme` = `sr25519`
- `prefix` = `42`
- `publicKey` = `true`

## Examples
- So, for example if the query parameters that are given are the following :
   - `prefix=2`
   - `scheme=ed25519`
   - `publicKey=false`

- then that means that the endpoint will give as an output :   
   
   - a Kusama formatted SS58 address (based on the [ss58-registry](https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json) the prefix=2 corresponds to the Kusama network)
   - based on the `ed25519` algorithm (default value)
   - assuming that the given path parameter is an `Account ID` (and not a Public Key (hex))

- If no query parameters are specified then the endpoint will take their default values (as defined above) and will output :
   - a Substrate formatted SS58 address (default value of `prefix` = 42 which corresponds to substrate)
   - based on the `sr25519` scheme (default value) and 
   - assuming that the given path parameter is a Public Key (hex)

## Special Case / Use of query param `publicKey`
In most of the cases we can output a valid SS58 address based **only** on the `scheme` and `prefix` since Public Key (hex) is identical to the Account ID (so specifying the query param `publicKey` does not make any difference).

However, there is a specific case for which we also need to know if the path parameter that is given is : 
- a Public Key (hex) or
- an Account ID 

hence we need the query parameter `publicKey` to be specified as `true` or `false`. 

This case is when we have `scheme=ecdsa`. 
When we have `ecdsa` as scheme then Public key is different than the Account ID so we first need to hash the public key with `blake2` and then get the SS58 address from that.

*** This is also explained in this substrate [PR](https://github.com/paritytech/substrate/pull/8674#issue-868636081) (_subkey: display SS58 encoding of public key_)

## Sample Response

Sample request :
/accounts/0x002b5bab9258f5aff3117fce47854ef4c29c0eeae3faf855a2d6a9b67861c598/validate?prefix=0

Sample response :

```
{
  "ss58Prefix": "42",
  "network": "substrate",
  "address": "5C4vjiCiAiVYcx5fZ9g9FftViRqRwFRNpSDVrh6dzE35x2d8",
  "accountId": "0x002b5bab9258f5aff3117fce47854ef4c29c0eeae3faf855a2d6a9b67861c598",
  "scheme": "sr25519",
  "publicKey": true
}
```

### chains-config
It was also added in `chains-config` -> Controllers of : 

- Polkadot
- Kusama
- Westend

## Todos
- [x] Double check if the conversion (from `accountId` to `address`) that happens in the `accountConvert` method is correct
- [x] Make use of the query param `public` (because currently is not used)
- [x] Update Docs
- [x] Add Tests


jsdw:
Dumb question probably (I read the comments this came from and didn't quite understand); what does the `public` query param do? Does it apply to every scheme?

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+	protected initRoutes(): void {
+		this.safeMountAsyncGetHandlers([['', this.accountConvert]]);
+	}
+
+	private accountConvert: RequestHandler = (
+		{ params: { address }, query: { scheme, prefix, publicKey } },
+		res
+	) => {
+		// Validation of the `scheme` query param
+		const cryptoScheme = (
```
TarikGul, line 40: We can clean this up with some more specific type level validation. 

For example: we can set the `RequestHandler` to accept types we create.

Ex: 
```typescript
interface IConvertQueryParams extends Query {
    scheme: string; // You might be able to set this to a union type of `'ed25519' | 'sr25519' | 'ecdsa'`
    prefix: string;
    publicKey: string;
}

// Then you would just have RequestHandler type be like
RequestHandler<unknown, unknown, unknown, IConvertQueryParams>
```

Please use https://github.com/paritytech/substrate-api-sidecar/blob/master/src/types/requests.ts#L49 as guidance

TarikGul: This will also clean up the type checks below.

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+				String(publicKey).toLowerCase() === 'true' ||
+				String(publicKey).toLowerCase() === 'false' ||
+				publicKey === undefined
+			)
+		) {
+			throw new BadRequest(
+				'`publicKey` can be either true or false (boolean value)'
+			);
+		}
+		const pubKey = typeof publicKey !== 'string' ? true : publicKey === 'true';
```
TarikGul, line 74:
```suggestion
		const pubKey = publicKey === 'true';
```

TarikGul: No need to commit this as it will break things probably on a type level. But easier to implement it like this once you clean the query Params types.

Imod7:
> Dumb question probably (I read the comments this came from and didn't quite understand); what does the `public` query param do? Does it apply to every scheme?

@jsdw Thank you so much for your question and it is a quite logical question to have since when I created this PR I was actually not even using the query param `public` (now renamed to `publicKey`). After a clarification from @IkerAlus it was also clear to me the purpose of this parameter and now I added some code that actually uses it.
Since this is one very important point/specific case of the implementation of this endpoint, I tried to explain it in the Description of the PR under the section "Special Case / Use of query param publicKey". I hope this explanation and the substrate PR (_subkey: display SS58 encoding of public key_) that I cite answers your question.

src/controllers/accounts/AccountsConvertController.ts
Imod7: Thank you @TarikGul for your comments and suggestions! Very much appreciated!

I tried the solution of defining the query params types in `requests.ts` but I can not see the value of this change very clearly since I still need to keep all of my validation code in the controller. Maybe I just implemented in a wrong way so let me explain below what I tried to do.

I added the `interface` you suggested (the `union` type does not seem to have any effect in runtime so I just set the types to `string`). Adding this, resulted in the following changes in my code : 

- I replaced wherever I had `typeof` with `undefined`, e.g.
    changed this check 
    ```
    const networkPrefix = typeof prefix !== 'string' ? '42' : prefix;
    ```
    to this
    ```
    const networkPrefix = prefix === undefined ? '42' : prefix;
    ```
- The above point is the same for the other 2 query params
- However, I still need to check if the values given in the query params are valid, e.g. valid values for `scheme` should be `'ed25519' | 'sr25519' | 'ecdsa'` and if they are not, throw an error. If I do not have a check for this in the `controller` then it throws when it goes to the `service` and there the error is 
  ```
  "Expected a keyring type of either 'ed25519', 'sr25519', 'ethereum' or 'ecdsa', found 'domi"
  ```
  which is not what I would like to convey to the user.
- The same for the values of the `publicKey`. I would like to throw an error and tell the user to give a value of `true` or `false` instead of not throwing any errors and just set `publicKey` to `false`.

If the validation checks in the `AccountsConvertController.ts` make it confusing then maybe I could add a middleware for this ? Like it is implemented in [validateAddressMiddleware.ts](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/middleware/validate/validateAddressMiddleware.ts) ?

src/controllers/accounts/AccountsConvertController.ts
Imod7: I tried to change this but then it does not get the default value `true` whenever publicKey is `undefined`. So I changed back (with a slight change) to `const pubKey = !publicKey ? true : publicKey === 'true';`

Imod7:
I added the tests + docs in this PR.

In the `docs` I populated some "keys" at my own discretion since I am not sure if we follow a pattern/rule to populate those. These are the following : 
- the schema at the "200" response
    ```
    schema:
       $ref: '#/components/schemas/AccountConvert'
    ```
- the `operationId`
- Added 3 responses codes (200, 400, 404)

TarikGul:
1. Nice job, sounds good. 

> Another point is that in src/chains-configI only added the AccountConvert controller in Polkadot, Kusama and Westend. Should I add it in all the chains ?

2. No need to add it to the others, they can always push up a PR if they would like it.

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/AccountValidation'
+  /accounts/{accountId}/convert:
+    get:
+      tags:
+      - accounts
+      summary: Convert a given account Id to a SS58 address.
```
TarikGul, line 374:
```suggestion
      summary: Convert a given AccountId to a SS58 address.
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+      description: Returns the SS58 prefix, the network address format, the SS58 address, the account ID
```
TarikGul, line 375:
```suggestion
      description: Returns the SS58 prefix, the network address format, the SS58 address, and the AccountId
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+        that was given as input parameter, the scheme that was used and if it is a public key or not (boolean).
+      operationId: accountConvert
+      parameters:
+      - name: accountId
+        in: path
+        description: Account ID or Public Key (hex).
```
TarikGul, line 381:
```suggestion
        description: AccountId or Public Key (hex).
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+        required: true
+        schema:
+          format: Account ID or Hex
```
TarikGul, line 384:
```suggestion
          format: AccountId or Hex
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+          type: string
+      - name: scheme
+        in: query
+        description: The cryptographic scheme to be used in order to convert the account ID to
```
TarikGul, line 388:
```suggestion
        description: The cryptographic scheme to be used in order to convert the AccountId to 
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+        in: path
+        description: Account ID or Public Key (hex).
+        required: true
+        schema:
+          format: Account ID or Hex
+          type: string
+      - name: scheme
+        in: query
+        description: The cryptographic scheme to be used in order to convert the account ID to 
+          a SS58 address. It can take one of those three values {sr25519, ed25519, ecdsa}.
```
TarikGul, line 389:
```suggestion
          an SS58 address. It can take one of three values: sr25519, ed25519, ecdsa.
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+        description: The address prefix which can be one of the values found in the SS58-registry.
+        required: false
+        schema:
+          type: string
+          format: number
+          default: 42
+      - name: publicKey
+        in: query
+        description: Defines if the given value in the path parameter is a Public Key (hex)
+          or not (hence Account ID).
```
TarikGul, line 407:
```suggestion
          or not (hence AccountId).
```

docs/src/openapi-v1.yaml
```diff
@@ -367,6 +367,68 @@ paths:
...
+          default: true
+      responses:
+        "200":
+          description: successfully converted the account Id and retreived the address info.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AccountConvert'
+        "400":
+          description: Invalid Account Id
```
TarikGul, line 421:
```suggestion
          description: Invalid AccountId
```

docs/src/openapi-v1.yaml
TarikGul: The idea being we should set all the AccountId's to be the same in this portion of the docs. We also will need to rebuild the docs after updating these changes.

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	private accountConvert: RequestHandler<
+		IAddressParam,
+		unknown,
+		unknown,
+		IConvertQueryParams
+	> = ({ params: { address }, query: { scheme, prefix, publicKey } }, res) => {
+		// Validation of the `scheme` query param
+		const cryptoScheme = !scheme ? 'sr25519' : scheme;
```
TarikGul, line 42: Tiny nit, generally if possible I think it's better to have a truthy value be first followed by what the default value will be so its easier to read. Example:
`const cryptoScheme = scheme ? scheme : 'sr25519';`

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+				cryptoScheme === 'ecdsa'
+			)
+		) {
+			throw new BadRequest(
+				'`scheme` can have one of those 3 values [ed25519, sr25519, ecdsa]'
+			);
+		}
+
+		// Validation of the `prefix` query param
+		const networkPrefix = !prefix ? '42' : prefix;
```
TarikGul, line 56: Same nit as above. Not a big deal though.

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+		const ss58Prefix = this.parseNumberOrThrow(
+			networkPrefix,
+			'`prefix` provided is not a number.'
```
TarikGul, line 59:
```suggestion
			'The `prefix` query parameter provided is not a number.'
```

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,81 @@
...
+		);
+
+		// Validation of the `publicKey` query param
```
TarikGul, line 62: This is alright for now, but just for tracking sake, once https://github.com/paritytech/substrate-api-sidecar/pull/1023 goes in, we can update the code to reflect this endpoint.

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+import { defaultMockApi } from '../test-helpers/mock';
+import { AccountsConvertService } from './AccountsConvertService';
+
+const mockApi = {
+	...defaultMockApi,
+} as unknown as ApiPromise;
+const validateService = new AccountsConvertService(mockApi);
+
+describe('Convert accounts', () => {
+	it('Should convert a Substrate accountId to a SS58 Address with scheme=ecdsa', () => {
```
TarikGul, line 29:
```suggestion
	it('Should convert a Substrate accountId to a SS58 Address when `scheme` equals `ecdsa`', () => {
```

TarikGul: Not sure what you think but I think it's a little cleaner to keep test descriptions as complete english. Same for the tests below.

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+			'0x8832f0039ef4164834ccb0b454378a6d50f6979ddf7582a1cf911788a6cde537';
+
+		expect(
+			sanitizeNumbers(
+				validateService.accountConvert(astarAccountId, 'sr25519', 5, false)
+			)
+		).toStrictEqual(expectedResponse);
+	});
+
+	// Since the input parameter is a Public key (hex) (and not an Account ID),
```
TarikGul, line 109:
```suggestion
	// Since the input parameter is a Public key (hex) (and not an AccountId),
```

src/services/accounts/AccountsConvertService.spec.ts
TarikGul: Not a huge deal though, and will leave this up to you. It was just a thought when I was reading it. That being said there is cases in Sidecar where we do exactly what you did above with comparison symbols such as `>` and `<=` so im totally cool with keeping it the way it is if you think its best.

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+			'0x026e42c81603c7eaa2bdc40115306f05c94b563ff9e92120e8ea3480fec939e2e9';
+
+		expect(
+			sanitizeNumbers(
+				validateService.accountConvert(polkadotPublicKey, 'ecdsa', 0, true)
+			)
+		).toStrictEqual(expectedResponse);
+	});
+
+	// Since the input parameter is a Public key (hex) (and not an Account ID), if we
```
TarikGul, line 131:
```suggestion
	// Since the input parameter is a Public key (hex) (and not an AccountId), if we
```

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+	// set the publicKey=false the output is the Public key (SS58) and not a SS58 Address.
```
TarikGul, line 132:
```suggestion
	// set the publicKey=false, the output is the Public key (SS58) and not a SS58 Address.
```

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+			'0x026e42c81603c7eaa2bdc40115306f05c94b563ff9e92120e8ea3480fec939e2e9';
+
+		expect(
+			sanitizeNumbers(
+				validateService.accountConvert(polkadotPublicKey, 'ecdsa', 0, false)
+			)
+		).toStrictEqual(expectedResponse);
+	});
+
+	// We try to convert a Polkadot Account ID to a SS58 Address by setting the publicKey=true
```
TarikGul, line 153:
```suggestion
	// We try to convert a Polkadot AccountId to a SS58 Address by setting the publicKey=true
```

src/services/accounts/AccountsConvertService.ts
```diff
@@ -0,0 +1,74 @@
...
+import { IAccountConvert } from '../../types/responses/AccountConvert';
+import { AbstractService } from '../AbstractService';
+
+export class AccountsConvertService extends AbstractService {
+	/**
+	 * Takes a given accountId and converts it to a SS58 address
+	 * based on the values of the variables scheme, ss58Prefix & publicKey.
+	 * It also returns the network name.
+	 *
+	 * @param address ss58 or hex address to validate
```
TarikGul, line 32: Is this an address or accountId, I think we should be specific here with the docs and what the function is taking in as a named value.

src/services/accounts/AccountsConvertService.ts
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	accountConvert(
+		accountId: string,
+		scheme: 'ed25519' | 'sr25519' | 'ecdsa',
+		ss58Prefix: number,
+		publicKey: boolean
+	): IAccountConvert {
+		const accountIdIsHex = isHex(accountId);
+		if (!accountIdIsHex) {
+			throw new BadRequest('`accountId` is not a valid hex value.');
```
TarikGul, line 45: When we log this value as an Error is this referring to the `address` param in the URI.

src/services/accounts/AccountsConvertService.ts
```diff
@@ -0,0 +1,74 @@
...
+		let network = null;
+		for (const networkParams of allNetworks) {
+			if (networkParams['prefix'] === ss58Prefix) {
+				network = networkParams['network'];
+				break;
+			}
+		}
+		if (network === null) {
+			throw new BadRequest(
+				'`prefix` does not correspond to a existing network.'
```
TarikGul, line 56:
```suggestion
				'The given `prefix` query param does not correspond to an existing network.'
```

src/services/accounts/AccountsConvertService.ts
```diff
@@ -0,0 +1,74 @@
...
+	 * Takes a given accountId and converts it to a SS58 address
+	 * based on the values of the variables scheme, ss58Prefix & publicKey.
+	 * It also returns the network name.
+	 *
+	 * @param address ss58 or hex address to validate
+	 * @param scheme
+	 * @param ss58Prefix
+	 * @param publicKey
+	 */
+	accountConvert(
```
TarikGul, line 37:
```suggestion
	public accountConvert(
```

TarikGul (commented):
Overall very solid job, just some nits, and a few questions above.

docs/src/openapi-v1.yaml
Imod7: Thank you for all the suggestions and time spent to review! :pray:
I implemented all the other changes in the docs excepting adding the ":" because it thinks it is a key and throws an error while rebuilding the docs.

src/controllers/accounts/AccountsConvertController.ts
Imod7: Noted down so I keep an eye when the other PR is merged to come back to this one! Thank you! :pray:

src/services/accounts/AccountsConvertService.spec.ts
Imod7: I agree in changing it! I like it better :)

src/services/accounts/AccountsConvertService.ts
Imod7: Yes it is referring to the path parameter AccountId or Public Key (hex) that should be converted to an SS58 address.

src/services/accounts/AccountsConvertService.ts
Imod7: It is the AccountId or Public Key (hex). I changed it.

src/services/accounts/AccountsConvertService.ts
TarikGul: Do you think we should write `The given path parameter "accountId" is not a valid hex value`?

Imod7: After your first suggestion I changed most of the error messages included this one. 
It is now 
`'The `accountId` parameter provided is not a valid hex value.' `

I also like the one you suggested. Whatever you think is best.

TarikGul (approved):
Great Job, LGTM 👍 Just need to resolve the merge conflict

TarikGul (approved):
New commits look great! Lets get this merged in soon.

src/controllers/accounts/AccountsConvertController.ts
```diff
@@ -0,0 +1,71 @@
...
+		IConvertQueryParams
+	> = ({ params: { address }, query: { scheme, prefix, publicKey } }, res) => {
+		// Validation of the `scheme` query param
+		const cryptoScheme = scheme ? scheme : 'sr25519';
+		if (
+			!(
+				cryptoScheme === 'ed25519' ||
+				cryptoScheme === 'sr25519' ||
+				cryptoScheme === 'ecdsa'
+			)
```
jsdw, line 50: Totally happy with this. When there are lots of conditions you can also use an approach like `if (!['ed25519', 'sr25519', 'ecdsa'].includes(cryptoScheme)) { ... }`, but your way is more efficient :)

src/services/accounts/AccountsConvertService.ts
```diff
@@ -0,0 +1,76 @@
...
+				'The `accountId` parameter provided is not a valid hex value.'
+			);
+		}
+		let network = null;
+		for (const networkParams of allNetworks) {
+			if (networkParams['prefix'] === ss58Prefix) {
+				network = networkParams['network'];
+				break;
+			}
+		}
```
jsdw, line 55: Assuming `allNetworks` is an array, you can also do:

```ts
let network = allNetworks.find(networkParams => networkParams['prefix'] === ss58Prefix)"
if (!network) {
   throw new BadRequest(...)
}
```

Not a big deal though :)

jsdw (approved):
lgtm, nice work!

PR #1009, opened 2022-08-05
fix: support fee calculation for BifrostPolkadot

hqwangningbo:

PR #1012, opened 2022-08-05
fix: add support for Parallel and Heiko

mclyk:


TarikGul:
hey @MrPai Please see this comment I put on a similar PR that was just submitted: https://github.com/paritytech/substrate-api-sidecar/pull/1009#issuecomment-1206449946

Same applies to you in terms of keeping the PR up.

MrPai:
@TarikGul reviewed the explaining in another PR, thanks for it, if it's possible to merge it if you think the PR is ok, because one of our cooperative partners relies on your repo, hope this will not block them for testing and developing.

calc/src/calc_fee.rs
```diff
@@ -85,6 +85,9 @@ impl Multiplier {
 
             ("bifrost", _v) => V2(new_u128(inner)),
 
+            ("heiko", _v) => V2(new_u128(inner)),
+            ("parallel", _v) => V2(new_u128(inner)),
+
```
Imod7, line 90: @MrPai Thank you for your PR!
As already mentioned, if you would like this PR to be merged ASAP you can do the following : 

1. Revert/remove this change in `calc_fee` and
2. Change the title of the PR to something that it reflects the change in the chains config, e.g. "fix: add support (controllers) for Parallel and Heiko" (just an example).

Regarding the fee calculation, adding anything to `calc_fee` will not have any impact since we do not use it anymore. Please check the related comments [here](https://github.com/paritytech/substrate-api-sidecar/pull/1009#issuecomment-1206449946) and [here](https://github.com/paritytech/substrate-api-sidecar/pull/974#issuecomment-1205298958) by Tarik and James. 

Having said that, for fees we are advising to use the `?feeByEvent=true` query param found in the `/blocks`related endpoints. Please check also the [endpoints docs](https://paritytech.github.io/substrate-api-sidecar/dist/) where it mentions more details on `feeByEvent` : 

> When set to true, extrinsics will have their fees retrieved by their events. This query param will use a fee estimation via rpc::payment::query_info to accurately identify which event has the corresponding fee information that is closest to the value provided by query_info.

Imod7 (changes requested):

calc/src/calc_fee.rs
MrPai: @Imod7 thanks for your information, I have done what you mentioned, please take a look.

MrPai:
Hi, @TarikGul @Imod7 do we have any update now?

TarikGul (approved):

TarikGul:
@MrPai Hey LGTM, it'll go in tomorrow, then a release will happen on tuesday which will include these changes. It's still Sunday for the team so it will get another review tomorrow.

MrPai:
@TarikGul Thanks, have a nice day.

Imod7 (approved):
LGTM

PR #1015, opened 2022-08-11
feat: add support for ink! contracts

### LINKED ISSUE #1018: [Feature] `/contract/ink/{address}`
This feature request is for adding contracts support for !ink in Sidecar. This is a tracking issue for milestones, and is an extension of https://github.com/paritytech/substrate-api-sidecar/issues/795.

Further support and endpoints may be added in the future for contracts. This is just a general start. 

rel: https://github.com/paritytech/substrate-api-sidecar/pull/1015

TarikGul:
## Summary

Give contract support for !Ink, currently this will give suuport for calling messages to a specific contract. 

### POST `/contracts/ink/{address}/query`

This endpoint will call any method that is passed in as a query param that is not payable or mutating. 

#### **Body**

Content-type: application/json
JSON: Requires a JSON body with the metadata attached for the contract

#### **Params**
- **address**: The address associated with the contract

#### **Query Params** (Optional)
- method: The method to query. Will default as a get
- gasLimit: The gas limit as an option for querying a contract
- storageDepositLimit: the Storage deposit limit as an option for querying a contract
- args: arguments for querying the message method

The return response will be:

```
{
	debugMessage: string,
	gasConsumed: string,
	gasRequired: string,
	output: boolean,
	result: Record<string, unkown>,
	storageDeposit: Record<string, unknown>,
}
```

closes: https://github.com/paritytech/substrate-api-sidecar/issues/1018


TarikGul:
As a note I have written a branch off of this PR that adds the functionality to call payable and mutatiing ink transactions. 
The one concern is: should sidecar be signing transaction for users since it requires a signature. 

^ Discussion to be had.

TarikGul:
After some discussion it was decided that we will not add the functionality to call payable, and mutatable transactions for a given contract.

docs/src/openapi-v1.yaml
```diff
@@ -643,6 +644,62 @@ paths:
...
+        schema:
+          default: -1
+          type: number
+      - name: storageDepositLimit
+        in: query
+        description: The storage deposit limit to be used as an option for the queried message.
+        required: false
+        schema:
+          deafult: null
+          type: number
```
jsdw, line 684: Dumb question; what effect do these have (maybe I'll see if I look lower!)?

Does a contract have gas which is used even when something just queries it?

src/controllers/contracts/ContractsInkController.ts
```diff
@@ -0,0 +1,89 @@
...
+	 * @param res
+	 */
+	private callContractQuery: IPostRequestHandler<
+		IBodyContractMetadata,
+		IContractQueryParam
+	> = async (
+		{
+			params: { address },
+			body,
+			query: { method = 'get', gasLimit, storageDepositLimit, args },
```
jsdw, line 57: GET by default but containing a body I think is techinically ok, but I'm hesitant!

https://stackoverflow.com/questions/978061/http-get-with-request-body was an interesting read, and it points out that many caching things and layers between eg sidecar the the world might ignore bodies in GET requests and could hand back cached results or something.

This is just a default though, and the openapi bit says POST, so maybe just set this default to POST too?

src/services/contracts/ContractsInkService.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+		};
+	});
+
+const mockContractPromise = {
+	query: {
+		get: getFlipper,
+	},
+} as unknown as ContractPromise;
+
+describe('ContractsInkService', () => {
```
jsdw, line 51: This doesn't really test that much (the service hardly does anything really, and the rest is mocked out). Makes sense to have; maybe an e2e test against a real life contract would be more reassuring if possible?

src/controllers/contracts/ContractsInkController.ts
TarikGul: Ahh So i think you are misunderstanding what the `get` method is for. This is the method that will be used to call the contract. `get` just happens to be the default standard. 

That being said the endpoint just uses a POST request to attach the body and receive a response. After some research it seemed like the appropriate decision.

src/services/contracts/ContractsInkService.spec.ts
TarikGul: Yea I agree, I just need to make sure that there is some live contracts on Kusama or westend to test against. But yea I totally agree, an e2e test would be great.

TarikGul: So the only issue is, AFAIK there is no live contracts on Westend, Kusama, or Polkadot. Apparently there is some on Rococo, but I am hesitant to add another chain for e2e-tests as it's more to maintain. Let me know what your thoughts are.

src/controllers/contracts/ContractsInkController.ts
jsdw: Ooohhh I see! Ok that makes much more sense, thanks!

src/services/contracts/ContractsInkService.spec.ts
jsdw: Ok yeah; I thought it might be a bit tricky!

In the ideal case I guess we'd spin up a test node, deploy a contract to it and then attempt to use this endpoint to query some methods on the contract. Do we have any test infrastructure to help do that already anywhere? Is it in the realms of possibility to do?

TarikGul: We actually dont have much infratructure for this currently. IMO It would actually probably just be better to add roccoco to the e2e tests than go through implementing that infrastructure.

marshacb (approved):
lgtm!

jsdw (approved):

PR #1017, opened 2022-08-12
fix!: integrate @substrate/calc@0.3.0 for partial fees 

TarikGul:
## Summary (BREAKING CHANGES)

This PR focuses on adding `calc_partial_fee` from `@substrate/calc` v0.3.0 so that fee calculations for both relay and parachains are completely accurate. It uses a combination of both `payment_queryInfo` and `payment_queryFeeDetails` in order to calculate the fee. If `payment_queryFeeDetails` is not present or available to use, it will default to `payment_queryInfo`. This PR also adds a field to the `info` of an extrinsic called `kind` which can either be `preDispatch`, or `postDispatch`. More Info below.  

### Removed

 - **Breaking Change** - Remove `feeByEvent` query param as it is no longer needed as Sidecar now has the accuracy it needs for fees. In the future if the event `TransactionFeePaid` is present it will take the fee from there. 

### Added

- `kind: 'preDispatch' | 'postDispatch'` field in `info` for extrinsics. 
    1. `preDispatch`: `payment_queryInfo` was used to get the `partialFee` given `preDispatch` information.
    2. `postDispatch`: `calc_partial_fee` was used to calculate the fee using post dispatch fee information.
    
- `calc_partial_fee`: `/blocks/*` now uses this method to accurately calculate `partialFee`'s for extrinsics.

### Fixed

- Pass in the previous blockHash for an extrinsic query for `payment_queryInfo`

### Notes

- Since there is no real way to evaluate if `payment_queryFeeDetails` is available in the runtime, and polkadot-js will throw an error regardless of catching it. Therefore we log a warning after the error is emitted to let users know the error is not important, than we cache the runtime to not call `payment_queryFeeDetails` again and instead recieve the partialFee from `queryInfo`. 

## Todo

- [x] e2e-tests
- [x] Ensure Docs are complete
- [x] Find out what can be done to correctly handle `payment_queryFeeDetails` for a runtime that doesn't support it. 
    - 1. We cant catch the error from polkadot-js rpc call. It will log regardless of handling it.


src/services/blocks/BlocksService.ts
```diff
@@ -250,6 +250,10 @@ export class BlocksService extends AbstractService {
 				continue;
 			}
 
+			/**
+			 * Grab the initial partialFee, and information required for calculating a partialFee
+			 * if queryFeeDetails is available in the runtime.
```
niklasad1: :smiling_face_with_tear:

docs/src/openapi-v1.yaml
```diff
@@ -15,7 +15,7 @@ tags:
 - name: node
   description: node connected to sidecar
 - name: pallets
-  description: pallets employed in the runtime
+  description: pallets employed in the runtimex
```
jsdw, line 18:
```suggestion
  description: pallets employed in the runtime
```

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,68 @@
...
+		versionWithCall: number | null
+	) {
+		this._versionWithoutCall = versionWithoutCall;
+		this._versionWithCall = versionWithCall;
+	}
+
+	public isQueryFeeDetailsAvail(specVersion: number) {
+		const { available, notAvailable, unknown } = QueryFee;
+
+		if (this._versionWithCall && specVersion >= this._versionWithCall) {
```
jsdw, line 40: Worth doing `this._versionWithCall !== null` too? Else spec versions of 0 may be mishandled.

Also on line 44?

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,68 @@
...
+		return unknown;
+	}
+
+	public setVersionWithoutCall(specVersion: number) {
+		if (this._versionWithoutCall === null) {
+			this._versionWithoutCall = specVersion;
+		}
+		if (specVersion > this._versionWithoutCall) {
+			this._versionWithoutCall = specVersion;
+		}
```
jsdw, line 57:
```suggestion
		if (this._versionWithoutCall === null || specVersion > this._versionWithoutCall) {
			this._versionWithoutCall = specVersion;
		}
```

Maybe? Or make the second `if` an `else if` just to be super obvious that it's one or the other.

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	public setVersionWithCall(specVersion: number) {
+		if (this._versionWithCall === null) {
+			this._versionWithCall = specVersion;
+		}
+		if (specVersion < this._versionWithCall) {
+			this._versionWithCall = specVersion;
+		}
```
jsdw, line 66: Combine into one test or `else if`?

src/services/blocks/BlocksService.ts
```diff
@@ -251,27 +251,61 @@ export class BlocksService extends AbstractService {
...
+						previousBlockHash,
+						weightInfo.weight,
+						weight
+					);
+					dispatchFeeType = 'postDispatch';
+					this.queryFeeErrCache.setVersionWithCall(specVersion.toNumber());
+				} catch {
+					this.queryFeeErrCache.setVersionWithoutCall(specVersion.toNumber());
+					console.warn(
+						'The error above is automatically emmitted from polkadot-js, and there is no inherit error from your request.'
```
jsdw, line 298:
```suggestion
						'The error above is automatically emitted from polkadot-js, and can be ignored.'
```

src/services/blocks/BlocksService.ts
```diff
@@ -294,6 +328,80 @@ export class BlocksService extends AbstractService {
...
+
+		const num = blockNumber.toBn();
+		return num.isZero()
+			? await api.rpc.chain.getBlockHash(num)
+			: await api.rpc.chain.getBlockHash(num.sub(new BN(1)));
+	}
+
+	/**
+	 * Calculate the partialFee for an extrinsic. This uses `calc_partial_fee` from `@substrate/calc`.
+	 * Please reference
```
jsdw, line 378: Is this comment unfinished?

src/services/blocks/BlocksService.ts
```diff
@@ -294,6 +328,80 @@ export class BlocksService extends AbstractService {
...
+
+			return calc_partial_fee(
+				baseFee.toString(),
+				lenFee.toString(),
+				adjustedWeightFee.toString(),
+				estWeight.toString(),
+				extrinsicSuccessWeight.toString()
+			);
+		} else {
+			// When the inlcusion fee isNone we are dealing with a unsigned extrinsic.
```
jsdw, line 400:
```suggestion
			// When the inclusion fee isNone we are dealing with a unsigned extrinsic.
```

src/services/blocks/BlocksService.ts
```diff
@@ -471,183 +579,6 @@ export class BlocksService extends AbstractService {
...
-	 *
-	 * The order of the events we search through are:
-	 * 1.Balances::Event::Withdraw
-	 * 2.Treasury::Event::Deposit
-	 * 3.Balances::Event::Deposit
-	 *
-	 * @param events The events to search through for a partialFee
-	 * @param partialFee Estimated partialFee given by `payment::queryInfo`
-	 */
-	private getPartialFeeByEvents(
```
jsdw, line 525: Lovely to see this go :D

src/types/chains-config/ControllerConfig.ts
```diff
@@ -38,12 +39,24 @@ export interface ControllerConfig {
...
-		finalizes: boolean;
-		minCalcFeeRuntime: IOption<number>;
-		blockStore: LRU<string, IBlock>;
-	};
+	options: ControllerOptions;
+}
+
+export interface ControllerOptions {
+	/**
+	 * Wether or not the chain finalizes blocks
```
jsdw, line 47:
```suggestion
	 * Whether or not the chain finalizes blocks
```

jsdw:
Bunch of small things but great stuff :)

src/chains-config/cache/QueryFeeDetailsCache.ts
TarikGul: Cool changed them to `else if`s, at first my thought was stacked `if`s themselves look so much cleaner to me. I love the readability of them, but you're right, it makes more sense to just do an `else if` here

src/services/blocks/BlocksService.ts
TarikGul: complete the comment. I just needed to reference the user to check out the rust code to see docs on the algorithm we use.

niklasad1:
As a side-not to this PR when https://github.com/paritytech/substrate/issues/11648 is a thing I guess it would simplify things for us.

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	constructor(
+		versionWithoutCall: number | null,
+		versionWithCall: number | null
+	) {
+		this._versionWithoutCall = versionWithoutCall;
+		this._versionWithCall = versionWithCall;
+	}
+
+	public isQueryFeeDetailsAvail(specVersion: number) {
```
niklasad1, line 37:
```suggestion
        /*
        * Checks whether the current runtime version has access to the runtime API:   
        * `TransactionPaymentCallApi_query_fee_details` 
        */
	public hasQueryFeeDetails(specVersion: number) {
```

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,72 @@
...
+			this._versionWithoutCall !== null &&
+			specVersion <= this._versionWithoutCall
+		) {
+			return notAvailable;
+		}
+
+		return unknown;
+	}
+
+	public setVersionWithoutCall(specVersion: number) {
```
niklasad1, line 57: I'm not sure that using `set` here is a good idea, as it may not be sent if a lower version already exist.

So perhaps `registerVersionWithoutApi`or something.

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,72 @@
...
+		if (this._versionWithoutCall === null) {
+			this._versionWithoutCall = specVersion;
+		} else if (specVersion > this._versionWithoutCall) {
+			this._versionWithoutCall = specVersion;
+		}
+	}
+
+	public setVersionWithCall(specVersion: number) {
```
niklasad1, line 65: registerWithApi

src/chains-config/cache/QueryFeeDetailsCache.ts
```diff
@@ -0,0 +1,72 @@
...
+	 * queryFeeDetails is not available
+	 */
+	notAvailable = 'notAvailable',
+	/**
+	 * We dont know is queryFeeDetails is available
+	 */
+	unknown = 'unknown',
+}
+
+export class QueryFeeDetailsCache {
```
niklasad1, line 19: this runtime API is called `TransactionPaymentCallApi` but no sure if we really care about `query_call_info` as well

src/types/chains-config/ControllerConfig.ts
```diff
@@ -38,12 +39,24 @@ export interface ControllerConfig {
...
+	 */
+	minCalcFeeRuntime: IOption<number>;
+	/**
+	 * LRU cache that stores the 2 most recent queries.
+	 */
+	blockStore: LRU<string, IBlock>;
+	/**
+	 * Cache for storing runtime versions that either have queryFeeDetails, or dont.
+	 */
+	queryFeeErrCache: QueryFeeDetailsCache;
```
niklasad1, line 61:
```suggestion
	hasQueryFeeApi: QueryFeeDetailsCache;
```

src/chains-config/acalaControllers.ts
```diff
@@ -43,5 +43,6 @@ export const acalaControllers: ControllerConfig = {
 		finalizes: true,
 		minCalcFeeRuntime: 2000,
 		blockStore: initLRUCache(),
+		queryFeeErrCache: new QueryFeeDetailsCache(null, null),
```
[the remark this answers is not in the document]
TarikGul: So the args are here are necessary in the sense that it allows us to pass in known versions so avoid receiving any errors at all. It also extends that same sentiment to parachains that want that same functionality too.

src/chains-config/cache/QueryFeeDetailsCache.ts
TarikGul: Yea we don't care for query_call_info here. Only query_fee_details

TarikGul: We only want to cache the versions of QueryFeeDetails

niklasad1: I was thinking of having some nice abstraction of the `TransactionPaymentApi` but if we only care about `query_call_info` then this is sufficient.

niklasad1 (approved):
looks alright, I think the `QueryFeeDetailsCache` would benefit with some additional docs such that "this is a workaround" because it's not possible to fetch which `runtime APIs` are available.

src/chains-config/cache/QueryFeeDetailsCache.ts
TarikGul: I think the `registerWithoutApi` sounds great. Getting the naming right with these things can feel tough sometimes, but this is a good fit.

niklasad1: yupp, naming is hard :)

TarikGul: I do think we should keep the `set` though after doing some refactoring, as it will keep naming differences between the classes instance variables seperated, but that being said I will clarify what it does with docs.

TarikGul: fixed

src/chains-config/cache/QueryFeeDetailsCache.ts
TarikGul: fixed

src/types/chains-config/ControllerConfig.ts
TarikGul: fixed

jsdw (approved):
Awesome!

PR #1021, opened 2022-08-19
fix: remove deprecated `reset` for lru cache and replace with `clear`

TarikGul:
`reset` was deprecated via 7.0.0 for `lru-cache`, therefore we replaced it with clear. This does not affect sidecar in any breaking way.


niklasad1 (approved):

lexnv (approved):
LGTM!

PR #1023, opened 2022-08-20
feat: validateBooleanMiddleware for controllers

### LINKED ISSUE #398: Boolean query param validation middleware
We should have middleware that validates a specified set of boolean query params. 

Since a number of endpoints use boolean query params this piece of middleware should be re-usable and allow the specification of the name of the params to validate. To achieve this there should be a function that allows specification of the param names and returns the actual middleware. 

The signature could look like 
```javascript
validateBoolQueryParams(...params: string[]): Express.RequestHandler
```
And then the middleware could be mounted directly on the `Express.Router` by doing something like this in the controller
```javascript
protected initRoutes(): void {
	this.router.use(this.path, validateBoolQueryParams('myParamA', 'myParamB'));
	...
}

### PR DESCRIPTION
**Pull request by TarikGul**
### Summary

This adds validation for boolean middleware inputs. It checks specifically for `true`, and `false` for known query params that expect booleans. It will error with a 400 BadRequest. For query params that are not part of the API, as normal we just ignore them as the API should not handle those query params. 

### Extra

I cleaned up the validation tests and utils for both validateBoolean and validateAddress by creating a util folder that handles that functionality. 

closes: https://github.com/paritytech/substrate-api-sidecar/issues/398

### DIFF
--- a/src/middleware/validate/util.ts
+++ b/src/middleware/validate/util.ts
@@ -18,10 +18,10 @@ import { Request, Response } from 'express';
 import { RequestHandler } from 'express-serve-static-core';
 
 /**
- * Assert that a middleware does not error with a the given `req`.
+ * Assert that a middleware does not error with the given request.
  *
- * @param name thing it does not error on
- * @param req Express Request containing thing it errors on
+ * @param name String for tests to log.
+ * @param req Express Request containing thing it errors on.
  */
 export const doesNotErrorWith = (
 	name: string,
@@ -40,9 +40,9 @@ export const doesNotErrorWith = (
  * Assert that a middleware passes `err` to next with the given
  * `req`.
  *
- * @param name thing it errors on
- * @param req Express Request containing thing it errors on
- * @param err expected error that it passes to next
+ * @param name String for tests to log.
+ * @param req Express Request containing thing it errors on.
+ * @param err Expected error that it passes to next.
  */
 export const errorsWith = (
 	name: string,
--- a/src/middleware/validate/util.ts
+++ b/src/middleware/validate/util.ts
@@ -1,0 +1,60 @@
+// Copyright 2017-2022 Parity Technologies (UK) Ltd.
+// This file is part of Substrate API Sidecar.
+//
+// Substrate API Sidecar is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { Request, Response } from 'express';
+import { RequestHandler } from 'express-serve-static-core';
+
+/**
+ * Assert that a middleware does not error with a the given `req`.
+ *
+ * @param name thing it does not error on
+ * @param req Express Request containing thing it errors on
+ */
+export const doesNotErrorWith = (
+	name: string,
+	req: Request,
+	middleware: RequestHandler
+): void => {
+	it(`does not error with ${name}`, () => {
+		const next = jest.fn();
+		middleware(req, null as unknown as Response, next);
+		expect(next).toBeCalledTimes(1);
+		expect(next).toBeCalledWith();
+	});
+};
+
+/**
+ * Assert that a middleware passes `err` to next with the given
+ * `req`.
+ *
+ * @param name thing it errors on
+ * @param req Express Request containing thing it errors on
+ * @param err expected error that it passes to next
+ */
+export const errorsWith = (
+	name: string,
+	req: Request,
+	err: unknown,
+	middleware: RequestHandler
+): void => {
+	it(`errors with ${name}`, () => {
+		const next = jest.fn();
+
+		middleware(req, null as unknown as Response, next);
+		expect(next).toBeCalledTimes(1);
+		expect(next).toBeCalledWith(err);
+	});
+};
--- a/src/middleware/validate/validateBooleanMiddleware.ts
+++ b/src/middleware/validate/validateBooleanMiddleware.ts
@@ -19,16 +19,16 @@ import { BadRequest } from 'http-errors';
 
 // Query Params used in controllers
 const queryParams = [
-	'eventDocs',
-	'extrinsicDocs',
-	'finalized',
-	'denominated',
-	'unclaimedOnly',
-	'actions',
-	'includeFee',
-	'adjustMetadataV13',
-	'currentLeaseHolders',
-	'noMeta',
+	'eventDocs', // BlocksController
+	'extrinsicDocs', // BlocksController
+	'finalized', // BlocksController
+	'denominated', // AccountsBalanceController
+	'unclaimedOnly', // AccountsStakingPayoutsController
+	'actions', // BlocksTraceController
+	'includeFee', // NodeTransactionPoolController
+	'adjustMetadataV13', // PalletsStorageController
+	'currentLeaseHolders', // ParasController
+	'noMeta', // TransactionMaterialController
 ];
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: src/middleware/validate/util.ts [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { Request, Response } from 'express';
+import { RequestHandler } from 'express-serve-static-core';
+
+/**
+ * Assert that a middleware does not error with a the given `req`.
```
niklasad1, line 21:
```suggestion
 * Assert that a middleware does not error with a the given request.
```

src/middleware/validate/util.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param name thing it does not error on
```
niklasad1, line 23: I don't understand the comment `name thing`, it looks like you just moved the code but would be good with some docs that I could understand :)

src/middleware/validate/util.ts
```diff
@@ -0,0 +1,60 @@
...
+		expect(next).toBeCalledTimes(1);
+		expect(next).toBeCalledWith();
+	});
+};
+
+/**
+ * Assert that a middleware passes `err` to next with the given
+ * `req`.
+ *
+ * @param name thing it errors on
```
niklasad1, line 43: same here, `name thing`?

src/middleware/validate/util.ts
lexnv, line 21:
```suggestion
 * Assert that middleware does not error with the given `req`.
```

src/middleware/validate/validateBooleanMiddleware.ts
```diff
@@ -0,0 +1,58 @@
...
+	const queryKeys = Object.keys(req.query);
+	const errQueryParams: string[] = [];
+	queryKeys
+		.filter((queryParam) => queryParams.includes(queryParam))
+		.forEach((queryParam) => {
+			const queryParamVal =
+				typeof req.query[queryParam] === 'string'
+					? (req.query[queryParam] as string).toLowerCase()
+					: '';
+			if (!(queryParamVal === 'true' || queryParamVal === 'false')) {
```
niklasad1, line 44: I guess this is the idiomatic way of checking the queryParams.

niklasad1 (approved):

src/middleware/validate/util.ts
jsdw: `@param name` is the name of the param, and then `thing it does not error on` is the comment, IIRC. It is confusing to see it all together like this though :)

src/middleware/validate/util.ts
jsdw: Also a redundant `a` in `with a the given`

src/middleware/validate/validateBooleanMiddleware.ts
jsdw: I guess it's because all query params are strings.

Maybe worth doing `queryParamVal.toLowerCase()` to support eg TRUE or True? Not too bothered though; nothing wrong with being strict and accepting only `true` or `false`.

In some apps, the mere presence or absense of a query param sets it to true or false, eg `?foo&bar` leads to foo and bar being true. Again, I'm happy if we just want to be stirct and support only `=true` or `=false`; just throwing it out there :)

src/middleware/validate/validateBooleanMiddleware.ts
```diff
@@ -0,0 +1,58 @@
...
+	'extrinsicDocs', // BlocksController
+	'finalized', // BlocksController
+	'denominated', // AccountsBalanceController
+	'unclaimedOnly', // AccountsStakingPayoutsController
+	'actions', // BlocksTraceController
+	'includeFee', // NodeTransactionPoolController
+	'adjustMetadataV13', // PalletsStorageController
+	'currentLeaseHolders', // ParasController
+	'noMeta', // TransactionMaterialController
+];
```
jsdw, line 32: I'm not a huge fan of this sort of "global list". What if "actions" appears in some other thing in the future and is not a boolean, there?

It makes me wonder more about the value of such a middleware?

It also makes me think of npm packages like Runtypes, which allow you to validate all of the type info out of some JS object.

I geuss I'm wondering; what is actually the benefit of such a middleware?

TarikGul: > I'm not a huge fan of this sort of "global list". What if "actions" appears in some other thing in the future and is not a boolean, there?

Im not the biggest fan either but its necessary given the contraints.
So the middleware is only ran if you add it to the endpoint's route itself, so it would have to be explicitly added and would error. It would be very hard to introduce a crossover. 

> It makes me wonder more about the value of such a middleware?

The value of this is centralizing an error for invalid boolean query params. It doesn't do any sanitizing or mutating. Just error handling.

TarikGul: Also this helps with cleaning up a lot of the controllers for error handling.

src/middleware/validate/util.ts
TarikGul: Yea with inline styling it is way more clear, but like @jsdw it is just the parameter it is mentioning. I also cleaned up the docs here to be more clear.

src/middleware/validate/validateBooleanMiddleware.ts
TarikGul: > Maybe worth doing queryParamVal.toLowerCase() to support eg TRUE or True? Not too bothered though; nothing wrong with being strict and accepting only true or false.

I actually already set `queryParamVal` to lower case above.

jsdw: Oh yeah; apologies; I somehow missed that entirely :)

src/middleware/validate/validateBooleanMiddleware.ts
TarikGul: Removed the global list.

src/middleware/validate/validateBooleanMiddleware.ts
```diff
@@ -0,0 +1,53 @@
...
+import { RequestHandler } from 'express';
+import { BadRequest } from 'http-errors';
+
+/**
+ * Validate that the given query params that are expected to be booleans are correct.
+ *
+ * @param queryParams An array of queryParams to check for. These are passed in at the controller level.
+ */
+export const validateBooleanMiddleware = (
+	queryParams: string[]
```
jsdw, line 26: If you wanted to save a few characters you could do `...queryParams: string[]` here (if i recall the spread syntax properly) to avoid needing array parens anywhere... since it's just a manually built list anyway each time.

But not important at all!

src/middleware/validate/validateBooleanMiddleware.ts
```diff
@@ -0,0 +1,53 @@
...
+): RequestHandler => {
+	return (req, _res, next) => {
+		const queryKeys = Object.keys(req.query);
+		const errQueryParams: string[] = [];
+		queryKeys
+			.filter((queryParam) => queryParams.includes(queryParam))
```
jsdw, line 32: Nit; not a big deal but this isn't the most efficient approach (looking over the keys and then for each key looping over the provided params (prolly only one or two).

More efficient would be something like:

```
for (const key of queryParams) {
    const param = req.query[key].toString().toLowerCase();
    if (param === 'true' || param === 'false') {
        // ....
    }
}
```

Ie only check the queryparams we asked to validate and ignore the rest completely :)

jsdw (approved):
Love it; a huge improvement from the global list :)

src/middleware/validate/validateBooleanMiddleware.ts
TarikGul: Ahh yes good catch, will fix that up

src/middleware/validate/validateBooleanMiddleware.ts
TarikGul: I thought about this at first too actually and I think I settled with the explicit array because it makes it more clear.

src/middleware/validate/validateBooleanMiddleware.ts
TarikGul: fixed

Issue #971, opened 2022-06-24
Support for Event::TransactionPaidFee for query param `feeByEvent`


TarikGul:
Once https://github.com/paritytech/substrate-api-sidecar/pull/970 is merged we should add support for also retrieving fees by `Event::TransactionPaidFee`, given [this comment](https://github.com/paritytech/substrate-api-sidecar/issues/756#issuecomment-1165223445) as well we need to be cognizant of the fact that parachains and other substrate-based chains might have slight implementation differences, therefore we should think about adding an `extra` field to the info provided for fee's which would spread all additional information in there.

TarikGul:
After https://github.com/paritytech/substrate-api-sidecar/pull/1017 goes in, then we can add it to the logic for determing the fee's for future runtimes, but anything earlier will have to use fee calculation

PR #1040, opened 2022-09-01
fix: add TransactionPayment::TransactionPaidFee support

### LINKED ISSUE #756: Fetch tx fee from onchain events

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/756 https://github.com/paritytech/substrate-api-sidecar/issues/971

This is a continuation of https://github.com/paritytech/substrate-api-sidecar/pull/1017. For all future blocks that have the event `TransactionPayment::TransactionPaidFee` we will retrieve the fee directly from the event. 

Adds a `fromEvent` dispatchFeeType.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -274,7 +274,7 @@ export class BlocksService extends AbstractService {
 				dispatchFeeType = 'preDispatch';
 			if (transactionPaidFeeEvent) {
 				finalPartialFee = transactionPaidFeeEvent.data[1].toString();
-				dispatchFeeType = 'postDispatch';
+				dispatchFeeType = 'fromEvent';
 			} else {
 				/**
 				 * Call queryFeeDetails. It may not be available in the runtime and will
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -72,6 +72,7 @@ interface FetchBlockOptions {
 enum Event {
 	success = 'ExtrinsicSuccess',
 	failure = 'ExtrinsicFailed',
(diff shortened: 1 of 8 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -264,40 +265,50 @@ export class BlocksService extends AbstractService {
...
-				finalPartialFee = await this.fetchQueryFeeDetails(
-					block.extrinsics[idx].toHex(),
-					previousBlockHash,
-					weightInfo.weight,
-					weight
-				);
-
+			if (transactionPaidFeeEvent) {
+				finalPartialFee = transactionPaidFeeEvent.data[1].toString();
 				dispatchFeeType = 'postDispatch';
```
jsdw, line 277: Wondering whether adding a new `dispatchFeeType` for this (eg `fromEvent`) makes sense?

Why? because the queryFee logic we use is hardcoded to follow the polkadot fee payment approach and may not work in chains that do something different, so adding a new dispatchFeeType allows chains to know whether they can rely on the value or not if they differ.

TarikGul: Ahh that makes complete sense. I think this is valid, and I agree that a `fromEvent` `dispatchFeeType` is a nice addition.

TarikGul: Added `fromEvent` dispatchFeeType

niklasad1 (approved):
LGTM

lexnv (approved):

PR #1044, opened 2022-09-03
perf: rework benchmarks

### LINKED ISSUE #1050: Tracking: Expand benchmarks for all endpoints.
This is a tracking PR to discuss and add information for the current expansion of the benchmarks we have in sidecar. 
https://github.com/paritytech/substrate-api-sidecar/pull/1044 

The goal is to be able to run a set of benchmarks per endpoint seemlessly.

TarikGul:
# Sidecar Benchmarks

Sidecar's benchmarks utilize `lua`, `wrk`, and an archive node (local). This PR aims at completely a few things:

- Providing benchmarks for every available endpoint that can be queried against a `polkadot` archive node.
- Writing a script that would allow simple access to benchmarking for either one or all endpoints.
- Having the script write the results to a `benchmarks.txt` file in the root of the repository.
- Re-activate eslint for the scripts folder (This is what makes the PR seem a little bigger than it should be).

## The script

This script's purpose is to run the benchmarks inside of the `<root>/benchmarks` directory. This script requires you to have `wrk`, `lua`, and a fully synced archive node. Please reference `<root>/benchmarks/README.md` for more information on installation. 

In order to start this script locally, run from the root directory of this repository:

ex1:
```bash
$ yarn
$ yarn bench
```

ex2:
```bash
$ yarn
$ yarn bench --log-level=info --time=30s  
```

ex3:
```
$ yarn
$ yarn bench --log-level=info --time=30s  --endpoint=/accounts/{accountId}/balance-info
```

### Flags

`--ws-url`: The Ws Url to run the benchmarks against. Default is `ws://127.0.0.1:9944`.

`--endpoint`: Run a single benchmark by specificing which endpoint to run. If this is absent it will run all benchmarks.

`--log-level`: The log-level to run the benchmarks in. Defaults to `http`.

`--time`: The amount of time each benchmark should run for. Ex: `1m`, `30s`, `15s`. Default is `1m`.

- [x] Benches for all endpoints.
    - [X] `/accounts/{accountId}/balance-info`
    - [x] `/accounts/{accountId}/vesting-info`
    - [x] `/accounts/{accountId}/validate`
    - [x]  `/accounts/{accountId}/convert`
    - [x] `/accounts/{accountId}/staking-info`
    - [x] `/accounts/{accountId}/staking-payouts`
    - [X] `/blocks/{blockId}`
    - [x] `/pallets/staking/progress`
    - [x] `/pallets/{palletId}/storage`
    - [x] `/pallets/{palletId}/storage/{storageItemId}`
    - [x] `/pallets/nomination-pools/info`
    - [x] `/pallets/nomination-pools/{poolId}`
    - [x] `/pallets/{palletId}/errors`
    - [x] `/pallets/{palletId}/errors/{errorItemId}`
    - [x] `/pallets/staking/validators`
    - [x] `/transaction/material`
    - [x] `/runtime/spec`
    - [x] `/paras`
    - [x] `/paras/leases/current`
    - [x] `/paras/auctions/current`
    - [x] `/paras/crowdloans`
    - [x] `/paras/{paraId}/crowdloan-info`
    - [x] `/paras/{paraId}/lease-info`
    - [x] `/blocks/{blockId}/header`
    - [x] `/blocks/{blockId}/extrinsics/{extrinsicIndex}`
    - [x] `/blocks/head`
    - [x] `/blocks/head/header`
    - [x] `/node/network`
    - [x] `/node/transaction-pool`
    - [x] `/node/version`
- [x] flushed out Lua util module
- [x] README is finished with instructions on running locally in both linux and mac
- [x] Script to run one or all benchmarks

closes: https://github.com/paritytech/substrate-api-sidecar/issues/1050


TarikGul:
@radupopa2010 Hey just wanted to give you a heads up as well, I am completely reworking the benchmarks in this PR. I will ping you when I work through the CI portion which should be sometime this week when I get the rest finished

cc: @alvicsam

scripts/runBenchmarks.ts
```diff
@@ -0,0 +1,224 @@
...
+	writeResultsToFile(results);
+};
+
+const parser = new ArgumentParser();
+
+parser.add_argument('--ws-url', {
+	required: false,
+	nargs: '?',
+	type: checkWsType,
+	default: 'ws://127.0.0.1:9944',
```
IkerAlus, line 186: what would happen if we specify one of our public endpoints here (f.e. "wss://rpc.polkadot.io")?

IkerAlus:
Couple of questions before actually running this:

- The performance of certain endpoints are highly dependent on the current (or at specific timepoint) status of the network (f.e., transaction pool may be empty or full). Will this be somehow captured in the benchmark report?
- Could this be run being connected to a public endpoint node and/or a non-archive node (assuming we omit those endpoints that query the blockchain in the past?

scripts/runBenchmarks.ts
TarikGul: It would use the public endpoint with no problem, but it is highly not recommended to use a public endpoint. 

Why?: Because the benchmarks will send hundreds of thousands of requests, and we don't want to flood it with storage queries.

TarikGul:
> Couple of questions before actually running this:
> 
> * The performance of certain endpoints are highly dependent on the current (or at specific timepoint) status of the network (f.e., transaction pool may be empty or full). Will this be somehow captured in the benchmark report?
> * Could this be run being connected to a public endpoint node and/or a non-archive node (assuming we omit those endpoints that query the blockchain in the past?

> Will this be somehow captured in the benchmark report?

So no because we should only be querying a local archive node, and not a public node. Most of the benchmarks work with historic blocks, there are some exceptions though. If the transaction-pool is full or not doesn't have a huge impact in my experience from running the benchmarks for a while now (But if it does it'll have a small impact as the amount of queries are so large that it will be a small outlier). 

> Could this be run being connected to a public endpoint node and/or a non-archive node (assuming we omit those endpoints that query the blockchain in the past?

As mentioned above too we definitely don't want to benchmark against public live nodes because of the stress the benchmarks would impose upon the network.

marshacb (approved):
running benchmarks worked great. lgtm!

benchmarks/README.md
```diff
@@ -0,0 +1,78 @@
...
+cd wrk
+make
+
+# move the executable to somewhere in your PATH, ex:
+sudo cp wrk /usr/local/bin
+```
+
+### Running Locally
+
+In order to run each benchmark you should `cd` into the appropriate directory you want to run, set the `WRK_TIME_LENGTH` env var (ex: `export WRK_TIME_LENGTH=30s`) and then `sh init.sh`. You are required to have sidecar running, as well as a fully synced archive node.
```
IkerAlus, line 78: I think it would be useful to add to this README:
- the examples 
- the flags

Both of them are already found in the PR description.

IkerAlus (approved):
Ran benchmarks, checked several lightweight endpoints and errors, everything as expected.

Amazing work!

Issue #301, opened 2020-09-21
Feature request: Add validator staking info endpoint


mariopino:
**Proposed Change or Idea**

Thanks , this is a great and very needed tool!

Please, can you consider adding an endpoint with validator staking information:

Something equivalent to this JS api query:

```
const validatorAddresses = await api.query.session.validators()
const validatorStakingInfo = await Promise.all(validatorAddresses.map((authorityId) => api.derive.staking.account(authorityId)))
```

If same endpoint (or different) could include waiting validators that would be great!

**Alternatives**

[Are there alternative approaches or solutions that you’ve considered?]

**Additional Information**

[Add any other details about the feature request here.]

emostov:
Hi - thanks for the proposal!

Is there specific information you are looking for from `api.derive.staking.account`? Is there a specific use case you have?

To outline Sidecar's current offerings that may help:

To get validator addresses you can query: `/pallets/session/storage/validators`.
Additionally `pallets/staking/progress` has the session validator set, among other info pertaining to era, session and election progress.

To retrieve precise payout information for any of those addresses you can query `/accounts/{validatorId}/staking-payout`.
And to get general staking info you can query `/accounts/{validatorId}/staking-info`.

If you have not already, I would recommend checking out the docs here: https://paritytech.github.io/substrate-api-sidecar/dist

CC @joepetrowski

mariopino:
Hi, thanks  for the response! Our use case is to replace costly subscription based connections to Polkadot JS API by light queries to api sidecar when possible :-).

I'm interested in the following data for validators:

- Total, own and other exposures
- Identity
- Commission

With current version, is there a way to get that information?

PD: I get `Cannot GET /pallets/session/storage/validators` using `https://polkadot-api.w3f.community/pallets/session/storage/validators`.

emostov:
@joepetrowski, @dvdplm is this still being considered?

emostov:
Hi @mariopino ,

Are you still interested in this feature? If so can you elaborate further on why api-sidecar is the right place for this feature?

From what I can tell after re-reviewing the requirements you mention and the proposed use case it seems this would just be faster to do as direct RPC calls, like the example you give in your initial comments. Since api-sidecar aims to be stateless, it won't have any advantage on those RPC calls (unless I am missing something?); it would probably just end up slower.

rjarasson-ledger:
Hi @emostov, @joepetrowski

[...]

Issue #1005, opened 2022-08-02
[Feature] Full Validator list endpoint


TarikGul:
### **Description**

Should return a list of all the validators. This requires retrieving all the entries and mapping out the `AccountId` keys from storage. 

```typescript
const validators = await api.query.staking.validators.entries();
const keys = validators.forEach(([key]) => {
  console.log('key arguments:', key.args.map((k) => k.toHuman()));
});
```

Example JSON response from the endpoint:

```
{
    validators: [
        `162DmeD1MHpWQhv8NPz7zbUEfCyTfG4Tuqz7P7852KiY2wXz`,
        `13pYWKctR5s8vQuyZt3pxQXue4SRH9coyAS9S9z5HtogAnhs `,
        ...
    ]
}
```

Imod7:
Another related endpoint to the one mentioned in this issue and maybe interesting to have is one that shows 
**"The list of nominators per validator"**

For example under an endpoint like :
`/pallets/staking/validators/validatorAddress`

This idea comes from seeing related questions in the Substrate Stack Exchange like the ones mentioned below : 
- [How to find all nominators of a validator?](https://substrate.stackexchange.com/questions/4056/polkadot-js-api-how-to-find-all-nominators-of-a-validator)
- [How to query the active nominators of X validator in past eras?](https://substrate.stackexchange.com/questions/2297/how-to-query-the-active-nominators-of-x-validator-in-past-eras)
- [How to get the nominators for a given validator address?](https://substrate.stackexchange.com/questions/4300/how-to-get-the-nominators-for-a-given-validator-address)

PR #1045, opened 2022-09-05
feat: add `/pallets/staking/validators` endpoint

Imod7:
Relates to this Issue #1005 

### Description
Adds `/pallets/staking/validators` endpoint which returns a list of all validators addresses and their corresponding status which can be either `active` or `waiting`. 
* `Active` means that the validator is part of the active set
* `Waiting` means that the validator did not get into the active set this era

## Sample Response

Sample request :
`/pallets/staking/validators`

Sample response :

```
{
  "validators": [
    {
      "address": "HYxzaXwkydDNAy3R8UnCvfprNm74TwwSELpyXSRoqmy9AQY",
      "status": "waiting"
    },
    {
      "address": "DfishveZoxSRNRb8FtyS7ignbw6cr32eCY2w6ctLDRM1NQz",
      "status": "active"
    },
    {
      "address": "HmsTAS1bCtZc9FSq9nqJzZCEkhhSygtXj9TDxNgEWTHnpyQ",
      "status": "active"
    },
    {
      "address": "Efishjj6mrTWVjwFoxtMmHH3mX6tRNnPKsSfQeCLZSv3jAE",
      "status": "waiting"
    },
    {
      "address": "Hn8ELsyUL9UgDZpPX6bE464o4sbi7gqgYEKjntK7xuVzGXH",
      "status": "active"
    },
...
```

## Todos
- [x] Update Docs
- [x] Add Tests


src/controllers/pallets/PalletsStakingValidatorsController.ts
```diff
@@ -0,0 +1,52 @@
...
+		this.safeMountAsyncGetHandlers([['', this.getPalletStakingValidators]]);
+	}
+
+	/**
+	 * Get the progress of the staking pallet system.
+	 *
+	 * @param _req Express Request
+	 * @param res Express Response
+	 */
+	private getPalletStakingValidators: RequestHandler = async (
```
TarikGul, line 43: As chatted about, lets have this accept a query param of `at`

src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -0,0 +1,58 @@
...
+	 * `Active` means that the validator is part of the active set and
+	 * `waiting` means that the validator did not get into the active set this era.
+	 *
+	 * @returns all the validators addresses and their corresponding status
+	 */
+	async derivePalletStakingValidators(): Promise<IPalletStakingValidator> {
+		const { api } = this;
+
+		// Validators in the active set
+		const validatorSession = await api.query.session.validators();
```
TarikGul, line 34: This api should be a result of `api.at(blockHash)` and named `historicApi`

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/301

src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -0,0 +1,63 @@
...
+	async derivePalletStakingValidators(
+		hash: BlockHash
+	): Promise<IPalletStakingValidator> {
+		const { api } = this;
+		const historicApi = await api.at(hash);
+
+		// Validators in the active set
+		const validatorSession = await historicApi.query.session.validators();
+		let validatorSet: string[] = [];
+		validatorSet = validatorSession.map((address) => address.toString());
```
jsdw, line 40: Let's make this an actual `Set`, so something like:

```ts
let validatorSet: Set<string> = new Set();
for (const addres of validatorSession) validatorSet.insert(address.toString());
```

And then below, `validatorSet.contains(address)` instead of `validatorSet.includes(address)`.

(Random note; I think we have ~1000 validators in total)

Have a think about why this is really beneficial :)

src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @returns all the validators addresses and their corresponding status
+	 */
+	async derivePalletStakingValidators(
+		hash: BlockHash
+	): Promise<IPalletStakingValidator> {
+		const { api } = this;
+		const historicApi = await api.at(hash);
+
+		// Querying to get the validators from the active set and populating
```
TarikGul, line 37: I don't think we need these comments. The code here is self documenting.

src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -0,0 +1,62 @@
...
+		for (const address of validatorSession)
+			validatorsActiveSet.add(address.toString());
+
+		// Populating the returned array with the Validator address and its
+		// status. If the address is found in the `validatorsActiveSet` then
+		// status is `active` otherwise is set to `waiting`
+		const validators: IValidator[] = [];
+		const validatorsEntries =
+			await historicApi.query.staking.validators.entries();
+		validatorsEntries.forEach(([key]) => {
```
TarikGul, line 50: This can be rewritten as a `map`. Something like:

`const validators = validatorsEntries.map(([key]) => {...})`

TarikGul (changes requested):
Looks great so far, just some nits, and code changes.

Imod7:
@TarikGul @IkerAlus If validators can change status (go from `active` to `waiting` for example) per `era` then I have to add as `param` in the endpoint also the `era` right ?
From [Polkadot Support](https://support.polkadot.network/support/solutions/articles/65000168050-what-is-an-era-) 
> _An era is a period of time during which there is a specific set of active validators. Each era has 6 epochs (or sessions). During the last epoch, the active set of the next era is elected._

IkerAlus:
@Imod7 Either you add `blockHeight` or `era` as input parameter but not both. There is an unequivocally relationship between a block and an era. 

If the question is what is more useful as input parameter, block heights or eras, I am not sure, I see reasons for both:
- block height is used everywhere in the endpoints and it is easy to find
- era are more related to this endpoint

src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -0,0 +1,61 @@
...
+	 * @param hash `BlockHash` to make call at
+	 */
+	async derivePalletStakingValidators(
+		hash: BlockHash
+	): Promise<IPalletStakingValidator> {
+		const { api } = this;
+		const historicApi = await api.at(hash);
+
+		const validatorSession = await historicApi.query.session.validators();
+		const validatorsActiveSet: Set<string> = new Set();
```
TarikGul, line 38:
```suggestion
		const validatorsActiveSet= new Set<string>();
```

TarikGul (approved):
Just one nit. Great Job! lgtm

marshacb (approved):
looks great to me! once merge conflicts and any remaining nits are addressed its good to go!

PR #1051, opened 2022-09-06
ci: use polkadot node managed by argo-cd

radupopa2010:
This is a CI infrastructure change.
The purpose of this PR is to update the benchmark CI job that uses a polkadot node.
We introduced `argo-cd` as an automation tool to maintain applications deployed on GKE and used a new name for the polkadot node:  [substrate-api-sidecar-benchmark](https://argocd.parity-build.parity.io/applications/substrate-api-sidecar-benchmark?view=tree&orphaned=false&resource=)

This is a job that use the new pod managed by argo: https://gitlab.parity.io/parity/mirrors/substrate-api-sidecar/-/jobs/1818316#L65
There are some [error messages](https://gitlab.parity.io/parity/mirrors/substrate-api-sidecar/-/jobs/1818316#L92)
but I've seen those in other jobs that were using the current polkadot node.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -174,7 +174,7 @@ deploy-production:
   <<:                              *kubernetes-env
   before_script:                   []
   variables:
-    POLKADOT_ADDRESS:              "ws://polkadot-node.nodes:9944"
+    POLKADOT_ADDRESS:              "ws://substrate-api-sidecar-benchmark-node.nodes:9944"
     BENCHMARK_DURATION:            "15m"                                   # Tests duration
     BENCHMARK_THREADS:             4                                       # Total number of threads to use
     BENCHMARK_CONNECTIONS:         12                                      # Total number of HTTP connections to keep open with each thread handling N = connections/threads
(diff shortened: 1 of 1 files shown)


rcny:
It looks like a harmless change indeed, but I consider that receiving `500` in the both cases isn't quite right.
 @TarikGul Could you please clarify what is supposed benchmarking jobs' behavior as for now?

TarikGul:
@rcny Well the behavior (if you mean in terms of our benchmarks) is to track the `/blocks/{blockId}` endpoint and alert each time there is a regression. 

I am drastically changing our benchmarks via https://github.com/paritytech/substrate-api-sidecar/pull/1044
It will have the same concept but extend functionality to all of our endpoints.

radupopa2010:
This is [ a job that failed](https://gitlab.parity.io/parity/mirrors/substrate-api-sidecar/-/jobs/1808918#L61) part of [this previous pipeline](https://gitlab.parity.io/parity/mirrors/substrate-api-sidecar/-/pipelines/212777) which has the same errors like the one for this PR. Like I said before, this PR does not  change the current state of the benchmark job.
@TarikGul  please let us know if it is OK to merge it.

TarikGul (approved):

sergejparity (approved):

PR #1053, opened 2022-09-10
feat!: support for NStorageMaps

### LINKED ISSUE #811: Query support for StorageNMaps
**Proposed Change or Idea**

I am using a StorageNMap with 3 NMapKeys. I am able to query it from polkadot.js.org by using the parameters. I notice that the storage service only seems to support two keys. Does the sidecar api support StorageNMap?

```rust
	type DocumentID = u32;
	type ContentGroupID = Vec<u8>;
	type ContentLabel = Vec<u8>;
	type ContentValue = Vec<u8>;

	pub(super) type Documents<T> = StorageNMap<
		_, // Prefix
		(
			NMapKey<Blake2_128Concat, DocumentID>,
			NMapKey<Blake2_128Concat, ContentGroupID>,
			NMapKey<Twox64Concat, ContentLabel>,
		),
		ContentValue, // value
		ValueQuery,
	>;
```
**Alternatives**
Not yet

**Additional Information**

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/811

## Breaking Change - `/pallets/{palletid}/storage/{storageId}`

The current endpoint accepts two query parameters `key1`, and `key2`. It is now updated to accept an array query param called `keys` that accepts and N amount of keys. ORDER MATTERS. The order in which the keys are inputted is the order they will be inputted into the storage query. 

Example: 
`/pallets/Staking/storage/ErasValidatorPrefs?keys[]=0x00&keys[]=0x01`

The response will no longer have `key1` and `key2` but will now have a `keys` field that returns an array of the keys.. 

Example:
```
{
  "at": {
    "hash": "0x7b713de604a99857f6c25eacc115a4f28d2611a23d9ddff99ab0e4f1c17a8578",
    "height": "789629"
  },
  "pallet": "staking",
  "palletIndex": "15",
  "storageItem": "erasValidatorPrefs",
  "keys": ["0x00", "0x01"],
  "value": ...
}
```

### Query Params

**Removed**:
`key1` and `key2`

**Added** - Order matters
`keys` - Needs to be formatted like: `?keys[]=<key1>&keys[]=<key2>&keys[]=<key3>`

### Response

**Removed**:
`key1` and `key2`

**Added**
`keys` - Array<string>

## Todo

- [x] Docs


src/controllers/pallets/PalletsStorageController.ts
```diff
@@ -48,20 +52,24 @@ export default class PalletsStorageController extends AbstractController<Pallets
...
-			['/:storageItemId', this.getStorageItem],
+			['/:storageItemId', this.getStorageItem as RequestHandler],
 			['/', this.getStorage],
 		]);
 	}
 
-	private getStorageItem: RequestHandler = async (
+	private getStorageItem: RequestHandler<
+		IPalletsStorageParam,
+		unknown,
```
CurlyBracketEffect, line 62: What are these unknown values here?

CurlyBracketEffect (commented):
Like watching a skilled surgeon at work. Just have that one question about the 'unkown' values.

src/controllers/pallets/PalletsStorageController.ts
TarikGul: The `unknown` values are keys we dont need to specify in the RequestHandler. 

For example heres the RequestHandler interface type

```
RequestHandler<P = core.ParamsDictionary, ResBody = any, ReqBody = any, ReqQuery = qs.ParsedQs, Locals extends Record<string, any> = Record<string, any>>
```

So here you can see we dont need to specify any time for ResBody, or ReqBody so we set those to `unknown`

CurlyBracketEffect: Got it TIL🎉

CurlyBracketEffect (approved):

marshacb (approved):
lgtm!

PR #1072, opened 2022-09-30
test(e2e): cleanup timeout ms

TarikGul:
This sets the timeout to a set const, and raises it. Raising the timeout is just in response to westend's responses being a bit a longer than the other chains. This avoids any testing failing for timeouts.

### DIFF
--- a/e2e-tests/runtimeTests.spec.ts
+++ b/e2e-tests/runtimeTests.spec.ts
@@ -35,9 +35,6 @@ const TIMEOUT_MS = 60000;
 const { blocks, accounts, runtime, paras } = endpoints[chain];
 
 describe('Runtime Tests for blocks', () => {
-	/**
-	 * Allows a timeout of 30 seconds for each response.
-	 */
 	jest.setTimeout(TIMEOUT_MS);
 
 	/**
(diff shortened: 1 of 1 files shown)


e2e-tests/runtimeTests.spec.ts
```diff
@@ -81,7 +82,7 @@ describe('Runtime Tests for `/runtime/*`', () => {
 	/**
 	 * Allows a timeout of 30 seconds for each response.
```
niklasad1, line 83: remove these comments or modify the timeout? :)

```suggestion
```

niklasad1 (approved):

e2e-tests/runtimeTests.spec.ts
TarikGul: Thanks for reminding me totally forgot to remove the comments!

TarikGul: fixed

Imod7 (approved):
LGTM! :)

Just a few questions of minor importance :
1. I was wondering if it is worth having different timeouts for westend vs rest of chains.
2. The choice of `60000` for `TIMEOUT_MS` was made based on which criteria ? I imagine the goal was to double the previous value of 30 seconds right ?
Thank you!!!

TarikGul:
> I was wondering if it is worth having different timeouts for westend vs rest of chains.

Unnecessary complexity, they can all be the same MS. The Timeout is there so just in case the request + test doesnt hang. This is a rare case in jest but it can happen. 

> The choice of 60000 for TIMEOUT_MS was made based on which criteria ? I imagine the goal was to double the previous value of 30 seconds right ?

The choice of 60000 was chosen because I have never really seen it go over 40000 before, but wanted to give it an extra buffer.

PR #1102, opened 2022-10-25
ci: increase bench threshold

TarikGul:
This will increase the benchmarks threshold so the repo doesn't get flooded with regression issues.

This will all get solved once https://github.com/paritytech/substrate-api-sidecar/pull/1044 goes in.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -224,7 +224,7 @@ check-benchmark:
   variables:
     GITHUB_REPO:                   "paritytech/substrate-api-sidecar"
     CI_IMAGE:                      "paritytech/benchmarks:latest"
-    THRESHOLD:                     35000
+    THRESHOLD:                     60000
     GITHUB_TOKEN:                  $GITHUB_PR_TOKEN
   script:
     - export RESULT=$(cat artifacts/result.txt | grep AvgRequestTime | awk '{print $2}')


marshacb (approved):
lgtm

niklasad1:
I suggest that we talk to @alvicsam so setup something similar as we have in jsonrpsee instead of having these hardcoded threshold value which isn't scalable.

Should be possible to copy-paste: https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/benchmarks_gitlab.yml
and ask @alvicsam so push the benchmark results to the gh-pages branch. You need to modify GHA for js but should be trivial I think.

The benefit is that you could modify [threshold](https://github.com/benchmark-action/github-action-benchmark#alert-threshold-optional) and get nice graphs published on the gh pages similar to https://paritytech.github.io/jsonrpsee/bench/dev2/

@TarikGul @marshacb thoughts?

niklasad1 (approved):

TarikGul:
> I suggest that we talk to @alvicsam and setup something similar as we have in jsonrpsee instead of having these hardcoded threshold value which isn't scalable.
> 
> Should be possible to copy-paste: https://github.com/paritytech/jsonrpsee/blob/master/.github/workflows/benchmarks_gitlab.yml and ask @alvicsam so push the benchmark results to the gh-pages branch. You need to modify GHA for js but should be trivial I think.
> 
> The benefit is that you could modify [threshold](https://github.com/benchmark-action/github-action-benchmark#alert-threshold-optional) and get nice graphs published on the gh pages similar to https://paritytech.github.io/jsonrpsee/bench/dev2/
> 
> @TarikGul @marshacb thoughts?

I agree, and I think as I pass on this PR to the team https://github.com/paritytech/substrate-api-sidecar/pull/1044, it should be implemented as such in there.

PR #1147, opened 2022-11-24
test(e2e-tests): reorg the e2e-tests to be under a historic folder to…

TarikGul:
… prepare for current tests

With the addition of a e2e-script that is going to test against the most current block, I want to organize and distinguish the e2e-tests from historic tests that expect historical runtime integrity, and present tests that check the integrity of sidecar against a new runtime. 

This PR is up because it has a lot of file name changes so its super large. I didn't want to add any new logic into this PR since it would be hard to review. 

To reiterate all this PR does is change the structure of the e2e tests, it does not change any code (For the most part other than renaming things paths).


marshacb (approved):
lgtm

lexnv (approved):

PR #1166, opened 2022-12-17
feat: add `/paras/head/included-candidates` and `/paras/head/backed-candidates`

TarikGul:
Summary: An endpoint where it returns the following parachains and their current header information. 

NOTE: Naming and descriptions are subject to change. 

## `/paras/head/included-candidates`

**Summary:** Get the heads of the included (backed and considered available) parachain candidates at the specified block height or at the most recent finalized head otherwise.

### Query Params

### Response

```
{
    at: {
        hash: string,
        height: string,
    }
    "<para_id>": {
        "hash": String
        "parentHash": String
        "number": String
        "stateRoot": String
        "extrinsicsRoot": String
         "digest": Digest
    },
    ...
}
```

## `/paras/head/backed-candidates`

**Summary:** Get the heads of the backed parachain candidates at the specified block height or at the most recent finalized head otherwise.

### Query Params

### Response

```
{
    at: {
        hash: string,
        height: string,
    }
    "<para_id>": {
        "hash": String
        "parentHash": String
        "number": String
        "stateRoot": String
        "extrinsicsRoot": String
         "digest": Digest
    },
    ...
}
```

TODO: 
- [x] Tests
- [x] Docs


marshacb (approved):
lgtm

docs/src/openapi-v1.yaml
```diff
@@ -1342,6 +1342,54 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ParasLeaseInfo'
+  /paras/head/included-candidates:
+    get:
+      tags:
+      - paras
+      summary: |
+        Get the heads of all the current included candidate parachains at the most recent finalized head.
```
IkerAlus, line 1350: One little but important detail: I would write here 

"Get the heads of the included (backed and considered available) parachain candidates at the specified block height or at the most recent finalized head otherwise."

With this endpoint we are pulling the heads of the candidates that are included at a given relay chain block, but this does not imply we are retrieving all the included candidates at this height (state info).

I also added more details so that the reader understands the meaning of "included" here and fix it for the optional input `at`.

docs/src/openapi-v1.yaml
```diff
@@ -1342,6 +1342,54 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ParasHeaders'
+  /paras/head/backed-candidates:
+    get:
+      tags:
+      - paras
+      summary: |
+        Get the heads of all the current backed candidate parachains at the most recent finalized head.
```
IkerAlus, line 1374: Similarly as above:

I would write here:

"Get the heads of the backed parachain candidates at the specified block height or at the most recent finalized head otherwise."

With this endpoint we are pulling the heads of the candidates that are backed at a given relay chain block, but this does not imply we are retrieving all the currently backed (but not yet available) candidates at this height.

IkerAlus: Actually, there is a storage entry `paraInclusion.pendingAvailability` which stores exactly this, all the candidates that have been backed so far but they have not been included yet (availability step).

docs/src/openapi-v1.yaml
```diff
@@ -1342,6 +1342,54 @@ paths:
...
+      tags:
+      - paras
+      summary: |
+        Get the heads of all the current included candidate parachains at the most recent finalized head.
+      description: |
+        Returns an object with all the parachain id's as keys, and their headers as values.
+      parameters:
+      - name: at
+        in: query
+        description: Block at which to retrieve para's leases at.
```
IkerAlus, line 1356: "leases" --> "heads"

docs/src/openapi-v1.yaml
```diff
@@ -1342,6 +1342,54 @@ paths:
...
+      tags:
+      - paras
+      summary: |
+        Get the heads of all the current backed candidate parachains at the most recent finalized head.
+      description: |
+        Returns an object with all the parachain id's as keys, and their headers as values.
+      parameters:
+      - name: at
+        in: query
+        description: Block at which to retrieve para's leases at.
```
IkerAlus, line 1380: "leases" --> "heads"

docs/src/openapi-v1.yaml
IkerAlus: This should be updated in the PR description too.

src/services/paras/ParasService.ts
```diff
@@ -429,6 +431,53 @@ export class ParasService extends AbstractService {
 		};
 	}
 
+	/**
+	 * Retrieve all the headers of the parachains connected to the relay chain given
```
IkerAlus, line 435: I would update this comment  as per my comments above. It is not all the parachain heads but those included/backed in the relevant relaychain block

IkerAlus (changes requested):
Consider some comments regarding the descriptions and docs.

The main point is that with these endpoints we are not retrieving all the currently backed and/or included parachain heads but those included in the relevant block.

docs/src/openapi-v1.yaml
TarikGul: Thanks for the input, I updated the description.

docs/src/openapi-v1.yaml
TarikGul: Okay cool, I updated the description. 

> paraInclusion.pendingAvailability

Is this something you would like to see included, maybe in a different endpoint?

IkerAlus: Perhaps this won't be as interesting/useful as the two endpoints here but it is something to consider for sure. Perhaps a lower priority task for the incoming months?

IkerAlus (approved):
The endpoints descriptions have been fixed and everything looks good.

docs/src/openapi-v1.yaml
TarikGul: > Perhaps a lower priority task for the incoming months?

Yea I agree with this. Im happy to tackle this again, as time goes on. Would be good to see if this will have merit among many users.

PR #1168, opened 2022-12-19
fix: use local version of wasm-pack

xlc:
so we can build:calc without having wasm-pack installed globally


TarikGul (approved):
lgtm, thanks so much for the contribution. This is a nice QOL change, and good way to have consistency.

marshacb (approved):

PR #1170, opened 2022-12-21
fix: set rpc to call for queryInfo in transaction/fee-estimate

TarikGul:
Switches the use of RPC calls for queryInfo to runtime calls for `/transaction/fee-estimate`. Will use rpc as a fallback.


marshacb (approved):
looks good

src/services/transaction/TransactionFeeEstimateService.ts
```diff
@@ -32,9 +32,20 @@ export class TransactionFeeEstimateService extends AbstractService {
 		transaction: string
 	): Promise<RuntimeDispatchInfo> {
 		const { api } = this;
+		const apiAt = await api.at(hash);
```
IkerAlus, line 35: for my understanding, what is the difference of pulling the hash at this point and not inside the `try`?

IkerAlus (approved):
LGTM

src/services/transaction/TransactionFeeEstimateService.ts
TarikGul: No big difference, more so for readability. The `try` is really just dealing with the async call whereas establishing the `api.at()` is declaring the runtime api.

PR #1173, opened 2022-12-26
fix: revert polkadot-js deps to 9.10.3

TarikGul:
9.10.4 has an issue where it will fail on historical blocks for RuntimeDispatchInfoV1 so im reverting to 9.10.3. I will follow this up with an issue in polkadot-js/api


marshacb (approved):
lgtm!

Issue #1153, opened 2022-11-28
[Bug] transaction-pool unwrapping undefined.


TarikGul:
While building out the extended e2e-tests, I was consistently getting this error:

```
2022-11-28 16:07:30 error: Cannot read properties of undefined (reading 'unwrap') 
 TypeError: Cannot read properties of undefined (reading 'unwrap')
    at NodeTransactionPoolService.computeExtPriority (/Users/tarikgul/Desktop/parity/tools-team/substrate-api-sidecar/build/src/services/node/NodeTransactionPoolService.js:89:80)
    at NodeTransactionPoolService.extractExtrinsicInfo (/Users/tarikgul/Desktop/parity/tools-team/substrate-api-sidecar/build/src/services/node/NodeTransactionPoolService.js:61:37)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 0)
    at async NodeTransactionPoolService.fetchTransactionPool (/Users/tarikgul/Desktop/parity/tools-team/substrate-api-sidecar/build/src/services/node/NodeTransactionPoolService.js:35:26)
    at async NodeTransactionPoolController.getNodeTransactionPool (/Users/tarikgul/Desktop/parity/tools-team/substrate-api-sidecar/build/src/controllers/node/NodeTransactionPoolController.js:47:62)
    at async /Users/tarikgul/Desktop/parity/tools-team/substrate-api-sidecar/build/src/controllers/AbstractController.js:226:9
```

I was able to reproduce it, and it should be looked into and resolved. 

This involved the following endpoint `/node/transaction-pool?includeFee=true`

PR #1177, opened 2023-01-05
fix: replace rpc calls for fees and give support for weight v1 and v2

TarikGul:
This replaces the rpc calls with runtime calls, and fizes the bug for weights, regarding both weights v1, and weights v2.

closes: https://github.com/paritytech/substrate-api-sidecar/issues/1153

### DIFF
--- a/src/services/test-helpers/mock/mockApi.ts
+++ b/src/services/test-helpers/mock/mockApi.ts
@@ -28,7 +28,6 @@ import {
 	SessionIndex,
 	StakingLedger,
 } from '@polkadot/types/interfaces';
-import {} from '@polkadot/types/types';
 import BN from 'bn.js';
 
 import { polkadotMetadata } from '../../../test-helpers/metadata/metadata';
--- a/src/services/test-helpers/mock/mockApi.ts
+++ b/src/services/test-helpers/mock/mockApi.ts
@@ -15,7 +15,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { ApiPromise } from '@polkadot/api';
-import { Vec } from '@polkadot/types';
+import { GenericExtrinsic, Vec } from '@polkadot/types';
 import { Option } from '@polkadot/types/codec';
 import {
 	AccountId,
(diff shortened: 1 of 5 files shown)


marshacb (approved):
lgtm!

Imod7 (commented):
Shouldn't `rpc.payment` be replaced also in the `.spec` file ? 
- [defaultMockApi.rpc.payment as any).queryInfo](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/node/NodeTransactionPoolService.spec.ts#L102)
- [(defaultMockApi.rpc.payment as any).queryInfo](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/node/NodeTransactionPoolService.spec.ts#L111)

TarikGul:
> Shouldn't `rpc.payment` be replaced also in the `.spec` file ?
> 
> * [defaultMockApi.rpc.payment as any).queryInfo](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/node/NodeTransactionPoolService.spec.ts#L102)
> * [(defaultMockApi.rpc.payment as any).queryInfo](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/node/NodeTransactionPoolService.spec.ts#L111)

Yup, great catch. It can just be removed, as I replaced the actual mockHistoricApi with the call instead.

TarikGul:
@Imod7 Should be all good now. I also specified the test-helpers better in https://github.com/paritytech/substrate-api-sidecar/pull/1177/commits/79c6e63d2348fcf9688d202d833032b936596c5c

src/services/test-helpers/mock/mockApi.ts
```diff
@@ -28,6 +28,7 @@ import {
 	SessionIndex,
 	StakingLedger,
 } from '@polkadot/types/interfaces';
+import {} from '@polkadot/types/types';
```
Imod7, line 31: If something is used from `@polkadot/types/types`, maybe we should explicitly import that type ? Btw, I comment this line and I do not see any errors but maybe I am missing something.

Imod7 (approved):
LGTM

src/services/test-helpers/mock/mockApi.ts
TarikGul: Awesome fixed, thanks so much for that last catch!

Issue #1188, opened 2023-01-19
[Feature]: Add a FileTransport for local logging options to help debug external user issues


TarikGul:
There should be a `FileTransport` added to the winston logging. There should be 2 options, one for the path to logging, and another whether or not logs should be active. The options might needed to be threaded down from where the logger is initiated.

PR #1189, opened 2023-01-19
feat: add fileTransport

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/1188

### Summary

Give users the ability to write logs to a file for better debugging etc. 

### Env Vars

- `SAS_LOG_WRITE`: Whether or not to write logs to a log file. Default is set to `false`. Accepts a boolean value. The log files will be written as `logs.log`. **NOTE**: Tt will only log what is available depending on what `SAS_LOG_LEVEL` is set to.
- `SAS_LOG_WRITE_PATH`: Specifies the path to write the log files. Default will be where the package is installed.
- `SAS_LOG_WRITE_MAX_FILE_SIZE`: Specifies in bytes what the max file size for the written log files should be. Default is `5242880` (5MB).
- `SAS_LOG_WRITE_MAX_FILES`: Specifies how many files can be written. Default is 5.


marshacb (approved):
lgtm

marshacb (approved):
looking good to me!

Imod7 (approved):
Looks great! 

I noticed the following on how the logs are saved in the file(s) : 
- Whenever the max size of file and max file(s) are reached, the logs are re-written.
- However whenever Sidecar is started again, the logs are appended in the file(s) until the max size file and max files are reached again.

If we would like the log files to be re-written also every time Sidecar starts/runs then we only need to add this [line](https://github.com/winstonjs/winston/issues/1271#issuecomment-384194787) : 
`options: { flags: 'w' }`
when we return the `transports.File`.

PR #1203, opened 2023-01-25
fix: rename nomination-pools to be to standard

TarikGul:
This fixes the naming of the new feature that came with v14.3.0 ie `/pallets/nominationPools/*` -> `/pallets/nomination-pools/*`. 

Since the new feature has only been out for less than a day, I think it's fitting to only make this a patch.


marshacb (approved):
lgtm!

CurlyBracketEffect (approved):
👍

PR #1205, opened 2023-01-26
fix: update errors endpoint to use latest error metadata for fetchErrorItem

marshacb:


TarikGul:
Whats the exact issue happening here with the types?

marshacb:
> Whats the exact issue happening here with the types?

There isn't an issue with these types remaining as is actually. It would just be a change to keep the return types of the errors endpoints more consistent with each other. It doesn't change any logic for the api so if preferred we can leave this unchanged.

TarikGul (approved):

Imod7 (commented):
Looks great! 
- Just to double check we still need the `| Option<PalletErrorMetadataV14> {` in [line 43](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/AbstractPalletsService.ts#L43) (`src/services/AbstractPalletsService.ts` file) right ?
- On that part of code why in [line 51](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/AbstractPalletsService.ts#L51) we hardcode 'errors' ? 
  ```
  return this.getProperty(meta as PalletMetadataV14, 'errors');
  ```
  We could again just use `metadataFieldType` right ? So maybe have 
   ```
   return this.getProperty(meta as PalletMetadataV14, metadataFieldType);
   ``` 
- One last question : why we did not use `ErrorMetadataLatest` since the beginning ?

marshacb:
> Looks great!
> 
> * Just to double check we still need the `| Option<PalletErrorMetadataV14> {` in [line 43](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/AbstractPalletsService.ts#L43) (`src/services/AbstractPalletsService.ts` file) right ?
> * On that part of code why in [line 51](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/AbstractPalletsService.ts#L51) we hardcode 'errors' ?
>   ```
>   return this.getProperty(meta as PalletMetadataV14, 'errors');
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   We could again just use `metadataFieldType` right ? So maybe have
>   ```
>   return this.getProperty(meta as PalletMetadataV14, metadataFieldType);
>   ```
> * One last question : why we did not use `ErrorMetadataLatest` since the beginning ?

Yes, we still need the reference to `Option<PalletErrorMetadataV14>` in order to determine the pallets metadata type when we are looking for the `errors` of the pallet metadata.

The `errors` in the `else` case has to be hardcoded because unlike the previous `if` check for `storage`, its value cannot be inferred based on its context. `getProperty` knows the value of `metadataFieldType` inside of the storage `if` check because it is being used within the `if` blocks context where it has already been determined to equal `storage`.

`ErrorMetadataLatest` wasn't used for the `fetchErrorItem` call because `PalletErrorMetadataV14` was an acceptable type to return. This simply makes the return types for both endpoints consistent and will likely be similar in future endpoints such as `events`, `constants` and `calls`/`dispatchables`.

Imod7 (approved):
Thank you for the explanations! 💯

PR #1206, opened 2023-01-30
chore(deps): update polkadot-js deps

TarikGul:
Now with the newest version of Polkadot-js, the Weight types are more narrowed and specified, we have to account for WeightsV1 and WeightsV2 when parsing the data. This PR resolves that while updating to the latest polkadot-js deps


CurlyBracketEffect (approved):
👍

PR #1210, opened 2023-01-31
feat: add pallets/consts endpoint

marshacb:
Summary: An endpoint that returns the constants for a given pallet

### Query Params

- `onlyIds`: boolean to return only the Id of each constant instead of the entirety of each constant

Sample response for /pallets/democracy/consts?onlyIds=true

```
{
	"at": {
		"hash": "0xf178fd381cbfbea8d3df1c6579cc918eb1dedeee6281cf3ee006a40e0139e469",
		"height": "14046577"
	},
	"pallet": "democracy",
	"palletIndex": "14",
	"items": [
		"EnactmentPeriod",
		"LaunchPeriod",
		"VotingPeriod",
		"VoteLockingPeriod",
		"MinimumDeposit",
		"InstantAllowed",
		"FastTrackVotingPeriod",
		"CooloffPeriod",
		"MaxVotes",
		"MaxProposals",
		"MaxDeposits",
		"MaxBlacklisted"
	]
}
```

### Query Params

- `metadata`: boolean to choose whether to include an consts' metadata in the response

Sample response for /pallets/democracy/consts/InstantAllowed

```
{
	"at": {
		"hash": "0xd4b121ad968b133091b837292d7a73d77234b07be0073ad5b97a95b192d8ae3f",
		"height": "14046588"
	},
	"pallet": "democracy",
	"palletIndex": "14",
	"constantsItem": "instantAllowed"
}
```


src/services/test-helpers/responses/pallets/fetchEnactmentPeriodConstsItem789629.json
```diff
@@ -0,0 +1,9 @@
+{
+	"at": {
+		"hash": "0x18707858d4a24cf7235d4e1b45ab1665e61050d01a8a01397f3423ffd6937655",
+		"height": "789629"
+	},
+	"pallet": "democracy",
+	"palletIndex": "14",
+	"constantsItem": "EnactmentPeriod"
+}
```
TarikGul, line 9: Newline

src/services/test-helpers/responses/pallets/fetchVotingPeriodConstsItem789629.json
```diff
@@ -0,0 +1,17 @@
...
+	"constantsItem": "VotingPeriod",
+	"metadata": {
+		"name": "VotingPeriod",
+		"type": "4",
+		"value": "0x00270600",
+		"docs": [
+			" How often (in blocks) to check for new votes."
+		]
+	}
+}
```
TarikGul, line 17: new line

src/services/AbstractPalletsService.ts
```diff
@@ -79,7 +91,7 @@ export abstract class AbstractPalletsService extends AbstractService {
 				metadataFieldType
 			);
 			filtered = pallets.filter(
-				(mod) => (mod[metadataFieldType] as typeof palletMetaType).isSome
+				(mod) => !(mod[metadataFieldType] as typeof palletMetaType).isEmpty
```
TarikGul, line 94: Whats the motivation behind this change? It's the same?

TarikGul (approved):
👍

src/services/AbstractPalletsService.ts
marshacb: I changed this to `isEmpty` because the outer type of the `constants` field on the `PalletMetadataV14` interface is a `Vec` (`Vec<PalletConstantMetadataV14>`) and not an `Option` like in the other cases that are handled(e.g. `Option<PalletStorageMetadataV14>`).  The property `isSome` doesn't exist for the `Vec` but `isEmpty` exists for both the `Option` and `Vec` types and checks if either of them are empty.

TarikGul: Ahh that makes total sense thanks for the clarification

src/services/AbstractPalletsService.ts
```diff
@@ -234,13 +238,19 @@ export abstract class AbstractPalletsService extends AbstractService {
...
 			);
-		} else {
+		} else if (metadataFieldType === 'events') {
 			[palletItemIdx, palletItemMeta] = this.getEventItemMeta(
 				historicApi,
 				palletMeta as PalletMetadataV14,
 				palletItemIdx,
 				palletItemId
 			);
+		} else {
```
Imod7, line 248: I see that this `if...else` statement keeps growing and I was wondering if it would a good idea to replace this conditionals with a lookup table or an object lookup ? since the only thing that changes is the name of the function called. 
I was checking some articles to find as references : 
- [js patterns - lookup tables](https://omwri.medium.com/javascript-patterns-lookup-tables-26bbaf693e24)
- [js - objects for lookups](https://levelup.gitconnected.com/javascript-algorithm-using-objects-for-lookups-d28f34a59504)
- [js ts code - lookups](https://dev.to/k_penguin_sato/use-lookup-tables-for-cleaning-up-your-js-ts-code-9gk)

Maybe it makes it also more readable and maintainable ? Just an idea because I used that in C in the past and I think it is quite fast but I am not sure in Typescript (performance wise).

TarikGul: I don't think this is necessary. A lookup table would require a few abstractions, and performance isn't really going to get affected here. `if...else` is acceptable IMO. We know the fixed size of this if else so I don't think its going to grow much more. 

That being said the only other option I would see for readability is a `switch` statement.

Imod7 (approved):
Looks Great!

PR #1222, opened 2023-02-09
ci: add e2e tests to ci

alvicsam:
PR adds e2e tests to CI

Close https://github.com/paritytech/ci_cd/issues/734

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,6 +36,7 @@ default:
       fi
 
 stages:
+  - test
   - build
   - deploy-staging
   - deploy-production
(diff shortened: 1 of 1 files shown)


alvicsam:
It seems that [regex](https://github.com/paritytech/substrate-api-sidecar/blob/master/scripts/e2eHelpers.ts#L84) doesn't allow current URLs

TarikGul:
@alvicsam I just fixed some issues and merged in the PR, so if you pull in master it will sort some of the issues out. 

1. added the tests for kusama and westend.

2. Removed the check for the ws url. Not necessary since we are using this internally. This also means you can remove the port numbers on the ws urls.

rcny (approved):

TarikGul (approved):
Looks great, thanks :)

PR #1223, opened 2023-02-09
tests: fix e2e tests, add kusama westend, and remove ws check

TarikGul:
Remove `checkWsUrl`,

Add config for kusama, and westend.

rel: https://github.com/paritytech/substrate-api-sidecar/pull/1222


marshacb (approved):
looks great!

PR #1226, opened 2023-02-12
chore: add e2e-tests to single lint configuration

TarikGul:
This PR:

- Adds the e2e-tests to be part of the single lint config.
- Cleans up the package.json
- Lints the previously non linted e2e files. 

Now moving forward all ts, and js files will be linted using the same command.


Imod7 (approved):

marshacb (approved):
looks good

Issue #1008, opened 2022-08-04
Prometheus Metrics


dtp263:
**Proposed Change or Idea**

Add prometheus metrics for monitoring the health of a sidecar instance.

**Additional Information**

In order to provide a clean consistent service you need to be able to easily monitor the health and status of each service.

TarikGul:
It's definitely possible to integrate prometheus into Sidecar as we have the helm setup already intrinsic to the Repo. I am not totally sure on what the workload on integration would look like though - @ArshamTeymouri Any thoughts on this?

ArshamTeymouri:
Hi Tarik!

That's a great idea to have the metrics exposed by the application. For this, we need a new endpoint on the application, say `/metrics`, that exposes application details in a Prometheus-compatible way. Please check [Prometheus docs](https://prometheus.io/docs/instrumenting/clientlibs/) for more details about how to integrate those metrics in your application.

Besides, two additional endpoints should also be introduced, `/readiness` and `liveness`. Those endpoints will expose the general health of the application. e.g. If the app is crashing, the `/liveness` should return a non-200 HTTP response and if the application is working fine and is just waiting for its backing database to sync, `/readiness` should be `non-200`. This way Kubernetes knows whether it needs to restart the pod(application) because of a crash or wait until the application becomes ready to serve requests.

We have already done the same thing for one of the internal tools, Gitspiegel. You can reference its corresponding issue and PR for more details: https://github.com/paritytech/gitspiegel/issues/84

Once those endpoints are in place, I can proceed with configuring Prometheus to scrape the metrics and store them in its database. Later, you would be able to create your own dashboards on Grafana and visualize your application performance with fancy panels.

Feel free to reach me on Matrix if you need anything.

dtp263:
@ArshamTeymouri I love the idea of adding the `readiness` and `liveness` endpoints. But I wonder if it would make more sense to do that in a separate effort.

Implementing and exposing the basic metrics at `/metrics` seems like a nice objective block of work. Determining `liveness` / `readiness` could be a touch more subjective. Opening up the possibility of one blocking the other.

Imod7:
[...]

PR #1232, opened 2023-02-15
feat: add prometheus metrics in dedicated port

Imod7:
### Description
Relates to the issue https://github.com/paritytech/substrate-api-sidecar/issues/1008

This PR is a first approach/test on adding prometheus metrics in Sidecar. More specifically this PR : 
- Uses [prom-client](https://github.com/siimon/prom-client)
- Runs a separate express instance and in a different port (`9100`) where the Prometheus metrics are exposed 
- Exposes the default Prometheus metrics
- Creates and exposes a custom metric of type `counter` that increments in every http error
- All metrics are prefixed with `sas_` that denotes that are "Substrate Api Sidecar" related metrics.

### Run the Prometheus metrics endpoint
- Checkout in the current branch
- prerequisite to have `prom-client` installed (you can with `yarn add prom-client`)
   _note: if you get an error while adding `prom-client` try to do a `yarn dedupe` first_
- `yarn build`
- You can start Sidecar & enable the metrics endpoint by running : 
    - `yarn start --prometheus`
- Or you can start Sidecar, enable the metrics endpoint & define a custom port for the metrics endpoint by running : 
    - `yarn start --prometheus --prometheus-port=<YOUR_CUSTOM_PORT>`
- Open Sidecar on the usual `http://127.0.0.1:8080` 
- Open Metrics endpoint : 
    - on the default port : `http://127.0.0.1:9100/metrics` or 
    - on your custom port if you defined one : `http://127.0.0.1:<YOUR_CUSTOM_PORT>/metrics`
- On the metrics endpoint you should see
    ```
    # HELP sas_http_errors Number of HTTP Errors
    # TYPE sas_http_errors counter
    sas_http_errors 0
    
    # HELP sas_process_cpu_user_seconds_total Total user CPU time spent in seconds.
    # TYPE sas_process_cpu_user_seconds_total counter
    sas_process_cpu_user_seconds_total 1.335734
    
    # HELP sas_process_cpu_system_seconds_total Total system CPU time spent in seconds.
    # TYPE sas_process_cpu_system_seconds_total counter
    sas_process_cpu_system_seconds_total 0.146793
    ...
    ...
    ```
   The first 3 lines denote the custom metric that we defined. The rest of the lines describe the default prometheus metrics.

### Testing
- If we request `http://127.0.0.1:8080/blocks` this will trigger an `httpError`
- This means that our custom metric `sas_http_errors` (of type `counter`) should be increased by 1 so we should see on the metrics endpoint the following : 
```
# HELP sas_http_errors Number of HTTP Errors
# TYPE sas_http_errors counter
sas_http_errors 1
```

### Additional Notes
Apart `counters` there are other types of metrics that can be added like Histograms, Gauges, etc. More details on metric types can be found in : 
- [prom-client - Custom Metrics](https://github.com/siimon/prom-client#custom-metrics)
- [Prometheus Docs - Metric Types](https://prometheus.io/docs/concepts/metric_types/#metric-types)


src/util/metrics.ts
```diff
@@ -0,0 +1,59 @@
...
+			);
+		});
+	}
+
+	/**
+	 * Mount the metrics endpoint.
+	 */
+	private metricsEndpoint() {
+		// Set up the metrics endpoint
+		this.app.get('/metrics', async (_req: Request, res: Response) => {
```
CurlyBracketEffect, line 54: Looks like you just need to add the expected return value here.

Imod7: Thanx for the review @CurlyBracketEffect ! 🙏 I added a return value but then I had another lint error. 

What I did now to remove the following lint errors
```
Error: 54:28 error Promise returned in function argument where a void return was expected @typescript-eslint/no-misused-promises
```
and
```
60:4  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```
 was to : 
- use an async [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) inside the callback
- add the `void` operator 

and it works but I am not sure if this is a good practice though 🤔

src/main.ts
```diff
@@ -88,8 +89,19 @@ async function main() {
 		host: config.EXPRESS.HOST,
 	});
 
+	// Create our App
+	const metricsApp = new Metrics_App({
```
TarikGul, line 93: This should be inside of the if statement below.

src/middleware/error/httpErrorMiddleware.ts
```diff
@@ -47,6 +48,8 @@ export const httpErrorMiddleware: ErrorRequestHandler = (
 		stack: err.stack,
 	};
 
+	httpErrorCounter.inc();
```
TarikGul, line 51: Should this be something active when prometheus is only active as well?

src/util/metrics.ts
```diff
@@ -0,0 +1,64 @@
+import express from 'express';
+import { Application, Request, Response } from 'express';
+import client from 'prom-client';
+
+import { Log } from '../logging/Log';
+import { parseArgs } from '../parseArgs';
+
+const { logger } = Log;
+
+const register = new client.Registry();
```
TarikGul, line 10: These top level calls should be encapsulated in a function and called.  Meaning they should sit in global scope.

src/util/metrics.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const httpErrorCounter = new client.Counter({
+	name: 'sas_http_errors',
+	help: 'Number of HTTP Errors',
+});
+
+register.registerMetric(httpErrorCounter);
```
TarikGul, line 17: Same with these two calls.

TarikGul (changes requested):
Overall, really good job! Just some nits on global scope, and only have things ran if prometheus is active.

src/util/metrics.ts
Imod7: I moved the `register...` under `metricsEndpoint`. Do I have to move also the `httpErrorCounter` ? Since this is a `const` that is exported so I should not move it right?

Imod7 (commented):
I have a question about the help page/message :

- If I run `yarn start -h` it shows me the prometheus options that I added so I get the output I expect : 
  ```
  usage: main.js [-h] [-v] [-p] [-pp PROMETHEUS_PORT]
  
  optional arguments:
    -h, --help            show this help message and exit
    -v, --version         print substrate-api-sidecar version
    -p, --prometheus      enable the prometheus metrics endpoint
    -pp PROMETHEUS_PORT, --prometheus-port PROMETHEUS_PORT
                          specify the port number on which the prometheus metrics are exposed [default: 9100]
  
  ```

- If I run `substrate-api-sidecar -h` I only get the help and version options. It probably does **not** get my local version/folder/branch of sidecar ? thats why ?

src/util/metrics.ts
```diff
@@ -0,0 +1,60 @@
...
+
+		this.port = Number(args.prometheus_port);
+		this.app = express();
+		this.host = host;
+
+		this.metricsEndpoint();
+	}
+
+	listen(): void {
+		const { logger } = Log;
```
TarikGul, line 37: This is fine with the Logger since it's only being used in here which makes sense. But that being said as a side note, the logger can be a top level variable since it's only ever instantiated once. The logger file itself has a static `this.create` which only gets called the first time the logger is instantiated in the entry file.

TarikGul (approved):

marshacb (approved):

PR #1240, opened 2023-02-27
test: fix benchmark url calls for pallets keys

TarikGul:
This fixes some of the benchmark calls to use the proper keys for pallets storage endpoint


marshacb (approved):

CurlyBracketEffect (approved):
LGTM

PR #1255, opened 2023-03-27
fix!: update polkadot-js, and adjust for breaking changes (`/accounts/{accountId}/balance-info`)

TarikGul:
## Breaking Changes

The following PR introduces some minor breaking changes to `/accounts/{accountId}/balance-info` but breaking none the less.

### Summary

With https://github.com/paritytech/substrate/pull/12951 merged we now need to take into account that the AccountInfo provided from `api.query.system.account` will not return the same format. Previously both `miscFrozen` and `feeFrozen` were returned whereas now only `frozen` is returned. In order to keep historical integrity we have kept each field, and added `frozen` as a new key. This breaking change being, when a specific key is not supported in the runtime instead of a balance being inputted into the key, it will have a message like `'miscFrozen does not exist for this runtime'`. A sample return value would be:

```json
{
  "at": {
    "hash": "0x7b713de604a99857f6c25eacc115a4f28d2611a23d9ddff99ab0e4f1c17a8578",
    "height": "789629"
  },
  "nonce": "6",
  "tokenSymbol": "DOT",
  "free": "501090793179",
  "reserved": "0",
  "miscFrozen": "miscFrozen does not exist for this runtime",
  "feeFrozen": "feeFrozen does not exist for this runtime",
  "frozen": "100000000000",
  "locks": [
    {
      "id": "0x7374616b696e6720",
      "amount": "100000000000",
      "reasons": "All"
    }
  ]
}

```

The new return type looks like:

```typescript
export interface IAccountBalanceInfo {
	at: IAt;
	tokenSymbol: string;
	nonce: Index;
	free: Balance | string;
	reserved: Balance | string;
	miscFrozen: Balance | string; // Note this can now return a not supported message
	feeFrozen: Balance | string; // Note this can now return a not supported message
	frozen: Balance | string; // Note this can now return a not supported message
	locks: Vec<BalanceLock> | IBalanceLock[];
}
```

### Additional Notes:

The historical e2e-tests were all updated.

Unit tests are tracked in this issue: https://github.com/paritytech/substrate-api-sidecar/issues/1262

## Additional Changes:

There was also a local type change from `PolkadotPrimitivesV2CandidateReceipt` to `PolkadotPrimitivesV4CandidateReceipt`. This change does not impact any users.


docs/src/openapi-v1.yaml
```diff
@@ -2096,12 +2096,21 @@ components:
...
           type: string
           description: The amount that `free` may not drop below when withdrawing
-            specifically for transaction fee payment.
+            specifically for transaction fee payment. Note, that some runtimes may not have
+            support for feeFrozen and if so the following will be returned `feeFrozen does not exist for this runtime`
+          format: unsignedInteger
+        frozen:
+          type: string
+          description: The amount that `free` may not drop below when reducing the balance, except for actions
+            where the account owner cannot reasonably benefit from thr balance reduction, such as slashing.
```
marshacb, line 2111:
```suggestion
            where the account owner cannot reasonably benefit from the balance reduction, such as slashing.
```

marshacb (approved):
just a minor spelling nit. Looks great to me!

CurlyBracketEffect (approved):
👀 🤯
👍

PR #1258, opened 2023-04-06
fix: validator staking endpoint with more info, and correct docs.

TarikGul:
This adds a field called `validatorsToBeChilled` in the response, it also the `at` field which was missing. I also corrected the docs, and cleaned up the code. `validatorsToBeChilled` has accounts that may be in the `validators` field and I made sure to have that in the docs.

### Response

```
{
  "at": {
    "hash": "string",
    "height": "string"
  },
  "validators": [
    {
      "address": "string",
      "status": "string"
    }
  ],
  "validatorsToBeChilled": [
    {
      "address": "string",
      "status": "string"
    }
  ]
}
```

Thanks to @Ank4n for info and clarification on the Storage query calls.


IkerAlus (approved):
LGTM, I checked in detail the new doc and looks on point. Good work.

Imod7 (approved):
LGTM!

marshacb (approved):

Issue #1268, opened 2023-04-21
[CI]: Change benchmarking chain url's off of testing nodes


TarikGul:
Ping: @alvicsam

For the new benchmarks we've been building and the CI you've been building for it, it seems that the node address [here](https://github.com/paritytech/substrate-api-sidecar/blob/master/.gitlab-ci.yml#L232) is the same as the nodes we use for e2e-testing which those are on the RC versions. Is it possible for us to change those to a node that is using a stable runtime version?

PR #1270, opened 2023-04-24
[ci] Change endpoints for e2e tests

alvicsam:
Use stable endpoints for e2e tests

Close https://github.com/paritytech/substrate-api-sidecar/issues/1268

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -115,7 +115,7 @@ stages:
   variables:
     CI_IMAGE:                      $NODE_IMAGE
     CHAIN_NAME:                    ""
-    CHAIN_URL:                     "wss://${CHAIN_NAME}-testing-rpc.parity-build.parity.io"
+    CHAIN_URL:                     ""
   script:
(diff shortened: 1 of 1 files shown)


alvicsam:
Hm, all tests have this error message:

```bash
2023-03-15 15:58:09 warn: `noMeta` query param will be deprecated in sidecar v13, and replaced with `metadata` please migrate
2023-03-15 15:58:09 error: RPC-CORE: peers(): Vec<PeerInfo>:: -32601: RPC call is unsafe to be called externally
```

but return code is 0 so CI thinks that tests are successful. @TarikGul is it planned?

TarikGul:
@alvicsam Yea those are expected, the first one is emitted via sidecar and is totally fine. The second one is emitted via polkadot-js and totally fine as well. I believe its's when we use this call specifically: `rpc.system.localPeerId()`

TarikGul (approved):
LGTM, thanks :)

Imod7 (approved):

PR #1273, opened 2023-04-26
test: add acala & karura to e2e tests and some improvements

xlc:
- Added `ts-node` to dev deps so I can run it with `yarn ts-node src/main.ts` and run e2e tests with `yarn ts-node e2e-tests/latest/index.ts`. There is no reason to build before run if we can just run it.
- Updated the e2e tests output so it is less confusing.
- Some code refactoring
- Added Acala & Karura e2e tests
- Allow config sidecar url for e2e test
- Fix exit code

### DIFF
--- a/e2e-tests/latest/endpoints/acala.ts
+++ b/e2e-tests/latest/endpoints/acala.ts
@@ -75,6 +75,6 @@ export const acala: IConfig = {
 	},
 	'/transaction/material': {
 		path: '/transaction/material',
-		queryParams: ['noMeta=true'],
+		queryParams: [],
 	},
 };
--- a/e2e-tests/latest/endpoints/acala.ts
+++ b/e2e-tests/latest/endpoints/acala.ts
@@ -18,7 +18,7 @@ import { IConfig } from '../types/endpoints';
 
 export const acala: IConfig = {
 	'/blocks': {
-		path: '/blocks?range=1-5',
+		path: '/blocks?range=3000000-3000005',
 		queryParams: [],
 	},
 	'/blocks/{blockId}': {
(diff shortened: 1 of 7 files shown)


TarikGul:
@xlc Thanks so much for the refactoring and contribution, lgtm!

> Added ts-node to dev deps so I can run it with yarn ts-node src/main.ts and run e2e tests with yarn ts-node e2e-tests/latest/index.ts. There is no reason to build before run if we can just run it.

Yea totally agree with this, its a nice addition.

TarikGul (approved):

marshacb (approved):
nice. lgtm

e2e-tests/latest/endpoints/acala.ts
```diff
@@ -0,0 +1,80 @@
...
+		path: '/runtime/code',
+		queryParams: ['at={blockId}'],
+	},
+	'/runtime/spec': {
+		path: '/runtime/spec',
+		queryParams: ['at={blockId}'],
+	},
+	'/transaction/material': {
+		path: '/transaction/material',
+		queryParams: ['noMeta=true'],
```
TarikGul, line 78:
```suggestion
		queryParams: [],
```

I would say this one param can be removed as I will follow up with a PR today to completely remove it. The default behavior when no param is inputted will be metadata will not be queried.

xlc:
FYI this is the script we using to test sidecar compatibility https://github.com/AcalaNetwork/chopsticks/blob/master/packages/e2e/scripts/sidecar-chopsticks.sh

TarikGul:
> FYI this is the script we using to test sidecar compatibility https://github.com/AcalaNetwork/chopsticks/blob/master/packages/e2e/scripts/sidecar-chopsticks.sh

Ahh super cool! Yea we could probably update a lot of scripts to also be just bash. It would be be a lot cleaner.

Issue #1274, opened 2023-04-27
there is no migration path for noMeta migration


xlc:
When `noMeta` is used, it will complain and ask for migration. But migrate to what?

TarikGul:
This is old legacy code from Pre V14 metadata. It was replaced with the `metadata` query param. Its long overdue for completely removing `noMeta`. Thanks for bringing this up, as its been lost in the queue for a while. I'll get this in tomorrow to finally finish this. 

The expected behavior will be: `metadata` query param takes in two values `scale` or `json` for the format of the metadata, and if the query param is not inputted the field wont be there. 

The controller code can be seen [here](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/controllers/transaction/TransactionMaterialController.ts#L94-L130).

PR #1275, opened 2023-04-27
fix!: remove `noMeta` query param from `/transaction/materials`

TarikGul:
## Summary

This is a long overdue change where `noMeta` is removed from `/transaction/materials`. It will be fully deprecated.

### Changes

`noMeta` removed.

`metadata` is the source of truth for the metadata key in the response. When the metadata field is not present, no metadata will be returned. There are two options for the metadata query param. 

1. `json` - decoded metadata in json format.
2. `scale` - scale encoded metadata format.

closes: https://github.com/paritytech/substrate-api-sidecar/issues/1274


marshacb (approved):
lgtm!

PR #1280, opened 2023-05-03
fix: `/accounts/{accountId}/convert` encoding for ecdsa

TarikGul:
This ensures that we only hash the accountId or key given for a ecdsa scheme type.


marshacb (approved):
looks good!

src/services/accounts/AccountsConvertService.spec.ts
```diff
@@ -158,7 +179,7 @@ describe('Convert accounts', () => {
 		const expectedResponse = {
```
Imod7, line 179: **Old Code**
By reading the comments above this test, it was giving as output an invalid address on purpose. 
The comments say that the user : 
- **gives** to the endpoint a **non** public key but 
- **says/states** falsely to the endpoint that it **is** a public key (by setting query param `publicKey=true`) 
and then the output is/should be an _invalid address_. So that was with the old code.

**New Code**
With the new code, the output is correct even if the user states that the input is a public key when its not. So in the case of `scheme=ecdsa`, if the user sets query param `publicKey` either `true` or `false`
it will go to this new line of code : 
`ecdsa: (p: string) => (hexToU8a(p).length > 32 ? blake2AsHex(p) : p),`
and it will encode correctly despite the value of `publicKey` query param.

**Thoughts**
So, some thoughts based on the above : 
- In retrospective, since I was the one who added the original test, I have to admit that it is a little bit confusing and not sure how useful it is anymore. I think my goal was to "catch" all possible combinations of inputs.
- However, if we keep this test with the new code, I think we should change the comments so that they reflect what the test is doing right now.
- Also, by checking the new code, I am not sure if the query param `publicKey` is needed anymore. Wdyt @TarikGul ? Since it will convert it correctly by checking the length (in the `ecdsa` case)

cc @IkerAlus Thank you for your time on trying to check this. It really helped on clearing my head and checking efficiently after our talk.

Imod7 (approved):

PR #1305, opened 2023-07-11
fix: query info feature detection

xlc:
The whole thing needs a refactoring so that the usage of API (i.e. fetchQueryInfo) is not detached to the API detection code but I don't all the time to do it so will just do the quick fix for now.

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -248,7 +248,10 @@ export class BlocksService extends AbstractService {
 				continue;
 			}
 
-			if (!api.rpc.payment || !api.rpc.payment.queryInfo) {
+			if (
+				!api.rpc.payment?.queryInfo &&
+				!api.call.transactionPaymentApi?.queryInfo
+			) {
 				extrinsics[idx].info = {
 					error: 'Rpc method payment::queryInfo is not available',
 				};


src/services/blocks/BlocksService.ts
```diff
@@ -248,7 +248,10 @@ export class BlocksService extends AbstractService {
 				continue;
 			}
 
-			if (!api.rpc.payment || !api.rpc.payment.queryInfo) {
+			if (
+				!api.rpc.payment?.queryInfo &&
+				!api.call.transactionPaymentApi?.queryInfo
+			) {
 				extrinsics[idx].info = {
 					error: 'Rpc method payment::queryInfo is not available',
```
Imod7, line 256: Thank you for the PR @xlc ! It looks good to me! I will just add here some of my thoughts : 

1. With this change, we do a complete check if `queryInfo` in general is available (either from runtime call or rpc) not just checking if `rpc` is available.
1. If we want this check, then we should change the error to something more generic like "The fee information is not available" ([src](https://polkadot.js.org/docs/kusama/rpc#queryinfoextrinsic-bytes-at-blockhash-runtimedispatchinfov1))
1. Also, if none of these methods is available then we should move this part of the [code](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/services/blocks/BlocksService.ts#L263-L267) to an `else` statement after this check. Since it makes sense to call the `fetchQueryInfo` function only if one of the two calls are available.
1. Additional secondary note : maybe we could also add the question mark wherever we call `apiAt.call.transactionPaymentApi.queryInfo`, e.g. this [line](https://github.com/paritytech/substrate-api-sidecar/blob/cafea8c8909fc77d4d636eaf1c5bac7eb4c78783/src/services/blocks/BlocksService.ts#L401C52-L401C52) or this [line](https://github.com/paritytech/substrate-api-sidecar/blob/cafea8c8909fc77d4d636eaf1c5bac7eb4c78783/src/services/transaction/TransactionFeeEstimateService.ts#L42), etc. ? So it would be `apiAt.call.transactionPaymentApi?.queryInfo`) as shown here in your PR.

TarikGul:
> The whole thing needs a refactoring so that the usage of API (i.e. fetchQueryInfo) is not detached to the API detection code but I don't all the time to do it so will just do the quick fix for now.

I see what you are saying, and I agree it has merit. Happy to put it on the backlog, and hopefully we can get to it as well when things start to lighten up.

TarikGul (approved):

src/services/blocks/BlocksService.ts
TarikGul: > Additional secondary note : maybe we could also add the question mark wherever we call apiAt.call.transactionPaymentApi.queryInfo, e.g. this [line](https://github.com/paritytech/substrate-api-sidecar/blob/cafea8c8909fc77d4d636eaf1c5bac7eb4c78783/src/services/blocks/BlocksService.ts#L401C52-L401C52) or this [line](https://github.com/paritytech/substrate-api-sidecar/blob/cafea8c8909fc77d4d636eaf1c5bac7eb4c78783/src/services/transaction/TransactionFeeEstimateService.ts#L42), etc. ? So it would be apiAt.call.transactionPaymentApi?.queryInfo) as shown here in your PR.

This wont be necessary (in the context of this PR and not a refactor) since the check is already done before any of the calls are made.

Issue #1287, opened 2023-05-30
Support for Foreign Assets on Assets Hub chains (Statemine / Statemint)


IkerAlus:
**Proposed Feature**

Statemine runtime v9420 and (most likely) Statemint runtime v9430 will have support for foreign fungible assets. This will allow other parachains to represent their assets, as well as bridges to represent assets from other networks. Hence external teams  with an integration with Assets Hub chains, particularly exchanges, will be interested in querying the foreign asset information from the chain. That is why API Sidecar should allow to query such information. 

This feature is realized by the addition of a new pallet, `foreignAssets`, to these runtimes. This new pallet has the same interface than the well-known `assets` pallet, thus a good approach may be to add similar endpoints to API Sidecar to query the storage of this new pallet. Following the current endpoints for `assets` pallet this could be:

- `GET /accounts/{accountID}/asset-balances`
- `GET /accounts/{accountID}/asset-approvals`
- `GET /pallets/assets/{assetID}/asset-info`

**Additional Information**

Check the respective PRs for [Statemine](https://github.com/paritytech/cumulus/pull/2133) and [Statemint](https://github.com/paritytech/cumulus/pull/2540) to learn the details of the feature.

IkerAlus:
partially done in https://github.com/paritytech/substrate-api-sidecar/pull/1314 

Closing this for now and reconsidering the remaining functionality in the future.

PR #1314, opened 2023-08-15
feat: add pallets/foreign-assets endpoint

Imod7:
Relates to: #1287 

## Description
Adds the endpoint `/pallets/foreign-assets/` which returns an array with the available foreign assets and their corresponding `AssetDetails` and `AssetMetadata`.

## Query Params

## Sample Response
While connected to `Kusama Asset Hub` and requesting `/pallets/foreign-assets/`, the return result is/should be the following : 

```
{
  "at": {
    "hash": "0xd890d7d27083b67d447406b55fb6bb07b382900d1643ce4343520b94e92055a0",
    "height": "5311742"
  },
  "items": [
    {
      "foreignAssetInfo": {
        "owner": "FxqimVubBRPqJ8kTwb3wL7G4q645hEkBEnXPyttLsTrFc5Q",
        "issuer": "FxqimVubBRPqJ8kTwb3wL7G4q645hEkBEnXPyttLsTrFc5Q",
        "admin": "FxqimVubBRPqJ8kTwb3wL7G4q645hEkBEnXPyttLsTrFc5Q",
        "freezer": "FxqimVubBRPqJ8kTwb3wL7G4q645hEkBEnXPyttLsTrFc5Q",
        "supply": "0",
        "deposit": "0",
        "minBalance": "100000000",
        "isSufficient": true,
        "accounts": "0",
        "sufficients": "0",
        "approvals": "0",
        "status": "Live"
      },
      "foreignAssetMetadata": {
        "deposit": "0",
        "name": "0x",
        "symbol": "0x",
        "decimals": "0",
        "isFrozen": false
      }
    },
    {
      "foreignAssetInfo": {
        "owner": "FBeL7DiQ6JkoypYATheXhH3GQr5de2L3hL444TP6qQr3yA9",
        "issuer": "FBeL7DiQ6JkoypYATheXhH3GQr5de2L3hL444TP6qQr3yA9",
        "admin": "FBeL7DiQ6JkoypYATheXhH3GQr5de2L3hL444TP6qQr3yA9",
        "freezer": "FBeL7DiQ6JkoypYATheXhH3GQr5de2L3hL444TP6qQr3yA9",
        "supply": "0",
        "deposit": "100000000000",
        "minBalance": "1000000000",
        "isSufficient": false,
        "accounts": "0",
        "sufficients": "0",
        "approvals": "0",
        "status": "Live"
      },
      "foreignAssetMetadata": {
        "deposit": "6693666633",
        "name": "0x54696e6b65726e6574",
        "symbol": "0x544e4b52",
        "decimals": "12",
        "isFrozen": false
      }
    }
  ]
}
```

## Todos
- [x] Add tests


src/services/test-helpers/mock/data/foreignAssets.ts
```diff
@@ -0,0 +1,12 @@
+// '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}';
```
marshacb, line 1: Do we still need this comment? Also, Im thinking we should add the copyright to this file. I see a few other mock data files that we missed the copyright for as well that we can update at some point.

marshacb (approved):
Awesome PR!! Just one small question/nit.

src/controllers/pallets/PalletsForeignAssetsController.ts
```diff
@@ -0,0 +1,58 @@
...
+ * - Foreign Assets Pallet Instance in Kusama Asset Hub: https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/asset-hub-kusama/src/lib.rs#L295
+ */
+export default class PalletsForeignAssetsController extends AbstractController<PalletsForeignAssetsService> {
+	constructor(api: ApiPromise) {
+		super(api, '/pallets/foreign-assets', new PalletsForeignAssetsService(api));
+		this.initRoutes();
+	}
+
+	protected initRoutes(): void {
+		this.safeMountAsyncGetHandlers([['', this.getForeignAssets]]);
```
TarikGul, line 45:
```suggestion
		this.safeMountAsyncGetHandlers([['/asset-info', this.getForeignAssets]]);
```

TarikGul: This needs to be added or else the only endpoint available is `/pallets/foreign-assets`

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ApiPromise } from '@polkadot/api';
+import { BlockHash } from '@polkadot/types/interfaces';
```
TarikGul, line 18:
```suggestion
import { BlockHash, AssetMetadata } from '@polkadot/types/interfaces';
```

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+				// remove any commas from multilocation key values e.g. Parachain: 2,125 -> Parachain: 2125
+				const foreignAssetMultiLocationStr = JSON.stringify(
+					foreignAssetData[0]
+				).replace(/(\d),/g, '$1');
+				const foreignAssetMultiLocation = api.registry.createType(
+					'MultiLocation',
+					JSON.parse(foreignAssetMultiLocationStr)
+				);
+
+				const assetMetadata = await api.query.foreignAssets.metadata(
```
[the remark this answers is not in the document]
TarikGul: So that the return type is correctly mapped instead of just a Codec.

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ApiPromise } from '@polkadot/api';
+import { BlockHash } from '@polkadot/types/interfaces';
+import { AnyJson } from 'src/types/polkadot-js';
```
TarikGul, line 19:
```suggestion
import { AnyJson } from '../../types/polkadot-js';
```

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ApiPromise } from '@polkadot/api';
```
TarikGul, line 17:
```suggestion
import type { ApiPromise } from '@polkadot/api';
```

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+import { BlockHash } from '@polkadot/types/interfaces';
+import { AnyJson } from 'src/types/polkadot-js';
+
+import { IForeignAssetInfo, IForeignAssets } from '../../types/responses';
```
TarikGul, line 21:
```suggestion
import type { IForeignAssetInfo, IForeignAssets } from '../../types/responses';
```

src/types/responses/ForeignAssets.ts
```diff
@@ -0,0 +1,31 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { AssetMetadata } from '@polkadot/types/interfaces';
+import { PalletAssetsAssetDetails } from '@polkadot/types/lookup';
+import { AnyJson } from 'src/types/polkadot-js';
+
+import { IAt } from '.';
+
+export interface IForeignAssetInfo {
+	foreignAssetInfo: PalletAssetsAssetDetails | AnyJson;
```
TarikGul, line 24: These types should have `AnyJson` removed from them

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+	 * Fetch all foreign asset's `AssetDetails` and `AssetMetadata`.
+	 *
+	 * @param hash `BlockHash` to make call at
+	 */
+	async fetchForeignAssets(hash: BlockHash): Promise<IForeignAssets> {
+		const { api } = this;
+
+		const [{ number }, foreignAssetInfo] = await Promise.all([
+			api.rpc.chain.getHeader(hash),
+			api.query.foreignAssets.asset.entries(),
```
TarikGul, line 39: Similar to below we need to correctly type the entries method. This is because both `assetStorageKeyData`, and `assetInfo` will have there own specific types.

`assetStorageKeyData: StorageKey<AnyTuple>`
`assetInfo: Option<PalletAssetsAssetDetails>`

```typescript
// Incorrect:
api.query.foreignAssets.asset.entries();

// Correct:
api.query.foreignAssets.asset.entries<Option<PalletAssetsAssetDetails>[]>()

// This is a little tricky but the above will map out to Promise<[StorageKey<AnyTuple>, Option<PalletAssetsAssetDetails>[]][]> once we input the correct type.
```

This will all ensure that the above is typed correctly so we wont need to type cast anymore and we can clean up all the types within the service.

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,90 @@
...
+				const foreignAssetMultiLocation = api.registry.createType(
+					'MultiLocation',
+					JSON.parse(foreignAssetMultiLocationStr)
+				);
+
+				const assetMetadata = await api.query.foreignAssets.metadata(
+					foreignAssetMultiLocation
+				);
+
+				if (assetMetadata) {
```
TarikGul, line 69: I dont think it's assetMetadata we need to check here, but instead `if (assetInfo.isSome)`, and then we need to unwrap the value when we key it into `foreignAssetInfo: assetInfo.unwrap()`.

But then we need to handle the case where the info is not available in an else statement. Such that if it doesn't exist we return an empty object.

TarikGul (changes requested):
Left some notes on how we can update the type accuracy so we dont need to type cast or have Union types with `AnyJson`. I think it should make our testing and endpoint a bit more accurate.

src/controllers/pallets/PalletsForeignAssetsController.ts
Imod7: Thank you so much @TarikGul for pointing this out. I had it before with `asset-info` but IIRC we said it is not needed (?) The main reasoning was that : 
- this endpoint shows just a list of foreign assets so `/pallets/foreign-assets` is shorter and enough to reflect the response that it gives
- also we do not have any path params like in the `/pallets/assets/{assetId}/asset-info` where we have the path param `{assetId}`

But it is no problem at all to add it again if you think it is better!

src/types/responses/ForeignAssets.ts
Imod7: I replaced with `T`/`Codec` like I had it in the beginning. If I remove completely and only leave types `PalletAssetsAssetDetails` and `AssetMetadata` it does not work.

src/services/pallets/PalletsForeignAssetsService.ts
Imod7: That is amazing!!! Thank you so much @TarikGul for spending the time to explain and showcase this. I was _fighting_ with the types quite a bit and now I get the logic! 🙏 🙏 🙏

src/services/test-helpers/mock/data/foreignAssets.ts
TarikGul: > in some other files that the copyright was with hear 2017-2022 I updated it to 2017-2023 but not all.

Yea lets try to save the rest for a seperate PR, it just adds a lot of bloat to the review.

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,94 @@
...
+				// remove any commas from multilocation key values e.g. Parachain: 2,125 -> Parachain: 2125
+				const foreignAssetMultiLocationStr = JSON.stringify(
+					foreignAssetData[0]
+				).replace(/(\d),/g, '$1');
+				const foreignAssetMultiLocation = api.registry.createType(
+					'MultiLocation',
+					JSON.parse(foreignAssetMultiLocationStr)
+				);
+
+				const assetMetadata = await api.query.foreignAssets.metadata<
```
TarikGul, line 69: `assetMetadata` is not an `Option` it's just `AssetMetadata`, `assetInfo` in the only type that is going to be wrapped in an option.

src/types/responses/ForeignAssets.ts
```diff
@@ -0,0 +1,33 @@
...
+import {
+	PalletAssetsAssetDetails,
+	PalletAssetsAssetMetadata,
+} from '@polkadot/types/lookup';
+
+import { IAt } from '.';
+
+export interface IForeignAssetInfo {
+	foreignAssetInfo: Option<PalletAssetsAssetDetails>;
+	foreignAssetMetadata: Option<PalletAssetsAssetMetadata>;
```
TarikGul, line 27: This needs to be changed to just `AssetMetadata`, I would prefer if we also just use `AssetMetadata` here because its the generic type for an Asset where as `PalletAssetsAssetMetadata` is referring to the Asset Pallet.

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -0,0 +1,96 @@
...
+		/**
+		 * This will iterate through all the foreign asset entries and for each entry it will create
+		 * the `foreignAssetMultiLocation` variable based on the MultiLocation of the foreign asset.
+		 * This variable will then be used as the key to get the corresponding metadata of the foreign asset.
+		 *
+		 * This is based on the logic implemented by marshacb in asset-transfer-api-registry
+		 * https://github.com/paritytech/asset-transfer-api-registry/blob/main/src/createRegistry.ts#L193-L238
+		 */
+		for (const [assetStorageKeyData, assetInfo] of foreignAssetInfo) {
+			console.log(assetInfo.toRawType());
```
TarikGul, line 54: Ahh this was my bad, but lets remove the console.log here

TarikGul (approved):
Just one tiny little nit with removing the console.log I pushed up by accident. Other than that, really great job!! LGTM

PR #1317, opened 2023-08-18
test(calc): add test for calc_payout

bee344:
* Added tests for calc_payout.rs .
* Added docs for calc_payouts and calc_fee.

TODO:
- [x] Gather tests in a single file.


calc/README.md
```diff
@@ -27,8 +27,84 @@ standalone through the npm registry.
...
+It uses the following formula:
+
+```
+partial_fee = base_fee + len_fee + ((adjusted_weight_fee/estimated_weight)*actual_weight)
+```
+
+Where:
+- `base_fee` is a fixed base fee to include some transaction in a block. It accounts
+  for the work needed to verify the signature and the computing work common to any tx.
+  It iss constant for any tx.
```
TarikGul, line 45:
```suggestion
  It is constant for any tx.
```

TarikGul (commented):
Looks really amazing so far 👍

calc/src/calc_payout.rs
```diff
@@ -102,3 +122,130 @@ impl CalcPayout {
...
     }
 }
+
+#[cfg(test)]
+mod test_payout {
+    use super::*;
+    /*
+    These tests are designed to test the accuracy of the preceding functions.
+    All the data can be retrieved and checked with polkadot-js and subscan.io.
+    NOTE: As these tests are setup, the eras should be regularly updated.
```
TarikGul, line 132: Im a bit confused by the `NOTE`. Does this mean these tests should be updated in the future?

calc/README.md
```diff
@@ -27,8 +27,84 @@ standalone through the npm registry.
...
+to swap out the weight used to calculate that `adjusted_weight_fee` with the 
+actual weight that was used from the `ExtrinsicSuccess` event. In the end, the 
+calculation itself is simple, but gathering the details needed is the main difficulty.
+We do this all in Rust simply to limit any precision loss.
+
+### calc_payout
+
+This is a tool to calculate the payout of a staking era, either for a validator 
+or a nominator. This is not a predictive estimation, instead it intakes data 
+from a concluded [era](https://wiki.polkadot.network/docs/kusama-parameters#periods-of-common-actions-and-attributes)
```
marshacb, line 91: do we mean something like `to arrive to the final amount.`?

marshacb (commented):
LGTM! just 1 nit/question

calc/src/calc_payout.rs
bee344: Noting this was a mistake on my part, only added confusion. It originated due to polkadot-js retrieveng upt to a certain amount of past eras, so I added the block number to the data so that it's easier to check the values and removed this `NOTE`.

TarikGul: Ahh makes sense!

Thanks for the explanation

TarikGul (approved):
LGTM, great job 👍

marshacb (approved):

PR #1335, opened 2023-09-27
docs: fix regex under the pattern keyword

Imod7:
### Issue

In 6 endpoints there is the following regex `'a-km-zA-HJ-NP-Z1-9{8,64}'` under the `pattern` keyword. The regex intends to validate the user input for:  
- the `blockId` parameter of the `block` related endpoints and
- the `address` parameter of the `contracts` endpoint

However, it is not working as expected hence is not allowing the user to input valid block heights or hashes. 

### Fix
In the `block` endpoints where the `blockId` is either a `block height` or a `block hash`, the regex was replaced by the following pattern : 

```
^0[xX][0-9a-fA-F]{1,64}$|[0-9]{1,12}
```

This pattern should match either :
- a hexadecimal string of total length of 66 digits. The pattern allows a string that starts with `0x` or `0X` and then its followed by 64 hexadecimal digits (0-9a-fA-F). This could work for the `block hash`.
- an integer number of length 1 to 12 digits. This could work for the `block height`.

In the `contracts` endpoint, the pattern was removed.

### Note
The replaced pattern in the `blockId` captures a lot of wrong inputs. However, it does not work 100% as expected (for example it ignores completely the length limit). I think this is a swagger issue since there are some open issues related to regex patterns not working correctly :
- https://github.com/swagger-api/swagger-ui/issues/8003
- https://github.com/swagger-api/swagger-ui/issues/8033


TarikGul (approved):
Great Job. LGTM.

bee344 (approved):
Nicely done

PR #1336, opened 2023-10-02
fix(deps): update pjs util-crypto

marshacb:


TarikGul:
We need to be careful with this since 10.9.1 of `@polkadot/api` uses: 12.3.1 ie: https://github.com/polkadot-js/api/blob/ac07123e54946dd6ba3d80ea8c1ab90ab86aaaa2/packages/api/package.json#L37

There could be some dependency misalignment if we have 2 seperate versions in our dep tree.

Imod7 (approved):

marshacb:
> We need to be careful with this since 10.9.1 of `@polkadot/api` uses: 12.3.1 ie: https://github.com/polkadot-js/api/blob/ac07123e54946dd6ba3d80ea8c1ab90ab86aaaa2/packages/api/package.json#L37
> 
> There could be some dependency misalignment if we have 2 seperate versions in our dep tree.

Hey Tarik, Great catch! So being that `pjs api` doesn't support this latest version of util-crypto yet we can postpone this dep update.

PR #1346, opened 2023-10-27
ci: Update runner tag to `linux-docker`

oleg-plakida:


TarikGul (approved):

TarikGul:
The only thing would be when this is emerged is to ensure that merge commit is the same as the title ie. conventional commit.

Imod7 (approved):

alvicsam (approved):

oleg-plakida:
That was premature. I might revert it later because of changed paradigm.

oleg-plakida:
No worries. It won't affect ci. But i need to optimize it. Mb will make another PR.

TarikGul:
> No worries. It won't affect ci. But i need to optimize it. Mb will make another PR.

That was my b as well (I should ask if its good to merge just in case). That being said thank so much for everything!

PR #1348, opened 2023-11-01
ci: adding gitspiegel-trigger workflow

mutantcornholio:
Using a workflow to trigger mirroring instead of a webhook allows us to reuse "Approving workflow runs from public forks" GitHub feature to somewhat protect us from malicious PRs


TarikGul (approved):
lgtm thanks so much for the PR. One thing: when this is merged lets make sure the merge commit is in the format of `conventional commits`, so mirroring the title.

alvicsam (approved):

Issue #1259, opened 2023-04-06
Add docker support for linux/arm64/v8


kitti-katy:
I have Mac M1 and when trying to run the docker image from `docker.io/parity/substrate-api-sidecar:latest` I get: 
```
➜  substrate-api docker run --rm -it --read-only -p 8080:8080 parity/substrate-api-sidecar
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

#
# Fatal process OOM in Failed to reserve virtual memory for CodeRange
#

qemu: uncaught target signal 5 (Trace/breakpoint trap) - core dumped
```

TarikGul:
Hmm, this is interesting. I am running it on Macbook Pro M2 and have no issues. I'll give this a look.

TarikGul:
Okay so I was able to reproduce the warning: That being said the image still runs which is also weird.

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

PR #1360, opened 2023-11-22
fix(calc): remove wasm-pack from node, and add it to calc library

TarikGul:
closes: https://github.com/paritytech/substrate-api-sidecar/issues/1259

This PR removes the wasm-pack package from node, and adds it into the script for building calc. It will ask if the user wants to download it globally in order to build. It wont add anything to your system without your explicit permission. 

This had to be done because wasm-pack was giving issues for building docker images using: `docker build --platform linux/arm/v8 -t substrate-api-sidecar .`


bee344 (approved):

calc/build.sh
```diff
@@ -2,11 +2,44 @@
...
+    WASM_PACK_VERSION=0.12.1
+    WASM_PACK="wasm-pack"
+
+    uname_out="$(uname -s)"
+    case "${uname_out}" in
+        Linux*)     machine=Linux;;
+        Darwin*)    machine=Darwin;;
+    esac
+
+    if [ $(cargo install --list | grep -q "wasm-pack v${WASM_PACK_VERSION}") ]; then
```
Imod7, line 15:
```suggestion
    if [ $(cargo install --list | grep -q "wasm-pack v${WASM_PACK_VERSION}"; echo $?) != 0 ]; then
```
This line did not work for me (idk if it is related to the fact that I am running in MacOS). It will always tell me that `wasm-pack` is already installed even if it is not. I tested it by removing `wasm-pack` with `cargo uninstall wasm-pack` and run `sh build.sh` again.

By changing this line of code, it works as expected. 
I added the `echo $?` so it prints the return value of `grep` which is :
- either `0` if it finds the `wasm-pack` hence it is installed
- or non zero, usually `1` if it is not found/not installed

It is tricky. I would expect that it works also without this change but maybe it is related to the fact that `grep` does not return the result so we need the `echo` ?

TarikGul: Yea this is a totally sane addition. I wonder how I missed this when running it, but your addition I think is needed and a way of making this way more accurate.

Imod7 (approved):

Issue #1365, opened 2023-12-21
API Sidecar request is very slow


zhy827827:
Since this morning, API requests have been very slow, previously very fast

What causes it.
```
2023-12-21 05:40:08 http: GET /blocks/18689818 200 26229ms
2023-12-21 05:40:10 http: GET /blocks/18687006 200 37203ms
2023-12-21 05:40:11 http: GET /blocks/18687150 200 41863ms
2023-12-21 05:40:11 http: GET /blocks/18687153 200 42968ms
2023-12-21 05:40:15 http: GET /blocks/18687004 200 42359ms
2023-12-21 05:40:15 http: GET /blocks/18687003 200 42816ms
2023-12-21 05:40:17 http: GET /blocks/18687005 200 43683ms
2023-12-21 05:40:19 http: GET /blocks/18687007 200 45851ms
2023-12-21 05:40:19 http: GET /blocks/head/header 200 141ms
2023-12-21 05:40:23 http: GET /blocks/head/header 200 56ms
2023-12-21 05:40:23 http: GET /blocks/head/header 200 56ms
2023-12-21 05:40:29 http: GET /blocks/head/header 200 67ms
2023-12-21 05:40:36 http: GET /blocks/18687005 200 378ms
2023-12-21 05:40:37 http: GET /blocks/18687007 200 529ms
2023-12-21 05:40:38 http: GET /blocks/18687007 200 261ms
2023-12-21 05:40:38 http: GET /blocks/18687005 200 289ms
2023-12-21 05:40:38 http: GET /blocks/head 200 42737ms
2023-12-21 05:40:38 http: GET /blocks/head/header 200 87ms
2023-12-21 05:40:38 http: GET /blocks/head/header 200 84ms
2023-12-21 05:40:42 http: GET /blocks/18689819 200 31477ms
2023-12-21 05:40:42 http: GET /blocks/18687154 200 42514ms
2023-12-21 05:40:42 http: GET /blocks/head/header 200 116ms
2023-12-21 05:40:42 http: GET /blocks/head/header 200 117ms
2023-12-21 05:40:44 http: GET /blocks/head 200 43090ms
2023-12-21 05:40:45 http: GET /blocks/18687152 200 45707ms
2023-12-21 05:40:46 http: GET /blocks/18687151 200 45929ms
2023-12-21 05:40:48 http: GET /blocks/18687006 200 42921ms
2023-12-21 05:40:50 http: GET /blocks/head 200 43634ms
[...]

PR #1366, opened 2023-12-21
fix: added query to calc fees

bee344:
Partially #1365


src/services/blocks/BlocksService.ts
```diff
@@ -168,6 +169,11 @@ export class BlocksService extends AbstractService {
 		const previousBlockHash = await this.fetchPreviousBlockHash(number);
 
 		for (let idx = 0; idx < block.extrinsics.length; ++idx) {
+			if (noFees === true) {
```
TarikGul, line 172:
```suggestion
			if (noFees) {
```

marshacb (approved):
LGTM

GCrispino:
Sorry to comment in a closed pull request, but if this was done to reduce performance issues, any specific reason for not making the loop that processes extrinsics concurrent?

TarikGul:
@GCrispino There is a few things to tackle here:

1. This PR is aimed at providing an option to opt out of the fees. It's one of many ways we are going to increase the performance here.

2. https://github.com/paritytech/substrate-api-sidecar/pull/1367 We have a follow up PR here to increase the performance as well that does not require using a query parameter.

3. You are correct as we should find fees for each extrinsic concurrently moving forward, and we are working on getting a good solution going. But we also need to optimize the way we structure the calls to get the information necessary for fee calculation. We already have a PromiseQueue implemented for this: https://github.com/paritytech/substrate-api-sidecar/blob/master/src/util/PromiseQueue.ts#L5

Overall, we are taking steps to do this correctly. It's a balance between getting some fixes out now, as well as sorting out the larger issue correctly.

GCrispino:
@TarikGul 

Got it! This makes sense, thanks for taking the time to explain 👍🏼

PR #1367, opened 2023-12-21
perf: transactionPaidFee event optimization

TarikGul:
This ensures no Promises are made when they are not needed.


marshacb (approved):
looks good!

bee344 (approved):
lgtm

PR #1368, opened 2023-12-21
perf: add concurrency to fee calls

TarikGul:
This adds concurrency to the extrinsic calls, with a task queue of 4 calls running concurrently.

Given the previous regression this was able to bring the request time from 70s to 26s.

Along with https://github.com/paritytech/substrate-api-sidecar/pull/1367, requests are now down to 3s or less.


marshacb (approved):
LGTM!

bee344 (approved):

PR #1369, opened 2023-12-21
chore(release): 17.3.3

bee344:


marshacb (approved):
lgtm!

CHANGELOG.md
```diff
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [17.3.3](https://github.com/paritytech/substrate-api-sidecar/compare/v17.3.2..v17.3.3) (2023-12-21)
+
+**NOTE** This release focuses on improving the performance of the tool resolving a regression where `blocks` where overwhelmed with transactions. `noFees` focuses on removing fee info for the blocks in order to improve performance and for the other cases, there is the `transactionPaidFee` PR, that covers a more wide approach to performance improvement.
```
TarikGul:
```suggestion
**NOTE** This release focuses on improving the performance of the server by resolving a regression where `blocks` were overwhelmed with transactions. The `noFees` query parameter focuses on removing fee info for the blocks if the user does not intend on needing fees. For more general cases where fees are necessary we have increased the performance of querying `/blocks` while also calculating fees. This was done with 2 cases: ensuring `transactionPaidFee`, and `ExtrinsicSuccess` or `ExtrinsicFailure` info is used to its fullest so we can avoid making any additional rpc calls, as well as ensuring the extrinsic's are called concurrently.
```

TarikGul (approved):

exp0nge:
Is this version safe to use against older node versions? We have 0.9.43

TarikGul:
> Is this version safe to use against older node versions? We have 0.9.43

We can't give any guarantees with older node versions, but I don't see why it wouldn't work. It's always recommended to be up to date if possible, but that being said sidecar always does its best to maintain historical integrity as much as possible so things that are runtime related are usually sorted out historically.

Issue #1033, opened 2022-08-30
feat: Increase the `history_depth` beyond the 84 eras in `staking-payouts` endpoint


Imod7:
### Issue
The endpoint `/accounts/{accountId}/staking-payouts` (see [docs](https://paritytech.github.io/substrate-api-sidecar/dist/)) will not show you staking info beyond 84 eras ago (there is a limitation on the depth). If we use the sideCar API to track the historical earning and total stake balance data, with this limitation it is hard to fill the historical data and use these data to monitor a validators performance changes.

### Proposed Change or Idea
Enable Sidecar's `GET /accounts/accountID/staking-payouts` [endpoint](https://paritytech.github.io/substrate-api-sidecar/dist/) to return the payout data for any past era, similar to what can be done for most of the other endpoints (`GET /accounts/accountID/staking-info` for example). We would still need to specify the `depth` and the `era` as input parameters. This would work of course only if we query an [archive node](https://wiki.polkadot.network/docs/maintain-sync#types-of-nodes) to have access to all the past blocks.

PR #1388, opened 2024-02-01
feat: add `at` query param for staking-payouts

TarikGul:
## Summary

This PR focuses on expanding the query params for `/accounts/{accountId}/staking-payouts` to include the `at` query param. This required me to remove the old `api.derive.staking.eraExposure(eraIndex)` that is exposed from polkadot-js and write a modified version that was abstracted from the polkadot-js source code to allow us to add a historic api to it.

### Query param

`at` - Block height (as a non-negative integer) or hash (as a hex string).

## Extra

I rewrote all the tests for the endpoint to be cleaner and include specific information on the block and the era so if anything needs to be added in the future it will be simple to extract the necessary data from polkadot.

closes: https://github.com/paritytech/substrate-api-sidecar/issues/1033


Imod7 (approved):
Great work! 💯 

It might be worth noting that Subscan and Sidecar display rewards differently (at least in the views I checked). I thought of adding the difference here in case it is helpful for others who might also check Subscan: 
- For example, while connected in Kusama and checking some rewards from this [list](https://kusama.subscan.io/reward?address=J47U9wGuwzccFPoz8bnMTKJt7WGpPp8ZNgvtXFDL9PHwpCt&category=Reward&role=validator&page=5) from Subscan I can see different rewards in different eras and blocks.
- For example, for block `20193750` I can see the reward `0.152 KSM`for era `5787`
- If I query the Sidecar endpoint for this block `accounts/J47U9wGuwzccFPoz8bnMTKJt7WGpPp8ZNgvtXFDL9PHwpCt/staking-payouts?unclaimedOnly=false&at=20193750` it returns the reward for era `5789`
- The difference in the blocks / eras is due to the fact that:
   - Sidecar gives back the rewards that correspond to each era (claimed or unclaimed) so it gives the `state` information.
   - Subscan gives back only the rewards that were claimed hence when and if someone made that call to get the corresponding reward. If no one makes that call and if 84 eras pass, this reward will be lost.

**Important Note**
- The reward amounts per era is consistent across both platforms.

Credits to @IkerAlus for the explanation and clarifications.

marshacb (approved):

PR #1391, opened 2024-02-05
ci: Replace helm deploy with argo cd

BulatSaif:
Issue: https://github.com/paritytech/devops/issues/2977

## changes
`helm deploy` command was replaced by `argocd set/sync` command

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,12 +9,6 @@
 
 variables:
   CONTAINER_REPO:                  "docker.io/parity/substrate-api-sidecar"
-  HELM_NAMESPACE:                  "substrate-api-sidecar"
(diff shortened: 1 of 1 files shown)


BulatSaif:
> Thanks for the PR! Does this make the current `/helm` directory obsolete?

ArgoCD uses the same ./helm directory. If the directory is not needed here, I can move it to https://github.com/paritytech/helm-charts as most of our applications.

TarikGul:
> > Thanks for the PR! Does this make the current `/helm` directory obsolete?
> 
> ArgoCD uses the same ./helm directory. If the directory is not needed here, I can move it to https://github.com/paritytech/helm-charts as most of our applications.

Totally fine for the helm charts to exist in sidecar. It was just a base question to get a better understanding. Thanks again

TarikGul (approved):

PR #1408, opened 2024-03-06
fix: multilocation type in pallets/foreign-assets

Imod7:
### Description of the issue
The endpoint `pallets/foreign-assets` throws the following error:
```
{
  "code": 500,
  "message": "createType(XcmV3MultiLocation):: DoNotConstruct: Cannot construct unknown type XcmV3MultiLocation",
  "stack": "Error: createType(XcmV3MultiLocation):: DoNotConstruct: Cannot construct unknown type XcmV3MultiLocation\n    at createTypeUnsafe ...",
  "level": "error"
}
```

### Root Cause
The error is thrown when it tries to create the `XcmV3MultiLocation` type in this [line](https://github.com/paritytech/substrate-api-sidecar/blob/96c9f11d7c4142c6f8a145eb932fb650941ade50/src/services/pallets/PalletsForeignAssetsService.ts#L60).

### Suggested Solution
First try to create the `StagingXcmV3MultiLocation` type and if it is not successful try the `XcmV3MultiLocation` type.

### Testing
Tested the endpoint by connecting to Polkadot/Kusama/Rococo Asset Hub

### DIFF
--- a/src/services/pallets/PalletsForeignAssetsService.ts
+++ b/src/services/pallets/PalletsForeignAssetsService.ts
@@ -18,7 +18,6 @@ import { ApiPromise } from '@polkadot/api';
 import { Option } from '@polkadot/types';
 import { AssetMetadata, BlockHash } from '@polkadot/types/interfaces';
 import { PalletAssetsAssetDetails } from '@polkadot/types/lookup';
-import { InternalServerError } from 'http-errors';
 
 import { IForeignAssetInfo, IForeignAssets } from '../../types/responses';
 import { AbstractService } from '../AbstractService';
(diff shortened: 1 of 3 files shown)


src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -56,10 +57,23 @@ export class PalletsForeignAssetsService extends AbstractService {
 			if (foreignAssetData) {
 				// remove any commas from multilocation key values e.g. Parachain: 2,125 -> Parachain: 2125
 				const foreignAssetMultiLocationStr = JSON.stringify(foreignAssetData[0]).replace(/(\d),/g, '$1');
-				const foreignAssetMultiLocation = api.registry.createType(
-					'XcmV3MultiLocation',
-					JSON.parse(foreignAssetMultiLocationStr),
-				);
+				const XcmV3MultiLocation = ['StagingXcmV3MultiLocation', 'XcmV3MultiLocation'];
```
IkerAlus, line 60: do we need to create the type from hardcoded options? Bear in mind this will be an issue soon when XCM v4 is life and XCM types are renamed.

IkerAlus: actually, this should be already the case if foreign assets in Rococo Asset Hub are queried.

Imod7: - You mean retrieve dynamically the `XCM Multilocation` related types, check what is available and try the ones that should work right ? I haven't tried to create types this way but one thing that comes into my mind is the potential impact in the `response time`. 🤔 If there is such an implementation, I am guessing it will take longer to process the request (go through these steps) in order to return the expected result.
- I just tested against Rococo Asset Hub and it returns the foreign assets correctly.

IkerAlus: - You are right, that solution would be an overkill, but we just need keep an eye for this changes.
- And this makes sense for now, because we are talking about the storage types not the XCM types from the messages. But if Asset hub follows XCM v4 typing, then this will break again, just have it in mind.

IkerAlus (approved):
LGTM.

src/services/pallets/PalletsForeignAssetsService.ts
```diff
@@ -56,10 +57,23 @@ export class PalletsForeignAssetsService extends AbstractService {
...
+						foreignAssetMultiLocation = api.registry.createType(
+							XcmV3MultiLocation[1],
+							JSON.parse(foreignAssetMultiLocationStr),
+						);
+					} catch (e) {
+						throw new InternalServerError(`Failed to create MultiLocation type for Foreign Asset`);
+					}
+				}
 
 				const assetMetadata = await api.query.foreignAssets.metadata<AssetMetadata>(foreignAssetMultiLocation);
```
TarikGul, line 78: Is it possible to just put the multilocation in the query as an Object or string and let the API take care of the type creation.

TarikGul: Just confirmed it can.

src/services/pallets/PalletsForeignAssetsService.ts
TarikGul, line 60: We can get rid of all of this, the try catches, and the xcm multilocation array, and just pass in an object of the multilocation. Given this [comment](https://github.com/polkadot-js/api/issues/5741#issuecomment-1772151935) we dont want to be using createType directly anymore unless its testing, and we want to avoid using a nested try catch at all costs.

TarikGul (changes requested):
Check comment above.

TarikGul (approved):
LGTM

PR #1415, opened 2024-03-20
fix: add Kusama chain check in historic blocks

Imod7:
### Description
Added a check to determine if we are connected in the Kusama chain, as the calculations are `Kusama` specific.

### DIFF
--- a/src/controllers/accounts/AccountsStakingPayoutsController.ts
+++ b/src/controllers/accounts/AccountsStakingPayoutsController.ts
@@ -105,17 +105,18 @@ export default class AccountsStakingPayoutsController extends AbstractController
 		let hash = await this.getHashFromAt(at);
 		let apiAt = await this.api.at(hash);
 		const runtimeInfo = await this.api.rpc.state.getRuntimeVersion(hash);
+		const isKusama = runtimeInfo.specName.toString().toLowerCase() === 'kusama';
 		const { eraArg, currentEra } = await this.getEraAndHash(apiAt, this.verifyAndCastOr('era', era, undefined));
-		if (currentEra <= 519 && depth !== undefined && runtimeInfo.specName.toString() === 'kusama') {
+		if (currentEra <= 519 && depth !== undefined && isKusama) {
 			throw new InternalServerError('The `depth` query parameter is disabled for eras less than 518.');
-		} else if (currentEra <= 519 && era !== undefined && runtimeInfo.specName.toString() === 'kusama') {
+		} else if (currentEra <= 519 && era !== undefined && isKusama) {
 			throw new InternalServerError('The `era` query parameter is disabled for eras less than 518.');
 		}
 		let sanitizedDepth: string | undefined;
 		if (depth) {
 			sanitizedDepth = Math.min(Number(depth), currentEra - 518).toString();
 		}
-		if (currentEra < 518 && runtimeInfo.specName.toString() === 'kusama') {
+		if (currentEra < 518 && isKusama) {
 			const eraStartBlock: number = earlyErasBlockInfo[currentEra].start;
 			hash = await this.getHashFromAt(eraStartBlock.toString());
 			apiAt = await this.api.at(hash);
--- a/src/controllers/accounts/AccountsStakingPayoutsController.ts
+++ b/src/controllers/accounts/AccountsStakingPayoutsController.ts
@@ -113,7 +113,7 @@ export default class AccountsStakingPayoutsController extends AbstractController
 			throw new InternalServerError('The `era` query parameter is disabled for eras less than 518.');
 		}
 		let sanitizedDepth: string | undefined;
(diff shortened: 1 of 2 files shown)


src/controllers/accounts/AccountsStakingPayoutsController.ts
```diff
@@ -104,17 +104,18 @@ export default class AccountsStakingPayoutsController extends AbstractController
 		const earlyErasBlockInfo: IEarlyErasBlockInfo = kusamaEarlyErasBlockInfo;
 		let hash = await this.getHashFromAt(at);
 		let apiAt = await this.api.at(hash);
+		const runtimeInfo = await this.api.rpc.state.getRuntimeVersion(hash);
 		const { eraArg, currentEra } = await this.getEraAndHash(apiAt, this.verifyAndCastOr('era', era, undefined));
-		if (currentEra <= 519 && depth !== undefined) {
+		if (currentEra <= 519 && depth !== undefined && runtimeInfo.specName.toString() === 'kusama') {
```
TarikGul, line 109: Why is the 519 and the one below but the other check is 518?

src/services/accounts/AccountsStakingPayoutsService.ts
```diff
@@ -474,18 +480,20 @@ export class AccountsStakingPayoutsService extends AbstractService {
 				commission = (prefs[0] as PalletStakingValidatorPrefs | ValidatorPrefsWithCommission).commission.unwrap();
 			}
 		} else {
-			commissionPromise = ancient
-				? historicApi.query.staking.validators(validatorId)
-				: historicApi.query.staking.erasValidatorPrefs(era, validatorId);
+			commissionPromise =
+				ancient && specName.toLowerCase() === 'kusama'
```
TarikGul, line 484: We can do the `isKusama` variable here as well if you'd like.

src/services/accounts/AccountsStakingPayoutsService.ts
```diff
@@ -136,6 +136,16 @@ export class AccountsStakingPayoutsService extends AbstractService {
...
 
+		const at: IBlockInfo = {
+			height: number.unwrap().toString(10),
+			hash,
+		};
+
+		// User friendly - we don't error if the user specified era & depth combo <= 0, instead just start at 0
+		const startEra = Math.max(0, sanitizedEra - (depth - 1));
+		const runtimeInfo = await this.api.rpc.state.getRuntimeVersion(at.hash);
+		const specName = runtimeInfo.specName.toString();
```
bee344, line 147: Couldn't we just call `toLowerCase()` here? Since it's repeating every time we use `specName`.

src/controllers/accounts/AccountsStakingPayoutsController.ts
```diff
@@ -104,17 +104,18 @@ export default class AccountsStakingPayoutsController extends AbstractController
 		const earlyErasBlockInfo: IEarlyErasBlockInfo = kusamaEarlyErasBlockInfo;
 		let hash = await this.getHashFromAt(at);
 		let apiAt = await this.api.at(hash);
+		const runtimeInfo = await this.api.rpc.state.getRuntimeVersion(hash);
```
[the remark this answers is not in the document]
bee344: Beat me to it

src/controllers/accounts/AccountsStakingPayoutsController.ts
bee344: This is an edge case where the old logic changes to the current one, it's on purpose, otherwise you completely skip 518

TarikGul: Ahh i see nice. Thanks for the explanation

marshacb (approved):

src/controllers/accounts/AccountsStakingPayoutsController.ts
```diff
@@ -104,17 +104,19 @@ export default class AccountsStakingPayoutsController extends AbstractController
...
 		const { eraArg, currentEra } = await this.getEraAndHash(apiAt, this.verifyAndCastOr('era', era, undefined));
-		if (currentEra <= 519 && depth !== undefined) {
+		if (currentEra <= 519 && depth !== undefined && isKusama) {
 			throw new InternalServerError('The `depth` query parameter is disabled for eras less than 518.');
-		} else if (currentEra <= 519 && era !== undefined) {
+		} else if (currentEra <= 519 && era !== undefined && isKusama) {
 			throw new InternalServerError('The `era` query parameter is disabled for eras less than 518.');
 		}
 		let sanitizedDepth: string | undefined;
 		if (depth) {
```
bee344, line 116:
```suggestion
		if (depth && isKusama) {
```
Because this check is only relevant for the old Kusama logic

bee344 (approved):
Looks mighty fine to me

PR #1444, opened 2024-05-28
chore(yarn): bump yarn to 4.2.2

TarikGul:
I was using sidecar to benchmark some stuff in pjs-api and saw that yarn could use an update to latest.


Imod7 (approved):
Thank you! 💯

marshacb (approved):

IkerAlus (approved):

PR #1465, opened 2024-07-30
feat: Add route based metrics across API

filvecchiato:
Enables the metrics for all routes available in the API. An extra arg "-pq" has been created to allow the user to select if queryparams should be taken into consideration when creating metrics. An extra command has been included to run all benchmarks in one command in /benchmarks/init.sh

the following histograms have been created: 
- route latency
- route response size
- route response size - latency ratio

example use:
yarn dev -p => metrics route based

yarn dev -p -pq => metrics route based with queryparams


benchmarks/init.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+for d in ./*/ ; do (cd "$d" && export WRK_TIME_LENGTH=30s; sh init.sh); done
```
TarikGul, line 3: New line. (Nice improvement here!)

src/main.ts
```diff
@@ -65,9 +65,24 @@ async function main() {
...
+		// Generate metrics middleware
+		middlewares.push(metricsApp.middleware());
+		// Start the Metrics server
+		metricsApp.listen();
+	}
+
 	// Create our App
 	const app = new App({
-		preMiddleware: [json(), middleware.httpLoggerCreate(logger)],
+		preMiddleware: middlewares,
```
TarikGul, line 85: Small nit -  but I think we can just call `middlewares` `preMiddleware`. That way its a bit more clear as to what value it holds.

src/util/metrics.ts
```diff
@@ -26,32 +78,179 @@ export default class Metrics_App {
 	constructor({ host }: IAppConfiguration) {
 		const args = parseArgs();
 
+		this.includeQueryParams = Boolean(args.prometheus_queryparams);
```
TarikGul, line 81: Generally speaking, I like to stay away from using `Boolean()`, `Number()` etc because its a bit underperforming. I think here you can just do `!!args.prometheus_queryparams`

src/util/metrics.ts
```diff
@@ -26,32 +78,180 @@ export default class Metrics_App {
...
+				throw new Error('Summary not implemented');
+			default:
+				throw new Error('Unknown metric type');
+		}
+	}
+
+	private getRoute(req: Request) {
+		let route = req.baseUrl;
+		if (req.route) {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
TarikGul, line 140: I think we can get the eslint disables removed by explicitly setting the type on `Request` since it takes in a set of Generics: ie `Request<P = core.ParamsDictionary, ResBody = any, ReqBody = any, ReqQuery = qs.ParsedQs, Locals extends Record<string, any> = Record<string, any>>`

TarikGul (approved):
Great Job! Lgtm 🚀

src/util/metrics.ts
```diff
@@ -26,32 +118,245 @@ export default class Metrics_App {
...
+			const end = request_duration_seconds.startTimer();
+
+			const oldJson = res.json;
+			res.json = (body: Body) => {
+				res.locals.body = body;
+				return oldJson.call(res, body);
+			};
+
+			res.once('finish', () => {
+				if (res.statusCode >= 400) {
```
Imod7, line 302:
```suggestion
				if (res.statusCode >= 400 && req.originalUrl != '/favicon.ico') {
```
I am suggesting this because with each request/endpoint we query, a second request is done for the favicon on the background and we get a `warn: GET /favicon.ico 404 3ms` in the console which comes by default from `pjs-api`. 

With this addition 
- we ignore the `favicon` request and do not count it as an error.
- the `sas_request_errors` counter will only increase if there is an actual error in the endpoint requested.
- it is less accurate because there is a GET request that is still being made but it depends on what you actually want to count with the error counter.

Without this addition
- with every successful request we will see both `sas_request_success` counter and `sas_request_errors` counter increase (due to the erroneous favicon request) which is a little bit confusing.
- even though this is more accurate as number of requests and errors, again it depends on what you want to count with these metrics.

Mentioning this just to highlight what is happening here but it is up to you on how you will move forward.

src/util/metrics.ts
```diff
@@ -26,32 +118,245 @@ export default class Metrics_App {
...
+					const request_errors = this.metrics['sas_request_errors'] as client.Counter;
+					request_errors.inc();
+				} else {
```
Imod7, line 305:
```suggestion
				} else if (res.statusCode < 400) {
```
This is needed also if the above is added because otherwise it will keep increasing the `sas_request_success` counter even for erroneous requests.

src/util/metrics.ts
```diff
@@ -26,32 +118,245 @@ export default class Metrics_App {
...
+				.observe(seconds / blocks);
+		}
+	}
+
+	preMiddleware() {
+		return (req: Query, res: Response, next: () => void) => {
+			const tot_requests = this.metrics['sas_total_requests'] as client.Counter;
+
+			// request count metrics
+			tot_requests.inc();
```
Imod7, line 291:
```suggestion
			if (req.originalUrl != '/favicon.ico') {
				tot_requests.inc();
			}
```
This is related to the comments below. Without this check and with every requested endpoint we will see the `sas_total_requests`  counter increase by 2 (it counts the request from the user plus the favicon request) instead of 1.

Imod7 (changes requested):
Nice work! Those metrics are great additions! 💯

README.md
```diff
@@ -256,12 +256,33 @@ You can also define a custom port by running :
...
+- `sas_request_errors_total`: type counter and tracks http errors occuring in sidecar
+- `sas_request_success_total`: type counter and tracks successfull http requests
+- `sas_requests_total`: type counter and tracks all http requests
+- `sas_request_duration_seconds`: type histogram and tracks the latency of the requests
+- `sas_response_size_bytes_seconds`: type histogram and tracks the response size of the requests
+- `sas_response_size_latency_ratio_seconds`: type histogram and tracks the response bytes per second of the requests
+
+The blocks controller also includes the following route-specific metrics:
 
+- `sas_extrinsics_in_request_count`: type histogram and tracks the number of extrinsics returned in the request
```
Imod7, line 282: All changes look great! 💯 Just a last tiny thing based on your feedback during our call. Since this metric `sas_extrinsics_in_request_count` is added more for the `../blocks?range...` endpoint, maybe we can add it as a comment here?

Imod7 (approved):
Awesome work! Just a last tiny comment. 
Very excited with these metrics and what insights they will give us in Sidecar's performance.

PR #1470, opened 2024-08-09
feat: add support for Enjin and Canary

Bradez:
This pull request adds support for both Enjin and Canary chains.

### DIFF
--- a/src/chains-config/canaryControllers.ts
+++ b/src/chains-config/canaryControllers.ts
@@ -1,0 +1,62 @@
+// Copyright 2017-2022 Parity Technologies (UK) Ltd.
+// This file is part of Substrate API Sidecar.
+//
+// Substrate API Sidecar is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { ControllerConfig } from '../types/chains-config';
+import { initLRUCache, QueryFeeDetailsCache } from './cache';
+
+/**
+ * Canary configuration for Sidecar.
+ */
+export const canaryControllers: ControllerConfig = {
+	controllers: [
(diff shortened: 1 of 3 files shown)


Bradez:
Thank you for your response, @Imod7.

There was a discussion that certain repositories that are currently under @paritytech serve the purpose of the wider Substrate ecosystem generally, and are not specific to Polkadot. The [ss58-registry](https://github.com/paritytech/ss58-registry) being a perfect example of this.

These changes have been observed with ecosystems outside of Polkadot and Kusama having being included in some of these repositories, despite your statement. Thereby, we see an exclusion being nothing but an attack against @enjin, rather than conforming with the wider beliefs of the public and members of your own team who have publicly said otherwise. I would propose a request for you to discuss this matter internally once more to ensure Parity is aligned with these Substrate-generic repositories and how you intend to handle them.

The alternative is a fork of these repositories which are managed by an independent party of all entities for the purpose of being truly generalised, non-biased and in the benefit of the wider ecosystem.

It is acknowledged the specific license for this particular repository.

Imod7:
Hello again @Bradez 

Just to expand on the license part, Sidecar operates under `GNU General Public License` (e.g. mentioned [here](https://github.com/paritytech/substrate-api-sidecar/blob/master/src/controllers/blocks/BlocksController.ts#L4L6)) which means if you wish to use this tool, you would also need to have your code open source which, as far as I know, is not the case for your team.
Quoting from the GNU website [link](https://www.gnu.org/licenses/gpl-faq.en.html#WhyUseGPL):
> _Using the GNU GPL will require that all the [released improved versions be free software](https://www.gnu.org/philosophy/pragmatic.html). This means you can avoid the risk of having to compete with a proprietary modified version of your own work._

Regarding my alignment with my team: 
- The links I provided in this [comment](https://github.com/paritytech/substrate-api-sidecar/pull/1470#issuecomment-2286102024) I think can serve as good evidence and
- Additionally, I had already discussed this internally before providing my initial response [here](https://github.com/paritytech/substrate-api-sidecar/pull/1470#issuecomment-2286102024).

Last but not least, I would like to highlight that there is no intention of targeting Enjin or any other team. Parity and the teams within the Polkadot ecosystem that I am aware of, are building great tools that serve the entire ecosystem and contributing a lot of high-quality open-source code.

TarikGul:
@Bradez 

> There was a discussion that certain repositories that are currently under https://github.com/paritytech serve the purpose of the wider Substrate ecosystem generally, and are not specific to Polkadot

Sidecar has always been Polkadot, and Kusama first - parachains second. The Default Chains config is there for everything else.

In addition, we will not be merging in any chains-config of a chain we can not properly access and debug when issues happen. I see no reason to add the following PR given that we would have the burden of maintaining it, if and when issues arise.

PR #1477, opened 2024-08-15
feat: Inject metrics registry in route controllers

filvecchiato:
This PR allows for route-specific metrics to be emitted from the controllers, by injecting in the Response object (in locals) the metrics registry, making it available for each request received in case the prometheus flag is true

### DIFF
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -18,6 +18,7 @@ import type { ApiPromise } from '@polkadot/api';
 import { isHex } from '@polkadot/util';
 import { RequestHandler } from 'express';
 import { BadRequest } from 'http-errors';
+import client from 'prom-client';
 
 import { validateBoolean } from '../../middleware/validate';
 import { BlocksService } from '../../services';
@@ -124,8 +125,26 @@ export default class BlocksController extends AbstractController<BlocksService> 
 	 * @param _req Express Request
 	 * @param res Express Response
 	 */
-	private getLatestBlock: RequestHandler = async (
-		{ query: { eventDocs, extrinsicDocs, finalized, noFees, decodedXcmMsgs, paraId } },
+	private getLatestBlock: RequestHandler<
+		unknown,
+		unknown,
+		unknown,
+		{
+			finalized?: string;
+			eventDocs?: string;
+			extrinsicDocs?: string;
+			noFees?: string;
+			decodedXcmMsgs?: string;
+			paraId?: string;
+		},
+		{
+			metrics?: {
+				registry: Record<string, client.Metric>;
+				timer: () => number;
+			};
+		}
+	> = async (
+		{ query: { eventDocs, extrinsicDocs, finalized, noFees, decodedXcmMsgs, paraId }, method, route },
 		res,
 	) => {
 		const eventDocsArg = eventDocs === 'true';
@@ -150,8 +169,7 @@ export default class BlocksController extends AbstractController<BlocksService> 
 		}
 		const noFeesArg = noFees === 'true';
 		const decodedXcmMsgsArg = decodedXcmMsgs === 'true';
-		const paraIdArg =
-			paraId !== undefined ? this.parseNumberOrThrow(paraId as string, 'paraId must be an integer') : undefined;
+		const paraIdArg = paraId !== undefined ? this.parseNumberOrThrow(paraId, 'paraId must be an integer') : undefined;
 
 		const options = {
 			eventDocs: eventDocsArg,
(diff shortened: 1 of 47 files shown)


src/controllers/blocks/BlocksController.ts
```diff
@@ -18,6 +18,7 @@ import type { ApiPromise } from '@polkadot/api';
 import { isHex } from '@polkadot/util';
 import { RequestHandler } from 'express';
 import { BadRequest } from 'http-errors';
+import client from 'prom-client';
```
TarikGul, line 21: If we are using `client` here just as a type - lets do `import type client from ''prom-client`

src/controllers/blocks/BlocksController.ts
```diff
@@ -124,13 +125,30 @@ export default class BlocksController extends AbstractController<BlocksService>
 	 * @param _req Express Request
 	 * @param res Express Response
 	 */
-	private getLatestBlock: RequestHandler = async (
-		{ query: { eventDocs, extrinsicDocs, finalized, noFees, decodedXcmMsgs, paraId } },
+	private getLatestBlock: RequestHandler<
```
TarikGul, line 128: I think we can abstract this into `src/types/request.ts`

src/controllers/blocks/BlocksController.ts
```diff
@@ -151,8 +169,7 @@ export default class BlocksController extends AbstractController<BlocksService>
 		}
 		const noFeesArg = noFees === 'true';
 		const decodedXcmMsgsArg = decodedXcmMsgs === 'true';
-		const paraIdArg =
-			paraId !== undefined ? this.parseNumberOrThrow(paraId as string, 'paraId must be an integer') : undefined;
+		const paraIdArg = paraId !== undefined ? this.parseNumberOrThrow(paraId, 'paraId must be an integer') : undefined;
```
TarikGul, line 172:
```suggestion
		const paraIdArg = paraId  ? this.parseNumberOrThrow(paraId, 'paraId must be an integer') : undefined;
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -176,12 +221,34 @@ export default class BlocksController extends AbstractController<BlocksService>
 	 * @param req Express Request
 	 * @param res Express Response
 	 */
-	private getBlockById: RequestHandler<INumberParam> = async (
-		{ params: { number }, query: { eventDocs, extrinsicDocs, noFees, finalizedKey, decodedXcmMsgs, paraId } },
+	private getBlockById: RequestHandler<
```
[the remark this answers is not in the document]
TarikGul: We can also make this Generic so you don't have to write it twice:

type IBlockRequestHandler<T> = {
  T,
  unknown,
  unknown,
  BlockQueryParams,
  Metrics
}

src/controllers/blocks/BlocksController.ts
```diff
@@ -250,10 +341,18 @@ export default class BlocksController extends AbstractController<BlocksService>
 	 * @param req Express Request
 	 * @param res Express Response
 	 */
-	private getBlocks: RequestHandler<unknown, unknown, unknown, IRangeQueryParam> = async (
-		{ query: { range, eventDocs, extrinsicDocs, noFees } },
-		res,
-	): Promise<void> => {
+	private getBlocks: RequestHandler<
```
TarikGul, line 344: Same with this type

src/controllers/blocks/BlocksRawExtrinsicsController.ts
```diff
@@ -41,9 +42,32 @@ export default class BlocksRawExtrinsicsController extends AbstractController<Bl
 	 * @param _req Express Request
 	 * @param res Express Response
 	 */
-	private getBlockRawExtrinsics: RequestHandler<INumberParam> = async ({ params: { blockId } }, res): Promise<void> => {
+	private getBlockRawExtrinsics: RequestHandler<
```
TarikGul, line 45: Same with this Type

TarikGul (changes requested):
Overall, really great job. Just some nits for the types.

src/controllers/blocks/BlocksController.ts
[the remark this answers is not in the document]
filvecchiato: Created a generic type for the request with metrics that extends the RequestHandler

TarikGul (approved):
Great job, LGTM 🚀

Imod7 (approved):
LGTM! Just two small comments.

src/controllers/blocks/BlocksController.ts
```diff
@@ -167,7 +166,35 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+				.labels({ method: method, route: path, status_code: res.statusCode })
+				.observe(block.extrinsics.length);
+
+			const extrinsics_per_second = res.locals.metrics.registry['sas_extrinsics_per_second_count'] as client.Histogram;
+			extrinsics_per_second
+				.labels({ method: method, route: path, status_code: res.statusCode })
+				.observe(block.extrinsics.length / seconds);
+
+			const extrinsicsPerBlockMetrics = res.locals.metrics.registry[
+				'sas_extrinsics_per_block_count'
```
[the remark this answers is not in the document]
filvecchiato: Nice catch, I renamed them!

PR #1459, opened 2024-06-27
chore(release): 19.0.2

Imod7:
### Description
This release is a `patch` release because it includes:
- multiple `chore` related PRs
- one fix in the `staking-info` endpoint
- an update of the pjs dependencies from `v11.1.1` to `v12.0.1`

### Notes
1. The pjs-api release [v12.0.1](https://github.com/polkadot-js/api/releases/tag/v12.0.1) introduced the possibility to add the `CheckMetadataHash` signed extension and was considered a major release. However, in the Sidecar side, it is considered as a regular dependency update since the usage of endpoints remains unchanged.
2. Having said that, since this a very important addition in pjs-api I could also mention it as a Note in the release notes of Sidecar. 
3. In the `Compatibility` section of the `Changelog` (and subsequently in the release notes later), I changed the node version format. Based on the format followed previously, you would expect to see the node version as `v11300` (for polkadot-sdk `v1.13.0`) but I changed it to `v1013000`. The reason is because I noticed that in the section `Assets` of [v1.13.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.13.0) release notes, some assets are labeled as `v1013000`, e.g. `asset-hub-rococo_runtime-v1013000.compact.compressed.wasm` so I think this is the correct way to denote the client version. I will need to update the previous release notes and changelog to the correct format also.

Let me know if you have any feedback on the points mentioned above. Thank you!

PR #1478, opened 2024-08-15
chore(release): 19.1.0

Imod7:
### Description
This is a `minor` release for Sidecar since: 
- it includes features and updates in docs and dependencies and
- in the pjs and non pjs deps no breaking change was identified 

### Notes
In the compatibility section (CHANGELOG), I reverted the node version in the previous format. Also, added in parenthesis the node release in the format (vX.XX.XX) as it appears on the Polkadot [releases](https://github.com/paritytech/polkadot-sdk/releases) page to avoid any confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,15 @@ All notable changes to this project will be documented in this file. See [standa
 
 ## Compatibility
 
-Tested against:
-- Polkadot v11500 (v1.15.0 / stable2407)
-- Kusama v11500 (v1.15.0 / stable2407)
-- Westend v11500 (v1.15.0 / stable2407)
+Tested against the following node releases:
+- Polkadot v1.15.0 (Polkadot stable2407)
+- Kusama v1.15.0 (Polkadot stable2407)
+- Westend v1.15.0 (Polkadot stable2407)
+
+Tested against the following runtime releases:
+- Polkadot v1002007
+- Kusama v1002006
+- Westend v1015000
 
 ## [19.0.2](https://github.com/paritytech/substrate-api-sidecar/compare/v19.0.1..v19.0.2) (2024-06-27)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [19.1.0](https://github.com/paritytech/substrate-api-sidecar/compare/v19.0.2..v19.1.0) (2024-08-15)
+
+### Feat
+
+- feat: add pallets/on-going-referenda endpoint ([#1471](https://github.com/paritytech/substrate-api-sidecar/pull/1471)) ([386fbb3](https://github.com/paritytech/substrate-api-sidecar/commit/386fbb360b60ee92f5ca9e543532f0d0092f995f))
+- feat: Add route based metrics across API ([#1465](https://github.com/paritytech/substrate-api-sidecar/pull/1465)) ([a4bbcb8](https://github.com/paritytech/substrate-api-sidecar/commit/a4bbcb8ff2de3257fa012d832e853cedb67a0508))
+
+### Docs
+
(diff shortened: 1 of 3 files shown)


bee344 (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,32 @@
...
+
+- chore(deps): update all non polkadot-js deps in root & docs folder ([#1475](https://github.com/paritytech/substrate-api-sidecar/pull/1475)) ([c766c1d](https://github.com/paritytech/substrate-api-sidecar/commit/c766c1df8757d44aa0263b4ca296f6d43363dfe2))
+- chore: bump axios from 1.6.7 to 1.7.4 in /docs ([#1474](https://github.com/paritytech/substrate-api-sidecar/pull/1474)) ([82c61df](https://github.com/paritytech/substrate-api-sidecar/commit/82c61df511ad1a56c62f80d0ca8bb6051dfd20ec))
+- chore(deps): update polkadot-js deps ([#1473](https://github.com/paritytech/substrate-api-sidecar/pull/1473)) ([0317ec9](https://github.com/paritytech/substrate-api-sidecar/commit/0317ec9fab569c6f560db313a3aa20ef2265a5ec))
+
+
+## Compatibility
+
+Tested against:
+- Polkadot v11500 (v1.15.0 / stable2407)
```
TarikGul, line 27: If these are referring to specVersion, I thought it was `v1015000` same for the below `v1013000`

TarikGul (approved):
lgtm

CHANGELOG.md
IkerAlus: I am not sure it makes sense to talk about spec version here. These version will never hit Polkadot or Kusama runtimes. 

I think we need to be more precise and tell apart between the Polkadot-SDK node release and the versions hitting Polkadot, Kusama, and Westend RCs.

TarikGul: Yea that definitely makes sense! I agree probably best to just mention the latest live polkadot kusama and westend runtimes.

Imod7: Yes, regarding the compatibility versioning. So, as you know I changed the formatting in the last release [here](https://github.com/paritytech/substrate-api-sidecar/pull/1459) and my reasoning was the following:
> _I noticed that in the section Assets of [v1.13.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.13.0) release notes, some assets are labeled as v1013000, e.g. asset-hub-rococo_runtime-v1013000.compact.compressed.wasm so I think this is the correct way to denote the client version._

However, I had doubts now if this is the ideal way to represent the node version so I reverted back to how it used to be, so `11500`. The reasons are the following : 
- `11500` looks more compact 
- `11500` is the formatting we used until now so I thought it is best to keep it and in case of doubts I now add the version in `x.x.x` format
- `1015000` has the same formatting as the runtime and I thought that it might cause confusion.

Having said that, if you think my previous change was better then I will go back to the `v1015000` format.

IkerAlus: In terms of node compatibility, I am all good with that @Imod7 said in the last comment. 

What it is not so clear to me is what to do with runtime compatibility. When runtimes and node releases were split, we didn't make any differentiation in the release notes but I think it makes sense to mention compatibility with runtimes too. Bear in mind that runtime releases for Westend are contained in the Polkadot-SDK release, so it would be something like:

## Compatibility

- Tested against the following node releases:
  - Polkadot v1.15.0
  - Kusama v1.15.0
  - Westend v1.15.0
  
- Tested against the following runtime releases:
- Polkadot v1002007
- Kusama v1002006
- Westend v1015000

Imod7: I really like this actually. Super clear and leaves no room for confusion 💯 Thank you! I will change it now.

Imod7: Changed it [here](https://github.com/paritytech/substrate-api-sidecar/pull/1478/commits/d89f8347f3f60be1109872d9f10adedccfaf196d):
- I also `(Polkadot stable2407)` next to the node releases so it is aligned with the p-sdk releases page.
- I was thinking of making it even more clear by adding the corresponding links but then for Westend it will be confusing.

IkerAlus (approved):

PR #1483, opened 2024-08-26
fix: Improve performance of blocks service by dependency injection

filvecchiato:
It slightly improves the performance of the blocks controller by injecting the historic api to calculate the fees removing two extra promises per extrinsic.

It improves the single block query by ~15% (tested on large blocks in different chains) and also improves the response time on range queries by ~10%. On large ranges > 200 blocks, the improvement reduces as I believe it becomes more of a memory issue.

We should probably have a discussion on what to do with the range query, since if extrinsics/block will increase, that specific query will become slower, even if performance is further improved.


TarikGul (approved):
LGTM, nice improvements

Imod7 (approved):

Issue #1361, opened 2023-11-28
[Meta] Performance


TarikGul:
# Summary

Recently we have been posed with the question of how we can make Sidecar more performant. This can have multiple positive side affects such as lower latency, lower server costs, more responsive api, etc. As shown below, there are multiple avenues all of which can contribute to the performance of sidecar, some internal, and some external.

NOTE: everything I cover below will be in reference to the `/blocks/{blockId}` endpoint. A majority of the other endpoints are very performant and don't have the same issues as the blocks endpoint does.

## My Environment

Macbook pro 16inch M2 32GB memory.

I run sidecar with `--inspect` in order to profile and inspect performance.
Also running an archive node locally.

### [**Subway**](https://github.com/AcalaNetwork/subway)

Currently in beta, this library acts as an RPC middleware proxy that can cache RPC calls to the node. I did load tests on sidecar for 60s with, and without subway to demonstrate the affects. An important observation though is that I am running an archive node locally, which doesn't show the full power of `Subway` which when having to query an external archive node, it really shows the benefits of the cache.

**Without**:
```
539 requests in 1.00m, 569.18MB read
Requests/sec:      8.97
Transfer/sec:      9.47MB
--------------------------
Total completed requests:       	539
Failed requests:                	0
Timeouts:                       	0
Avg RequestTime(Latency):          475.98ms
Max RequestTime(Latency):          2722.973ms
Min RequestTime(Latency):          0.622ms
```

**With**:
```
619 requests in 1.00m, 631.96MB read
Requests/sec:     10.30
Transfer/sec:     10.52MB
--------------------------
Total completed requests:       	619
Failed requests:                	0
Timeouts:                       	0
Avg RequestTime(Latency):          449.25ms
Max RequestTime(Latency):          3353.909ms
Min RequestTime(Latency):          0.406ms
```

Overall we can see a substantial benefit from the RPC cache layer, which is increasing the output from sidecar.

### Local RPC vs non-local RPC node

- External RPC node: `/blocks/7753833 1477ms` <- Hosted in Germany
- Internal RPC node `/blocks/7753833 479ms` <- Hosted on my local machine

This is pretty obvious, therefore I didn't go to deep into this but having your server closer to your rpc node helps lower latency.

## Internal

[...]

PR #1489, opened 2024-09-13
fix: Moves the LRUcache to Controller level

filvecchiato:
This PR decreases the memory consumption produced by the LRU cache in the blocks Service by moving it to the Blocks controllers and use it only for the necessary endpoints, rather than on all fetchBlock calls. It also slightly improves response times on those endpoints by not checking the cache for every single block.

The PR also makes sure that the reference to promises in the arrays are removed by setting the array length to 0. 

Local tests have shown improvement in the Promise store in heap memory which now stays constant.

To test locally, I have run a benchmark script to hit the BlocksController with range queries. It allows to check how the endpoint behaves with large queries.

While the initial LRU Cache was increasing in size as a larger range was queried, with the current set up the cache is bypassed and used only where necessary. The current cache set up only stores the two latest queried blocks, therefore in a range query it would update for each block. 

Using a remote RPC endpoint, the following query http://127.0.0.1:8080/blocks?range=18707756-18707770 went from ~8s to ~4.5s, because we do not constantly read/write into cache

memory usage for the specific query before the change was increasing for each repetition by about 50-100MB while now it stabilises at around 60MB total. 

Partially suporting [issue](https://github.com/paritytech/substrate-api-sidecar/issues/1361)

### DIFF
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -18,6 +18,7 @@ import type { ApiPromise } from '@polkadot/api';
 import { isHex } from '@polkadot/util';
 import { RequestHandler, Response } from 'express';
 import { BadRequest } from 'http-errors';
+import type { LRUCache } from 'lru-cache';
 import type client from 'prom-client';
 
 import { validateBoolean } from '../../middleware/validate';
@@ -102,16 +103,14 @@ import AbstractController from '../AbstractController';
  * - `OnFinalize`: https://crates.parity.io/frame_support/traits/trait.OnFinalize.html
  */
 export default class BlocksController extends AbstractController<BlocksService> {
+	private blockStore: LRUCache<string, IBlock>;
 	constructor(
 		api: ApiPromise,
 		private readonly options: ControllerOptions,
 	) {
-		super(
-			api,
-			'/blocks',
-			new BlocksService(api, options.minCalcFeeRuntime, options.blockStore, options.hasQueryFeeApi),
-		);
+		super(api, '/blocks', new BlocksService(api, options.minCalcFeeRuntime, options.hasQueryFeeApi));
 		this.initRoutes();
+		this.blockStore = options.blockStore;
 	}
 
 	protected initRoutes(): void {
@@ -205,10 +204,28 @@ export default class BlocksController extends AbstractController<BlocksService> 
 			checkDecodedXcm: decodedXcmMsgsArg,
 			paraId: paraIdArg,
 		};
+		// Create a key for the cache that is a concatenation of the block hash and all the query params included in the request
+		const cacheKey =
+			hash.toString() +
+			Number(eventDocs) +
+			Number(extrinsicDocs) +
+			Number(options.checkFinalized) +
+			Number(noFees) +
+			Number(options.checkDecodedXcm) +
+			Number(paraId);
+
+		const isBlockCached = this.blockStore.get(cacheKey);
+
+		if (isBlockCached) {
+			BlocksController.sanitizedSend(res, isBlockCached);
+			return;
+		}
 
 		const historicApi = await this.api.at(hash);
 
 		const block = await this.service.fetchBlock(hash, historicApi, options);
+		this.blockStore.set(cacheKey, block);
+
 		BlocksController.sanitizedSend(res, block);
 
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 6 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -223,6 +203,7 @@ export class BlocksService extends AbstractService {
 
 		await Promise.all(feeTasks);
 
+		feeTasks.length = 0;
```
TarikGul, line 206: Whats this for?

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -62,10 +66,23 @@ export default class BlocksExtrinsicsController extends AbstractController<Block
 			paraId: undefined,
 		};
 
+		const cacheKey =
+			hash.toString() +
+			Number(eventDocs) +
```
TarikGul, line 71: For the use of `Number' here I think it would be incorrect to pass in the args as strings - but instead would need to be booleans. Example: `eventDocs` should be `eventDocsArg`

src/controllers/blocks/BlocksController.ts
```diff
@@ -263,9 +280,28 @@ export default class BlocksController extends AbstractController<BlocksService>
 			paraId: paraIdArg,
 		};
 
+		// Create a key for the cache that is a concatenation of the block hash and all the query params included in the request
+
+		const cacheKey =
+			hash.toString() +
+			Number(eventDocs) +
```
TarikGul, line 287: Same with this.

TarikGul (changes requested):
Some nits, but overall looks great!

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -62,10 +66,23 @@ export default class BlocksExtrinsicsController extends AbstractController<Block
...
+			Number(extrinsicDocs) +
```
Imod7, line 72:
```suggestion
			Number(options.extrinsicDocs) +
```
Same here. Otherwise, `Number(extrinsicDocs)` always evaluates to `NaN` in the cacheKey, whether `extrinsicDocs` is true or false.

src/controllers/blocks/BlocksExtrinsicsController.ts
Imod7: I agree with Tarik here. We should not pass a string since it will always result to `NaN`. You can also use `options.eventDocs`.

src/controllers/blocks/BlocksController.ts
Imod7: I agree with Tarik on this as well. You can also use `options.eventDocs`.

src/controllers/blocks/BlocksController.ts
```diff
@@ -263,9 +280,28 @@ export default class BlocksController extends AbstractController<BlocksService>
...
+			Number(extrinsicDocs) +
+			Number(options.checkFinalized) +
+			Number(noFees) +
```
Imod7, line 290:
```suggestion
			Number(options.noFees) +
```

src/controllers/blocks/BlocksExtrinsicsController.ts
```diff
@@ -62,10 +66,23 @@ export default class BlocksExtrinsicsController extends AbstractController<Block
...
+			Number(options.checkFinalized) +
+			Number(noFees) +
```
Imod7, line 74:
```suggestion
			Number(options.noFees) +
```

src/controllers/blocks/BlocksController.ts
```diff
@@ -205,10 +204,28 @@ export default class BlocksController extends AbstractController<BlocksService>
 			checkDecodedXcm: decodedXcmMsgsArg,
 			paraId: paraIdArg,
 		};
+		// Create a key for the cache that is a concatenation of the block hash and all the query params included in the request
+		const cacheKey =
+			hash.toString() +
+			Number(eventDocs) +
+			Number(extrinsicDocs) +
```
Imod7, line 211: I didn't check these ones but I assume we have the same issue here.

Imod7 (changes requested):
Great progress! Just some additional notes
- Is it possible to add a brief description of the local testing you performed and show how the changes in the current PR affect the performance ? Maybe some numbers or screenshots on the memory usage (before and after) or anything you find relevant.
- Can you link this PR to the performance issue it is related to ?

src/controllers/blocks/BlocksExtrinsicsController.ts
filvecchiato: updated

src/controllers/blocks/BlocksController.ts
```diff
@@ -263,9 +280,28 @@ export default class BlocksController extends AbstractController<BlocksService>
 			paraId: paraIdArg,
 		};
 
+		// Create a key for the cache that is a concatenation of the block hash and all the query params included in the request
+
+		const cacheKey =
+			hash.toString() +
+			Number(eventDocs) +
+			Number(extrinsicDocs) +
```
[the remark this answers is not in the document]
filvecchiato: I didn't link it using the keywords otherwise once this is merged the performance Issue will also be closed, quite annoying

src/controllers/blocks/BlocksExtrinsicsController.ts
Imod7, line 71:
```suggestion
			Number(options.eventDocs) +
```

src/controllers/blocks/BlocksExtrinsicsController.ts
Imod7, line 72:
```suggestion
			Number(options.extrinsicDocs) +
```

src/controllers/blocks/BlocksExtrinsicsController.ts
Imod7, line 74:
```suggestion
			Number(options.noFees) +
```

Imod7 (changes requested):

Imod7 (approved):
LCTM (Looks Cool to me!)

TarikGul (approved):

Issue #1492, opened 2024-09-14
Express 5.0 is out


TarikGul:
Express [5.0.0](https://github.com/expressjs/express/releases/tag/v5.0.0) is out and should be considered for an upgrade.

PR #1502, opened 2024-10-08
chore(deps): update express to v5 & jest deprecations

Imod7:
## Description
Closes https://github.com/paritytech/substrate-api-sidecar/issues/1492 

The changes in this PR include the following : 
- Updates `express` to [v5.0.0](https://github.com/expressjs/express/releases/tag/v5.0.0)
- Updates also `express` related types packages
- Replaces deprecated jest matchers with their native ones
- Updates the code so it builds corrently

## Migrating to Express 5
- [Moving to Express 5](https://expressjs.com/en/guide/migrating-5.html)
- [What's New in Express.js v5.0 ](https://www.trevorlasn.com/blog/whats-new-in-express-5)
- [Express.js 5: Node’s Latest Web Framework Release](https://heynode.com/blog/2020-08/expressjs-5-nodes-latest-web-framework-release/)

### Some highlights
- `res.send(obj, status)` becomes `res.status(status).send(obj)`.
- Express 5 will handle rejections from Promises returned from middleware or route handlers by passing the rejection as the first argument to next.
- The app.router object, which was removed in Express 4, has made a comeback in Express 5.

## 🧑‍💻 Code Updates 🧑‍💻

**src/App.ts**
Adjusted the syntax (basically just the brackets) of `app.get` to be as expected. I was thinking that I needed to change the syntax of `res.send(...` to `res.status(status).send(...` (based on the v5.0.0 [changes](https://expressjs.com/en/guide/migrating-5.html#res.send.body)) but it is not needed. I think it is because we do not use `status`.

**AbstractControllers.spec.ts**
It does not expect any return values 
```
Type 'Promise<string>' is not assignable to type 'Promise<void>'
```
I just changed it so it does not return the "Great success!" message and just _console.log_'s it.

**testTools.ts and util.ts**
I get a _floating-promises_ error from the linter
```
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be....
```
so I just added the `async - await` syntax to the middleware calls.

### Jest Deprecations
Replaced the following matchers : 
- toBeCalled
- toBeCalledWith
- toBeCalledTimes

with the corresponding ones : 
- toHaveBeenCalled
- toHaveBeenCalledWith
- toHaveBeenCalledTimes

based on this [PR](https://github.com/jestjs/jest/issues/13164).

## ⚠️ Important Note ⚠️
When running `yarn upgrade-interactive` it suggests as latest version the `^4.21.0` for the `express` package, even though currently the package is in `^5.0.0`. This feels a bit counterintuitive but I think it is because express 5.0.0 is not yet the default one.

## Checks
- [x] `yarn build`
- [x] `yarn lint --fix`  
- [x] `yarn test`
- [x] `yarn test:latest-e2e-tests`
- [x] `yarn test:historical-e2e-tests`


bee344 (approved):
LGTM :+1:

filvecchiato (approved):
LGTM !

PR #1506, opened 2024-10-09
test: add test for fee-estimate fix

Imod7:
### Description
Adds a test that covers the case related to the recent fix (https://github.com/paritytech/substrate-api-sidecar/pull/1505) in the `fee-estimate` endpoint. The test is mocked at block `22887036` and with Polkadot registry `v1003000`.

### How to test
Run `yarn test` or `yarn test transactionFeeEstimateService` and all tests should pass.


filvecchiato (approved):
Looking good! LGTM

src/services/transaction/TransactionFeeEstimateService.spec.ts
```diff
@@ -71,6 +105,17 @@ describe('TransactionFeeEstimateService', () => {
 			).toStrictEqual(validRuntimeResponse);
 		});
 
+		it('Works with a valid transaction at block 22887036', async () => {
```
IkerAlus, line 108: `transactionPaymentApi.queryInfo` doesn't care of the block number, it is always queried against the current runtime version. So being strict, we don't need here the detail at what block it should work.

IkerAlus (approved):

src/services/transaction/TransactionFeeEstimateService.spec.ts
Imod7: Thank you for this very good point! 
In production environment, yes you are absolutely right, the `transactionPaymentApi.queryInfo` call will work against the finalized block (related line of code [here](https://github.com/paritytech/substrate-api-sidecar/blob/68be48b715e1f53abcc91eb54ac447432715f26f/src/controllers/transaction/TransactionFeeEstimateController.ts#L73)). However in the test environment, I cannot query the live chain and get the finalized block. That is why I need to specify a block height manually and pass it as an argument which is what happens here.
In general, when writing tests in Sidecar we mock everything (or what we need). This allows us to simulate being at a specific block height using that specific type registry and metadata, while also mocking the expected response. This is also what we want from the mock tests because we want to control all these different parameters.

Issue #1508, opened 2024-10-10
v19.2.1 fails to start


BulatSaif:
**Description**

Last version fails to start

**Steps to Reproduce**

Version: docker.io/parity/substrate-api-sidecar:v19.2.1

**Expected vs. Actual Behavior**

Error:
```
2024-10-10 07:35:37 info: Listening on http://0.0.0.0:8080/
2024-10-10 07:35:37 info: Check the root endpoint (http://0.0.0.0:8080/) to see the available endpoints for the current node
2024-10-10 07:35:44 error: Cannot read properties of undefined (reading 'stack')
 TypeError: Cannot read properties of undefined (reading 'stack')
    at App.getRoutes (/usr/src/app/build/src/App.js:93:33)
    at /usr/src/app/build/src/App.js:83:30
    at Layer.handleRequest (/usr/src/app/node_modules/router/lib/layer.js:145:17)
    at next (/usr/src/app/node_modules/router/lib/route.js:159:13)
    at Route.dispatch (/usr/src/app/node_modules/router/lib/route.js:119:3)
    at handle (/usr/src/app/node_modules/router/index.js:428:11)
    at Layer.handleRequest (/usr/src/app/node_modules/router/lib/layer.js:145:17)
    at /usr/src/app/node_modules/router/index.js:293:15
    at processParams (/usr/src/app/node_modules/router/index.js:575:12)
    at next (/usr/src/app/node_modules/router/index.js:289:5)
2024-10-10 07:35:44 error: GET / 500 10ms
2024-10-10 07:35:44 error: Cannot read properties of undefined (reading 'stack')
 TypeError: Cannot read properties of undefined (reading 'stack')
    at App.getRoutes (/usr/src/app/build/src/App.js:93:33)
    at /usr/src/app/build/src/App.js:83:30
    at Layer.handleRequest (/usr/src/app/node_modules/router/lib/layer.js:145:17)
    at next (/usr/src/app/node_modules/router/lib/route.js:159:13)
    at Route.dispatch (/usr/src/app/node_modules/router/lib/route.js:119:3)
    at handle (/usr/src/app/node_modules/router/index.js:428:11)
    at Layer.handleRequest (/usr/src/app/node_modules/router/lib/layer.js:145:17)
    at /usr/src/app/node_modules/router/index.js:293:15
    at processParams (/usr/src/app/node_modules/router/index.js:575:12)
    at next (/usr/src/app/node_modules/router/index.js:289:5)
[...]

PR #1510, opened 2024-10-10
fix: how to access router in getRoutes

Imod7:
## Description
Rel https://github.com/paritytech/substrate-api-sidecar/issues/1508

Due to the upgrade in Express 5, we need to adjust how the router is accessed. 

## Test
Run Sidecar from the latest release and check  
```
http://127.0.0.1:8080
```
in your browser or with curl from your terminal


IkerAlus (approved):
Tested with a Asset Hub public endpoint, everything works as expected.

marshacb (approved):

PR #1512, opened 2024-10-14
fix: filtering in assets endpoint & update guides

Imod7:
### Issue
The endpoint `/accounts/{accountId}/asset-balances` no longer returns results for specific assets which means that the array query param format `?assets[]=1&assets[]=2&assets[]=3`does not work as expected. 
The `assets[]` array query params are also used in: 
- `/accounts/{accountId}/pool-asset-balances`
and other endpoints are using array query params and are probably affected but haven't tested them: 
- `/pallets/{palletId}/storage/{storageItemId}`
- `/contracts/ink/{address}/query`

### Suggested Solution
This issue is due to the upgrade from Express v4 to v5. In order to correctly handle `[]` in query params in `v5`, we need to set the default parser to `extended` (related [resource](https://github.com/expressjs/express/issues/5060)).

#### Test `/accounts/{accountId}/asset-balances` in Polkadot Asset Hub
You can test that this change fixes the issue when connected to Polkadot Asset Hub and querying:
`accounts/13Fet2MzBBUbJ2oREa7JJrswGeJPY4p7cY3Z2MC9oash8kzN/asset-balances?assets[]=1337`

It returns 
```
{
  "at": {
    "hash": "0xfb9b662bcc33e303349a640a85ce35044d38f50428284717db9ecf16a44bfc0a",
    "height": "7334813"
  },
  "assets": [
    {
      "assetId": "1337",
      "balance": "1260526400000",
      "isFrozen": "isFrozen does not exist for this runtime",
      "isSufficient": true
    }
  ]
}
```
which is the expected output.

#### Test `/accounts/{accountId}/pool-asset-balances` in Westend Asset Hub
You can test that this change fixes the issue when connected to Westend Asset Hub and querying:
`/accounts/5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty/pool-asset-balances?assets[]=55`

It returns 
```
{
  "at": {
    "hash": "0x3853a2bafd9d3eccec5f776bc99caf333fc4f17753f0145b961b23d1fbb89eb9",
    "height": "9357129"
  },
  "poolAssets": [
    {
      "assetId": "55",
      "balance": "0",
      "isFrozen": false,
      "isSufficient": false
    }
  ]
}
```
which is the expected output.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,10 @@
 
 1. Make sure to be in the `master` branch, and `git pull origin master`.
 
+1. Ensure that all CI/CD jobs from the last merged commit are passing. You can verify this by checking the [commits](https://github.com/paritytech/substrate-api-sidecar/commits/master/) on the master branch, where a green check (✅) should be present. If you see a red cross (❌) then you can click on it and do the following checks:
+    - If the failed job is related to benchmarks, e.g. `continuous-integration/gitlab-bench-polkadot` or `continuous-integration/gitlab-push-benchmark`, it is not critical, and you can proceed with the next steps of the release.
+    - If the failed job is related to staging deployment, e.g. `continuous-integration/gitlab-deploy-staging`, this is critical. In this case, you should check with the CI/CD team to get the relevant logs and fix the issue before continuing with the release.
+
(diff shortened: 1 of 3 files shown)


IkerAlus (approved):
LGTM!

bee344 (approved):

Issue #1487, opened 2024-09-11
Remove Rococo references after deprecation


Imod7:
**Description**
Remove all references to Rococo, including mentions in code, tests, or registry, once the deprecation is finalized.

**Expected Deprecation Date**
October 14th 2024 ([link](https://forum.polkadot.network/t/rococo-to-be-deprecated-in-october/8702/4?u=imod7))

**Examples of Rococo references in Sidecar codebase**
- [XcmDecoder.ts](https://github.com/paritytech/substrate-api-sidecar/blob/628644297ff05cceec5ec87c1c617be13d9c5ec4/src/services/blocks/XCMDecoder.ts#L54)
- [rococoRegistry.ts](https://github.com/paritytech/substrate-api-sidecar/blob/628644297ff05cceec5ec87c1c617be13d9c5ec4/src/test-helpers/registries/rococoRegistry.ts#L27)
- [rococoMetadata.ts](https://github.com/paritytech/substrate-api-sidecar/blob/628644297ff05cceec5ec87c1c617be13d9c5ec4/src/test-helpers/metadata/rococoMetadata.ts#L17)
- [ParasService.ts](https://github.com/paritytech/substrate-api-sidecar/blob/628644297ff05cceec5ec87c1c617be13d9c5ec4/src/services/paras/ParasService.ts#L53)

**Actions Required**
- Make a thorough search for all instances where Rococo is referenced/used
- Remove these references by updating the code and docs if needed

**Related Links**
Follow this forum post for updates : [Rococo to be deprecated in October](https://forum.polkadot.network/t/rococo-to-be-deprecated-in-october/8702)

PR #1517, opened 2024-10-16
fix: rococo deprecation changes

Imod7:
## Description
Closes https://github.com/paritytech/substrate-api-sidecar/issues/1487

The Rococo Test network has been deprecated/shut down (related forum [post](https://forum.polkadot.network/t/rococo-to-be-deprecated-in-october/8702/9)) so all references to Rococo in Sidecar's codebase were either removed or replaced with Paseo.

## Code Changes
- The only instance were `Rococo` was replaced with `Paseo` is in the `XCMDecoder`. You can test the change by connecting to Paseo (_SAS_SUBSTRATE_URL=wss://paseo-rpc.dwellir.com_) and querying `http://127.0.0.1:8080/blocks/3312403?decodedXcmMsgs=true`. This should returns a Horizontal and an Upward msg under `decodedXcmMsgs`.
- The tests for `Paras` endpoint are now using existing Polkadot mock data instead of Rococo mock data. The reason I did not use Paseo data is because this endpoint's info (such as auctions and crowdloans) is being phased out so it did not seem worthwhile to mock new Paseo mock data.
- In the `Paras` service, the constant `LEASE_PERIODS_PER_SLOT_FALLBACK` was removed since it is/should be present now in the metadata under [consts.auctions.leasePeriodsPerSlot](https://github.com/paritytech/substrate-api-sidecar/blob/22a143a53a20af766345d858d6c61d56dfdd2c66/src/services/paras/ParasService.ts#L302) so no fallback option is needed.


IkerAlus (approved):
LGTM

bee344 (approved):

TarikGul (approved):

PR #1519, opened 2024-10-17
feat: update dry run endpoint to use new runtime api call

filvecchiato:
It updates the transaction dry run endpoint to accept a body with the transaction string, the sender address and the block number. it returns if the dry run was successful or if it errored and for which block hash

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -1067,7 +1067,7 @@ paths:
       description: Use the dryrun call to practice submission of a transaction.
       operationId: dryrunTransaction
       requestBody:
-        $ref: '#/components/requestBodies/Transaction'
+        $ref: '#/components/requestBodies/TransactionDryRun'
       responses:
         "200":
           description: successful operation
@@ -4494,6 +4494,18 @@ components:
         tx:
           type: string
           format: hex
+    DryRunBody:
+      type: object
+      properties: 
+        at: 
+          type: string
+          format: unsignedInteger
+        tx: 
+          type: string
+          format: hex
+        senderAddress:
+          type: string
+          format: ss58
     TransactionDryRun:
       type: object
       properties:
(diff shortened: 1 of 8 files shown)


IkerAlus (commented):
The interface, inputs and expected outputs look good to me. I haven't tested this properly though.

Also, remember to update the endpoint docs accordingly: https://paritytech.github.io/substrate-api-sidecar/dist/

src/controllers/transaction/TransactionDryRunController.ts
```diff
@@ -54,13 +57,24 @@ export default class TransactionDryRunController extends AbstractController<Tran
...
-	private dryRunTransaction: IPostRequestHandler<ITx> = async ({ body: { tx }, query: { at } }, res): Promise<void> => {
+	private dryRunTransaction: IPostRequestHandler<ITx> = async (
+		{ body: { tx, at, senderAddress } },
+		res,
+	): Promise<void> => {
 		if (!tx) {
 			throw new BadRequest('Missing field `tx` on request body.');
 		}
 
+		if (!at) {
```
TarikGul, line 68: It doesn't seem like `at` needs to be required for this endpoint?

src/services/transaction/TransactionDryRunService.ts
```diff
@@ -14,43 +14,47 @@
...
 		try {
-			const [applyExtrinsicResult, { number }] = await Promise.all([
-				api.rpc.system.dryRun(transaction, hash),
+			const originCaller: SignedOriginCaller = {
+				System: {
+					Signed: senderAddress,
+				},
+			};
+
+			const [dryRunResponse, header] = await Promise.all([
```
TarikGul, line 45:
```suggestion
			const [dryRunResponse, { number }] = await Promise.all([
```
Small nit but can't this remain the same as before?

TarikGul (changes requested):
Overall looks great, just a few questions.

src/controllers/transaction/TransactionDryRunController.ts
filvecchiato: I made it not required, required some changes in the service as well

src/services/transaction/TransactionDryRunService.ts
filvecchiato: you are right, updated!

TarikGul (approved):

TarikGul:
Great job :)

Imod7 (commented):
Regarding the docs that @IkerAlus mentioned, you need to update this [file](https://github.com/paritytech/substrate-api-sidecar/blob/master/docs/src/openapi-v1.yaml) because the `dry-run` endpoint now accepts 2 additional parameters basides `tx`: `senderAddress` and `at`. 
The usual steps I follow are : 
- I check the endpoint and schema sections in that file and make any necessary updates.
- After I complete the changes, I verify that the yaml file has no errors by copy pasting it in the [Swagger Editor](https://editor.swagger.io/).
- If everything looks good, I run `yarn build:docs`
- Then I open the page `index.html` (from folder `docs/dist`) in my browser to confirm that all changes have been applied.
- If everything looks good, I push the changes.

filvecchiato:
I updated the docs and also updated the response to include response Type defining if its an error or a dispatch outcome

docs/src/openapi-v1.yaml
```diff
@@ -4504,32 +4516,20 @@ components:
...
+            paysFee:
+              type: string
+              format: boolean
+              description: Whether the transaction pays a fee.
       description: >-
         References:
-        - `UnknownTransaction`: https://crates.parity.io/sp_runtime/transaction_validity/enum.UnknownTransaction.html
-        - `InvalidTransaction`: https://crates.parity.io/sp_runtime/transaction_validity/enum.InvalidTransaction.html
+          - `PostDispatchInfo`: https://docs.rs/frame-support/38.0.0/frame_support/dispatch/struct.PostDispatchInfo.html
+          - `Error Type`: https://paritytech.github.io/polkadot-sdk/master/xcm_runtime_apis/dry_run/enum.Error.html
```
Imod7, line 4532: We could also get a [DispatchError](https://docs.rs/sp-runtime/39.0.1/sp_runtime/enum.DispatchError.html) no ? Maybe we can add this here along with an additional schema (like `TransactionDispatchOutcome` and `TransactionValidityError` _helper_ schemas) ?

Imod7 (changes requested):

Imod7 (approved):
Approving since the changes in the docs I can add in a separate PR. Thank you!

PR #1520, opened 2024-10-17
feat: improve performance with new version of PJS

filvecchiato:
The provider injected in the ApiPromise now accepts an optional cache size, which when set to zero will allow PJS to bypass caching on the requests, improving performance and decreasing memory leaks.

TODO: PJS version needs to be patched in package.json

Only merge this with the release of PJS coming in the next week or so (after 17th October 2024, possibly v13.2.2, TBD)


Imod7 (commented):
The linter needs to be run and the rest I will test as soon as pjs gets updated. Great work!

filvecchiato:
I think the linter is currently failing since the PJS/api version does not include the changes to the provider outlined in the PR

Imod7:
Just to mention for any other reviewers, I approved because: 
- If I pull the changes from master branch into this one and then `yarn build`, it looks good which was the **goal**.
- If the changes from master are not merged here, it does not build and that is **expected**.

Same applies for the linter.

TarikGul (approved):
@Imod7 Feel free to update this PR so the linter will pass :)

PR #1531, opened 2024-10-31
ci: Move from Gitlab to Github

alvicsam:
PR moves CI from gitlab to github.

Benchmarks are moved to be used with `benchmark-action/github-action-benchmark`. This is a nice tool that can save benchmark results and add graphs for historical data (like [here](https://paritytech.github.io/jsonrpsee/bench/dev2/)). I also noticed that currently github-pages use `/docs` dir in master branch. Docs publishing can be automated and docs can be moved to `gh-pages` branch so both benchmarks and docs are visible on github pages.

cc https://github.com/paritytech/ci_cd/issues/1046


AndWeHaveAPlan (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,156 @@
...
+          app_packages: "${{ env.APP }}"
+          argocd_server: ${{ env.ARGOCD_SERVER }}
+          teleport_token: ${{ env.APP }}
+          teleport_app_name: "argocd-stg"
+          argocd_auth_token: ${{ secrets.ARGOCD_AUTH_TOKEN }}
+
+  deploy-prod-kusama:
+    name: Deploy Production
+    runs-on: ubuntu-latest
+    needs: [set-variables, deploy-stg-kusama, deploy-stg-polkadot]
```
mordamax, line 115: perhaps only stg-kusama needed. not sure though

mordamax (approved):

.github/workflows/deploy.yml
alvicsam: Thanks, makes sense.

Imod7 (approved):

TarikGul (approved):
🚀

Issue #1538, opened 2024-11-06
Sidecar fails to handle multi block migration events


ermalkaleci:
**Description**
Sidecar fails to handle events generated by pallet-migrations. Sidecar tries to find extrinsic but there're no extrinsic other than inherents during migrations

`SAS_SUBSTRATE_URL=wss://rpc.shiden.astar.network npx @substrate/api-sidecar`
`curl http://localhost:8080/blocks/0xc26eb7b19378e953c80ef9749df2b5ae67a7bdb17b3ad9b9a0ee9c1810438a9f`
"Missing extrinsic 2 in block 0xc26eb7b19378e953c80ef9749df2b5ae67a7bdb17b3ad9b9a0ee9c1810438a9f"

Imod7:
@ermalkaleci Thank you for the issue.

The error you are seeing is triggered [here](https://github.com/paritytech/substrate-api-sidecar/blob/cef2d10ead615b81c11a493310c84411fd45738f/src/services/blocks/BlocksService.ts#L594) in the `BlockService` file which means that the [check](https://github.com/paritytech/substrate-api-sidecar/blob/cef2d10ead615b81c11a493310c84411fd45738f/src/services/blocks/BlocksService.ts#L589) for `phase.isApplyExtrinsic` was found `true`. 

In Sidecar, when an event is in the `ApplyExtrinsic` phase, we expect that there is an extrinsic associated to it (mentioned also in a [SE](https://substrate.stackexchange.com/a/5438/234) answer). Then, we retrieve that extrinsic by accessing the index found in [phase.asApplyExtrinsic.toNumber](https://github.com/paritytech/substrate-api-sidecar/blob/cef2d10ead615b81c11a493310c84411fd45738f/src/services/blocks/BlocksService.ts#L590). 

In the block you indicated, if I check the contents of `phase` for the event `{method: 'MigrationAdvanced', section: 'multiBlockMigrations'...}` that you mentioned, I get `{"applyExtrinsic":2}` which means you relate this event with the extrinsic in index 2. I can see the same in [Subscan](https://shiden.subscan.io/block/7675808?tab=event) :

![mbm](https://github.com/user-attachments/assets/df5eea45-af2b-4798-8027-77fb8642096d)

and if I click on the indicated extrinsic [7675808-2](https://shiden.subscan.io/extrinsic/7675808-2) I am redirected to a page not found as you would expect.

[...]

PR #1541, opened 2024-11-09
fix: ignore extrinsicIndex in multiBlockMigrations event

Imod7:
### Description
Closes https://github.com/paritytech/substrate-api-sidecar/issues/1538 

### Issue
In the `blocks` endpoint, we return a list of extrinsics included in the queried block and under each extrinsic their related event. However, during `multiBlockMigrations` we encounter cases where events reference extrinsic indexes that do not exist. 
In the code, this situation is represented by having an event with : 
- phase set as `ApplyExtrinsic` [here](https://github.com/paritytech/substrate-api-sidecar/blob/cef2d10ead615b81c11a493310c84411fd45738f/src/services/blocks/BlocksService.ts#L589) (which means we expect an extrinsic associated with this event)
- then we retrieve the extrinsic index [here](https://github.com/paritytech/substrate-api-sidecar/blob/cef2d10ead615b81c11a493310c84411fd45738f/src/services/blocks/BlocksService.ts#L590)
- the index retrieved in this case is `2` but there is no extrinsic in index `2` so we return an error msg

### Proposed Solution
During `multiBlockMigrations` : 
- we ignore the extrinsic index since it is a non existent one anyway
- we also ignore the `multiblockmigrations / MigrationAdvanced` event (shown in Subscan [here](https://shiden.subscan.io/block/7675808?tab=event)) since we cannot associate it with any valid/existing extrinsic.

In the code, we do that by : 
- **not** throwing an error if we find an event with a non existent extrinsic during `multiBlockMigrations`
- push the event only if the extrinsic is present.

This does not break any past logic since the error message will be thrown in all other cases.

### Testing
While connected to Shiden network (e.g. through `wss://rpc.shiden.astar.network`), query block 7675808 `http://127.0.0.1:8080/blocks/7675808`.

Before the fix we would get : 
```
Missing extrinsic 2 in block 0xc26eb7b19378e953c80ef9749df2b5ae67a7bdb17b3ad9b9a0ee9c1810438a9f
```

After the fix : 
```
...
...
"extrinsics": [
    {
      "method": {
        "pallet": "timestamp",
        "method": "set"
      },
     ...
     ...
      },
      "events": [
        {
          "method": {
            "pallet": "system",
            "method": "ExtrinsicSuccess"
          },
...
...
{
      "method": {
        "pallet": "parachainSystem",
        "method": "setValidationData"
      },
...
...
"events": [
        {
          "method": {
            "pallet": "system",
            "method": "ExtrinsicSuccess"
          },
```

The response now returns correctly the 2 extrinsics present in this block and their associated events. There is also the `multiblockmigrations / MigrationAdvanced` event (mentioned earlier) but we chose to ignore it since we cannot attach it to any valid/existing extrinsics.

### Downside
This solution has the downside that we are not showing an existing event.

### Long term Solution
As soon as this PR https://github.com/paritytech/polkadot-sdk/pull/3666 gets merged, I will expect that `multiBlockMigrations` events will have phase set as `ApplyInherent` so we will not have this issue for future blocks. However this fix needs to be kept for historic blocks.

### Todos
- [x] Added issue https://github.com/paritytech/substrate-api-sidecar/issues/1542 in Sidecar to track the progress of this polkadot-sdk [PR](https://github.com/paritytech/polkadot-sdk/pull/3666) and make the necessary changes as soon as it is applied.

### DIFF
--- a/src/services/blocks/BlocksService.spec.ts
+++ b/src/services/blocks/BlocksService.spec.ts
@@ -149,7 +149,7 @@ describe('BlocksService', () => {
 		});
 
 		it('throws when an extrinsic is undefined', async () => {
-			// Create a block with undefined as the first extrinisic and the last extrinsic removed
+			// Create a block with undefined as the first extrinsic and the last extrinsic removed
 			const mockBlock789629BadExt = polkadotRegistry.createType('Block', block789629);
 
 			mockBlock789629BadExt.extrinsics.pop();
@@ -377,7 +377,7 @@ describe('BlocksService', () => {
 			paraId: undefined,
 		};
 
-		it('Returns the correct extrinisics object for block 789629', async () => {
+		it('Returns the correct extrinsics object for block 789629', async () => {
 			const block = await blocksService.fetchBlock(blockHash789629, mockHistoricApi, options);
 
 			/**
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -438,7 +438,7 @@ export class BlocksService extends AbstractService {
 
 	/**
 	 * Retrieve the blockHash for the previous block to the one getting queried.
-	 * If the block is the geneisis hash it will return the same blockHash.
+	 * If the block is the genesis hash it will return the same blockHash.
 	 *
 	 * @param blockNumber The blockId being queried
 	 */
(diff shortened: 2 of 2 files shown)


IkerAlus (approved):
LGTM

PR #1552, opened 2024-11-20
ci: fix benchmark workflow, move docs to gh-pages

alvicsam:
PR fixes script for benchmarks and adds workflow for automatic publishing docs.

@TarikGul  @Imod7 docs workflow now works only if files in the `docs` folder has changed. Is it ok?

cc https://github.com/paritytech/ci_cd/issues/1086


Imod7:
Thank you @alvicsam for this follow up PR.

Regarding the docs workflow, we need it to run when : 
- the src/[openapi-v1.yaml](https://github.com/paritytech/substrate-api-sidecar/blob/master/docs/src/openapi-v1.yaml) is updated/changed
- and as soon as a PR that changes this file is merged in `master` branch

There is also the case where we update the `docs` folder by changing the `package.json` file so in this case we do not need the docs to be published again.

alvicsam:
@Imod7 

Both conditions are now in the workflow

Imod7 (approved):
Thank you for the last changes! It looks great! 
I have one last question : Since in the `gh-pages` branch under `dev/bench` we have a `data.js`, I would expect that these [graphs](https://paritytech.github.io/substrate-api-sidecar/dev/bench/) would be populated already but they are not. Is that expected ?

EDIT : Question already answered int he chat: I will check again when the benchmarks workflow runs. Thank you!

AndWeHaveAPlan (approved):

PR #1565, opened 2025-01-02
feat: Add configuration parameter for request body size

mjakuszko-punks:
I am building a solution which have quite heavy extrinsic in terms of payload, I would like to make configurable request body payload size for express middleware. Looking forward for your comments.


mjakuszko-punks:
#1565

Imod7:
> User @mjakuszko-punks, please sign the CLA [here](https://cla.parity.io/sign_cla/137152553/3).

Hey @mjakuszko-punks did you sign the CLA ?

TarikGul (approved):
LGTM, my only note would be now that the request body size is 100kb by default - this could potentially be considered a breaking change if there are existing setups that depend on a higher limit. 

Still fine with me but something to consider.

TarikGul:
> > my only note would be now that the request body size is 100kb by default - this could potentially be considered a breaking change if there are existing setups that depend on a higher limit.
> 
> My understanding from [here](https://expressjs.com/en/api.html) is that 100kb is always the default (so even in the current setup), so potentially we should be good!

Ahh perfect, thanks for clearing that up :)

mjakuszko-punks:
> > User @mjakuszko-punks, please sign the CLA [here](https://cla.parity.io/sign_cla/137152553/3).
> 
> Hey @mjakuszko-punks did you sign the CLA ?

Yes, I did it.

PR #1592, opened 2025-02-13
feat: inject controllers using metadata's definition of pallets

filvecchiato:
This PR aims to introduce a new method to inject the required/available controllers for a chain in express by using the pallets defined in the metadata of such chain. It also allows to add a config file in chains_config and then use that by setting SAS_EXPRESS_CONTROLLERS_CONFIG to true.

Each controller now requires two new static variables: controllerName and requiredPallets

ControllerName is the name of the controller to be used in the configs and requiredPallets are the pallets required in the metadata  to have the controller injected in the server.

### DIFF
--- a/src/controllers/AbstractController.ts
+++ b/src/controllers/AbstractController.ts
@@ -278,6 +278,9 @@ export default abstract class AbstractController<T extends AbstractService> {
 	}
 
 	static canInjectByPallets(availablePallets: string[]): boolean {
+		if (!this.requiredPallets) {
+			return true;
+		}
 		if (this.requiredPallets.length === 1) {
 			if (this.requiredPallets[0].length === 0) {
 				return true;
--- a/src/controllers/accounts/AccountsCompareController.ts
+++ b/src/controllers/accounts/AccountsCompareController.ts
@@ -24,6 +24,8 @@ import { ICompareQueryParams } from '../../types/requests';
 import AbstractController from '../AbstractController';
 
 export default class AccountsCompareController extends AbstractController<AccountsCompareService> {
+	static controllerName = 'AccountsCompare';
+	static requiredPallets = [];
 	constructor(api: ApiPromise) {
 		super(api, '/accounts/compare', new AccountsCompareService(api));
 		this.initRoutes();
--- a/src/controllers/controllerInjection.spec.ts
+++ b/src/controllers/controllerInjection.spec.ts
@@ -85,7 +85,7 @@ describe('controllerInjection', () => {
 			// get controllers by pallets
 			Object.values(controllers).forEach((controller) => {
 				if (controller.canInjectByPallets(pallets)) {
-					injectedControllers.add(controller.controllerName);
+					injectedControllers.add(controller.controllerName || controller.name);
 				}
 			});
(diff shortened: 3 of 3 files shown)


src/controllers/AbstractController.ts
```diff
@@ -45,11 +45,27 @@ type SidecarRequestHandler =
 	| RequestHandler<IParaIdParam>
 	| RequestHandler;
 
+/*
+ * The required pallets for a controller can be a list of pallets in string, or a definition using AND | OR and the list of pallets
+ * Example:
+ * 		- ['pallet1', 'pallet2']
+ * 		- { 'OR': ['pallet1', 'pallet2'] }
+ * 		- { 'AND': ['pallet1', 'pallet2'] }
```
IkerAlus, line 53: In what scenario is the usage of `AND` expected? Is not providing a list with every pallet enough to cover this case?

TarikGul:
Question: Don't we no longer need the chains-config `controllers` key anymore after this PR?

src/chains-config/index.ts
```diff
@@ -115,3 +115,22 @@ function getControllersFromConfig(api: ApiPromise, config: ControllerConfig) {
...
+/**
+ * Return an array of instantiated controller instances based off of a `specName`.
+ * @param pallets pallets available to define controllers
+ * @param api ApiPromise to inject into controllers
+ * @param specName specName of chain to get options
+ */
+
+export const getControllersByPallets = (pallets: string[], api: ApiPromise, specName: string) => {
+	const controllersSet: AbstractController<AbstractService>[] = [];
+	const config = specToControllerMap[specName]?.options || specToControllerMap.defaultControllers.options;
```
Imod7, line 128:
```suggestion
	const config = specToControllerMap?.[specName]?.options || specToControllerMap?.defaultControllers?.options;
```
I tested the case where: 
- I remove the `src/chains-config/....Controller.ts` file of a chain
- Started Sidecar with `SAS_EXPRESS_INJECTED_CONTROLLERS=true` for that chain
- and then it breaks here with the error 
  ```
  TypeError: Cannot read properties of undefined (reading 'options')
  ```
- I fixed this by adding these additional `optional chaining` operators.
- But then it breaks again right after with the error
  ```
  TypeError: Cannot read properties of undefined (reading 'minCalcFeeRuntime')
  ```
- So it looks for the `options` and they are also undefined.
- To fix this I added the entry `defaultControllers: defaultControllers,` in `const specToControllerMap:`
- Then it works 
- But then I was thinking also about the logic. So in this case it will get the `options` from the `defaultControllers` right? But is this correct? Because in defaultControllers, for example we have `minCalcFeeRuntime: null` and for Kusama it is `minCalcFeeRuntime: 1058`. So if I start for Kusama with injection then I get `minCalcFeeRuntime: null` and not `1058`. I do not know right now the impact of these options, I need to check more in detail but maybe we should have at least these options somewhere statically? Or maybe there is a dynamic way to retrieve/calculate them.

Imod7 (approved):
Looks great! 💯 This is such a great feature so kudos for the initiative and the excellent suggestion. Also, the implementation was super fast!

PR #1593, opened 2025-02-17
feat: add `includeClaimedRewards` query param in staking-info

Imod7:
### Description
This PR adds the `includeClaimedRewards` query param in the endpoint `/accounts/{accountId}/staking-info`. When set to `false`:
 - the field `claimedRewards` will be omitted from the response and
 - all internal calculations for claimed rewards in `AccountsStakingInfoService` will be skipped, potentially speeding up the response time. 

Defaults to `true` so it does not break the current functionality of the endpoint.

### Motivation
An exchange reported significantly long response times for the endpoint `/accounts/{accountId}/staking-info` after the release of [v20.0.0](https://github.com/paritytech/substrate-api-sidecar/releases/tag/v20.0.0). This is most likely due to the breaking changes introduced to this endpoint (calculation of `claimedRewards` changed). To address this, a new flag/query parameter has been added to improve performance when the `claimedRewards` data is not required, speeding up the response time.

### DIFF
--- a/docs/src/openapi-v1.yaml
+++ b/docs/src/openapi-v1.yaml
@@ -435,6 +435,15 @@ paths:
           type: string
           description: Block identifier, as the block height or block hash.
           format: unsignedInteger or $hex
+      - name: includeClaimedRewards
+        in: query
+        description:  When set to `false`, the `claimedRewards` field is not included
+          in the response.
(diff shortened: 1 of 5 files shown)


TarikGul (approved):
🚀

Issue #1611, opened 2025-03-09
Update staking/progress endpoint to handle breaking changes


Imod7:
**Description**
In this [PR](https://github.com/paritytech/polkadot-sdk/pull/7424) there is a change in the `UnappliedSlashes` query that needs to be updated also in Sidecar. 

It changed from `StorageMap` to `StorageDoubleMap` which needs to be updated in staking/progress endpoint [here](https://github.com/paritytech/substrate-api-sidecar/blob/eae70c15b5a48db3a3ecb7fc1c2b24ff85145534/src/services/pallets/PalletsStakingProgressService.ts#L56).

**Test**
You can test the breaking change currently in `Westend`.

**Notes**
Check the rest of the changes in the PR in case more updates need to be done in Sidecar.

PR #1612, opened 2025-03-11
fix: unappliedSlashes call in staking/progress endpoint

Imod7:
### Description
Closes https://github.com/paritytech/substrate-api-sidecar/issues/1611 which describes the issue that this PR solves.

### Testing
Without the current change in this PR, running Sidecar connected to Westend and requesting this endpoint:
```
http://127.0.0.1:8080/pallets/staking/progress?at=12718998
```

returns an empty array in `unappliedSlashes`.

After applying the current change, for the same chain and same endpoint, the response is 18 entries in `unappliedSlashes`.

### Todos
- [x] Update the code
- [x] Update current
- [x] Add a new test
- [x] Submit an issue in pjs-apps regarding the same


filvecchiato (approved):

TarikGul (approved):
👍

PR #1637, opened 2025-04-17
Fv ahm registry

filvecchiato:
Added the ApiPromise registry (first version, quite barebones), but this allows you to load an api connection from anywhere without having to inject it. 

I had to update all tests and controllers, but only to remove the injection, the rest remains the same.


TarikGul (commented):
Overall it looks amazing. I think this is exactly what we needed. Can you point me to where we can have multiple API instances though?

Like lets say we are on asset hub and it gives us the api based on that specname, how do we know to grab the RC api?

src/apiRegistry/index.ts.ts
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { ApiPromise } from '@polkadot/api';
```
TarikGul, line 2: This file name is index.ts.ts btw :)

src/main.ts
```diff
@@ -19,22 +19,53 @@
...
+/*
+ * initApis function prepares the API registry and initializes the API
+ * it returns the specName of the main API to be used by sidecar
+ */
+async function initApis(): Promise<string> {
+	const { config } = SidecarConfig;
+	const { logger } = Log;
+
+	logger.info('Initializing APIs');
+	const requiredApis = [config.SUBSTRATE.URL, config.SUBSTRATE.MULTI_CHAIN_URL].filter((el) => !!el);
```
TarikGul, line 44: What is `el` mean?

src/apiRegistry/index.ts.ts
filvecchiato: ah ups, updating it

filvecchiato:
> Overall it looks amazing. I think this is exactly what we needed. Can you point me to where we can have multiple API instances though? 

in src/main.ts, where I call initApis, i instantiate all the APIs in the registry. However I can also implement a method that instantiate the API at a later stage


> 
> Like lets say we are on asset hub and it gives us the api based on that specname, how do we know to grab the RC api?

So if we are connected on AH, and the RC is defined (either through .env or in a generic config - tbd-) you can call ApiPromiseRegistry.getApi(RCSpecName) and it would return the RC connection, directly from where you need the API. To know that we need to grab the api itself we need to configure either the controller to detect it.

src/main.ts
filvecchiato: it's the api urls available, just making sure they are not undefined

TarikGul: For `el` could we make a more well defined name. I think it would just be a bit better for readability.

src/apiRegistry/index.ts
```diff
@@ -0,0 +1,65 @@
...
+	 * Get the ApiPromise instance for the given spec name.
+	 * @param specName The spec name to get the ApiPromise instance for.
+	 * @returns The ApiPromise instance for the given spec name.
+	 */
+
+	public static async initApi(url: string): Promise<ApiPromise> {
+		const { logger } = Log;
+		logger.info(`Initializing API for ${url}`);
+
+		if (!this._instancesByUrl.has(url)) {
```
TarikGul, line 25: You can have multiple api instances to the same url - depends on what the nodes max rpc capacity is I believe!

TarikGul (approved):
👍

TarikGul (approved):
Great Changes!

PR #1663, opened 2025-06-09
linter

filvecchiato:


src/services/paras/ParasService.ts
```diff
@@ -28,7 +28,7 @@ import {
 	ParaLifecycle,
 	WinningData,
 } from '@polkadot/types/interfaces';
-import { PolkadotPrimitivesVstagingCommittedCandidateReceiptV2 } from '@polkadot/types/lookup';
+import { PolkadotPrimitivesV8CommittedCandidateReceipt } from '@polkadot/types/lookup';
```
TarikGul, line 31: I thinik this keeps coming up because of a version mismatch - with pjs specifically

filvecchiato: yes I agree, which is the correct version?

TarikGul: 16.1.2 I think

TarikGul:
Closing as I solved the main issue with the deps and conflicts in the main PR.

PR #1698, opened 2025-07-14
fix: Fixes authorId for chains using the nimbus consensus engine

RomarQ:
This PR adds a new heuristic to correctly extract the authorId for chains using the nimbus consensus engine. Previously the nimbusId was being returned instead of the author address.

It also adds an historical test to make sure regressions are not added in the future.


TarikGul:
Thanks for taking the time to put the PR together. I'll be sure to review it, tomorrow or Wednesday as bandwidth clears up :)

TarikGul (approved):
It looks great to me, thanks for the contribution!

Just a heads up, a release is planned for the second half of next week.

Issue #1693, opened 2025-07-07
Asset Hub Migration: Remaining Tasks for Single Instance Multi Chain Support


TarikGul:
# Asset Hub Migration: Remaining Tasks for Single Instance Multi Chain Support

## Overview
Complete the remaining work for Asset Hub Migration support in substrate-api-sidecar. The goal is to ensure everything is reachable from a single instance of sidecar. Currently, everything works with the exception of staking progress when using 2 instances. These tasks address historical data access and cross-chain query capabilities needed for a seamless transition to single-instance operation.

### 1. Add Historical Support for Staking Progress Endpoint

**Endpoint:** `/pallets/staking/progress?at={height}`

**Problem:** Currently lacks historical support for querying staking progress at specific block heights.

**Solution:** 
- Hardcode BABE information for epochs and sessions
- Eliminate the need for multi-chain queries by using static configuration
- Enable historical lookups without cross-chain dependencies

**Acceptance Criteria:**
- [ ] Endpoint supports `at={height}` parameter for historical queries
- [ ] BABE epoch and session data is hardcoded and accessible
- [ ] No multi-chain queries required for historical staking progress

### 2. Cross-Chain Historical Data Handling

**Problem:** When endpoints require `era` fields or need to retrieve historical data that spans the migration boundary (Asset Hub → Relay Chain), sidecar needs intelligent handling.

**Requirements:**
- Determine migration start and end points
- Map relay chain blocks to corresponding parachain block numbers
- Handle historical queries that cross the migration boundary

**Implementation Strategy:**
- Identify the exact migration start/end blocks
- Create a mapping system between relay chain and parachain block numbers
- Implement logic to route historical queries to the appropriate chain

**Acceptance Criteria:**
- [ ] Migration start/end blocks are identified and stored
- [ ] Block number mapping between relay chain and parachain is implemented
- [ ] Historical queries automatically route to correct chain based on block height
- [ ] Era-based queries work seamlessly across migration boundary

### 3. Create Mirrored Relay Chain Endpoints

**Goal:** Enable a single Asset Hub sidecar instance to access both Asset Hub and Relay Chain historical data.

**Implementation:**
- Create mirrored endpoints that proxy relay chain requests
- Allow Asset Hub sidecar to serve historical relay chain data
- Maintain API consistency while supporting dual-chain access

**Acceptance Criteria:**
- [ ] Asset Hub sidecar instance can serve relay chain historical data
[...]

PR #1718, opened 2025-07-28
fix: Ensure `staking-payouts` depth integrates with the relay chain post AHM

TarikGul:
Currently this ensures that sidecar integrates with the relay chain when connected to AH, and using a depth of greater than 1 (up to 84). 

rel: https://github.com/paritytech/substrate-api-sidecar/issues/1693

^ This is just a introduction PR - once https://github.com/paritytech/polkadot-sdk/issues/9292 is finished and added to the runtime we should add functionality that says - if the values are not hardcoded we will fetch them.

### How this works:

If the migration is finished at asset hub block 100, the era is 9, and the depth is 84. Sidecar will query the first era at 100 for staking and get the remaining 83 eras from the relay chain.


src/controllers/accounts/AccountsStakingPayoutsController.ts
```diff
@@ -154,15 +156,28 @@ export default class AccountsStakingPayoutsController extends AbstractController
...
+						hash,
+						address,
+						this.verifyAndCastOr('depth', sanitizedDepth, 1) as number,
+						eraArg,
+						unclaimedOnly === 'false' ? false : true,
+						currentEra,
+						apiAt,
+					);
+				} else {
+					result = await this.service.fetchAccountStakingPayout(
```
filvecchiato, line 171: does this require a change in the api docs?

filvecchiato (approved):
Approving as code looks good, just not sure if the docs need updating

andrew-ifrita (approved):

src/controllers/accounts/AccountsStakingPayoutsController.ts
TarikGul: Added to the docs

PR #1722, opened 2025-07-30
feat: Add transferable field to `/accounts/{accountId}/balance-info`

TarikGul:
## Add transferable balance field and documentation

  This PR adds the `transferable` field to balance-info endpoints and provides documentation explaining its calculation.

  ### Changes
  - Added `transferable` field to the `IAccountBalanceInfo` interface with detailed JSDoc documentation
  - Updated OpenAPI schema to include the `transferable` field for both balance-info endpoints:
    - `/accounts/{accountId}/balance-info`
    - `/rc/accounts/{address}/balance-info`

  ### Transferable Balance Formula
  The transferable balance uses the formula: `free - max(maybeEd, frozen - reserve)`

  Where:
  - `maybeEd` is the existential deposit if there are frozen funds or reserves, otherwise zero
  - This represents the actual spendable balance after accounting for locks, reserves, and the existential deposit
  
  NOTE: The polkadot-js implementation I am using here is a UI-friendly calculation that considers:
  
- When the account has NO reserves and NO frozen funds, the user can practically transfer their entire free balance.
- When the account HAS reserves or frozen funds, the ED must be protected

### DIFF
--- a/src/types/responses/AccountBalanceInfo.ts
+++ b/src/types/responses/AccountBalanceInfo.ts
@@ -28,6 +28,15 @@ export interface IAccountBalanceInfo {
 	miscFrozen: Balance | string;
 	feeFrozen: Balance | string;
 	frozen: Balance | string;
+	/**
+	 * Calculated transferable balance. This uses the formula: free - max(maybeEd, frozen - reserve)
+	 * Where `maybeEd` means if there is no frozen and reserves it will be zero, else it will be the existential deposit.
+	 * This is only correct when the return type of `api.query.system.account` is `FrameSystemAccountInfo`.
+	 * Which is the most up to date calculation for transferable balances.
+	 *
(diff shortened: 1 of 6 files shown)


src/types/responses/AccountBalanceInfo.ts
```diff
@@ -28,6 +28,15 @@ export interface IAccountBalanceInfo {
 	miscFrozen: Balance | string;
 	feeFrozen: Balance | string;
 	frozen: Balance | string;
+	/**
+	 * Calculated transferable balance. This uses the formula: free - max(maybeEd, frozen - reserve)
+	 * Where `maybeEd` means if there is no frozen and reserves it will be zero, else it will be the existential deposit.
```
TarikGul, line 33: This description actually needs to be adjusted. If no frozen or reserve it will be MAX(ED, frozen - reserve)

PR #1729, opened 2025-08-21
add evmFormat option to blocks queries for revive pallet

### LINKED ISSUE #1660: Hub Pallet Revive and Ethereum Accounts
**Description**

I'm not certain if this is a bug because I'm not familiar enough with Sidecar API. Maybe is a types bundle issue.

One of the previously discussed topics is enabling exchanges to receive tokens into an Ethereum-style account (and hopefully send them to an Ethereum-style account). 

As they use Sidecar, I performed a balance transfer via Metamask (Revive) and asked Sidecar for that block. In Sidecar I see the following (unfortunately, the endpoint seems not to be set in Archive mode, and so the state might be already discarded, but luckily I had this saved) Block 120655:

```
{
      "method": { "pallet": "revive", "method": "ethTransact" },
      "signature": null,
      "nonce": null,
      "args": {
        "payload": "0x02f87584190f1b45188203e88203e88505114b8b40947abce41c5153d199eafff2cd80b43adf4c71f8f688016345785d8a000080c080a01ec183188e4ea9f03b52844138f91b1015bc35336a60651e5f658e62378271b6a05ca4fb34e8cc0122483c4ea263155be444fe7e3e8b1929c65c528074fe30e844"
      },
      "tip": null,
      "hash": "0xb4810cbb429affa17bd23c6ab78de8e6720a1ee9b1176036610bac8cfb84f62a",
      "info": {},
      "era": { "immortalEra": "0x00" },
      "events": [
        {
          "method": { "pallet": "balances", "method": "Withdraw" },
          "data": [
            "5GvmgeGTbYrL2wyCM5ZzJHtvfcHLuPS33p43cPb6s3CMCYiQ",
            "44029740"
          ]
        },
        {
          "method": { "pallet": "balances", "method": "Transfer" },
          "data": [
            "5GvmgeGTbYrL2wyCM5ZzJHtvfcHLuPS33p43cPb6s3CMCYiQ",
            "5Eqdp212hzbdQjuUnx7LuQ1EwuhKr1mxWtM4XLzpwgsYJ1Kh",
            "100000000000"
          ]
        },
        {
          "method": { "pallet": "balances", "method": "Deposit" },
          "data": [
            "5EYCAe5cKPAoFh2HnQQvpKqRYZGqBpaA87u4Zzw89qPE58is",
            "44029740"
          ]
        },
```

Accounts `5GvmgeGTbYrL2wyCM5ZzJHtvfcHLuPS33p43cPb6s3CMCYiQ` and `5Eqdp212hzbdQjuUnx7LuQ1EwuhKr1mxWtM4XLzpwgsYJ1Kh` are two accounts I have in Metamask (`0xD720165D294224A7d16F22ffc6320eb31f3006e1_eeeeeeeeeeeeeeeeeeeeeeee` and `0x7AbCE41C5153d199EafFf2Cd80B43ADf4C71F8f6_eeeeeeeeeeeeeeeeeeeeeeee`). 

Ideally, there is a way to represent the Ethereum addresses in Sidecar. This would ease the integration process from exchanges to support Ethereum-style addresses.

**Steps to Reproduce**

Do a simple balance transfer in Metamask (connected to PassetHub) and fetch that block with Sidecar running against PassetHub.

Ideally we can show Ethereum addresses as is, without showing the converted version of it.

filvecchiato:
This PR aims to implement an option to convert the data included in the emitted events from pallet-revive into evm format (for addresses). The default method returns the standard SS58 format, while using the query param `useEvmFormat=true` will return the data with evm formatted addresses.

closes #1660

### DIFF
--- a/src/services/blocks/BlocksService.ts
+++ b/src/services/blocks/BlocksService.ts
@@ -38,7 +38,7 @@ import type {
 } from '@polkadot/types/interfaces';
 import type { AnyJson, Codec, Registry } from '@polkadot/types/types';
 import { u8aToHex } from '@polkadot/util';
-import { blake2AsU8a } from '@polkadot/util-crypto';
+import { blake2AsU8a, decodeAddress, isAddress } from '@polkadot/util-crypto';
 import { calc_partial_fee } from '@substrate/calc';
 import BN from 'bn.js';
 import { BadRequest, InternalServerError } from 'http-errors';
@@ -76,6 +76,7 @@ interface FetchBlockOptions {
 	noFees: boolean;
 	checkDecodedXcm: boolean;
 	paraId?: number;
+	useEvmAddressFormat?: boolean;
 }
 
 interface ExtrinsicSuccessOrFailedOverride {
@@ -119,6 +120,7 @@ export class BlocksService extends AbstractService {
 			noFees,
 			checkDecodedXcm,
 			paraId,
+			useEvmAddressFormat = false,
 		}: FetchBlockOptions,
 	): Promise<IBlock> {
 		const { api } = this;
@@ -238,6 +240,38 @@ export class BlocksService extends AbstractService {
 			finalized,
 			decodedXcmMsgs,
 		};
+
+		const convertToEvm =
+			useEvmAddressFormat &&
+			(this.api.registry.metadata.toJSON().pallets as unknown as Record<string, unknown>[])
+				.map((p) => (p.name as string).toLowerCase())
+				.includes('revive');
+
+		if (convertToEvm) {
+			// Convert SS58 addresses to EVM addresses if the revive pallet is present
+			const updatedExtrinsics = extrinsics.map((ext) => {
+				if (isFrameMethod(ext.method) && ext.method.pallet === 'revive') {
+					return {
+						...ext,
+						events: ext.events.map((event) => {
+							return {
+								...event,
+								data: event.data.map((d) => {
+									if (isAddress(d.toString())) {
+										return u8aToHex(decodeAddress(d.toString()).subarray(0, 20));
+									}
+									return d;
+								}),
+							};
+						}),
+					};
+				}
+				return ext;
+			});
+
+			response.extrinsics = updatedExtrinsics as IBlock['extrinsics'];
+		}
+
 		return response;
 	}
 
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -180,7 +180,11 @@ export default class BlocksController extends AbstractController<BlocksService> 
 	 * @param res Express Response
 	 */
 	private getLatestBlock: IRequestHandlerWithMetrics<unknown, IBlockQueryParams> = async (
-		{ query: { eventDocs, extrinsicDocs, finalized, noFees, decodedXcmMsgs, paraId, useRcBlock }, method, route },
+		{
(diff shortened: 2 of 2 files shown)


src/services/blocks/BlocksService.ts
```diff
@@ -238,6 +240,38 @@ export class BlocksService extends AbstractService {
 			finalized,
 			decodedXcmMsgs,
 		};
+
+		const convertToEvm =
+			useEvmAddressFormat &&
+			(this.api.registry.metadata.toJSON().pallets as unknown as Record<string, unknown>[])
+				.map((p) => (p.name as string).toLowerCase())
+				.includes('revive');
```
andrew-ifrita, line 248: stupid question, but why is this limited to just `revive`? It might be worth a short and concise comment in the code.

src/services/blocks/BlocksService.ts
```diff
@@ -238,6 +240,38 @@ export class BlocksService extends AbstractService {
...
+								data: event.data.map((d) => {
+									if (isAddress(d.toString())) {
+										return u8aToHex(decodeAddress(d.toString()).subarray(0, 20));
+									}
+									return d;
+								}),
+							};
+						}),
+					};
+				}
```
andrew-ifrita, line 268: Is it possible to get addresses nested in any way here where we would haveto recurse through the data instead?

andrew-ifrita (commented):
In general it looks like. Just had some minor questions.

I doubt address types will change at all, but it may be worth changing `useEvmFormat` to something like `addressFormat` or similar just to future proof in case we add another at any point in the future.

filvecchiato:
> In general it looks like. Just had some minor questions.
> 
> I doubt address types will change at all, but it may be worth changing `useEvmFormat` to something like `addressFormat` or similar just to future proof in case we add another at any point in the future.

Are you proposing to have the query param as an open-ended enum? I do prefer the boolean option tbh just because for a user its easier to remember to put true/false rather than evm.accountId32/solana etc

src/services/blocks/BlocksService.ts
filvecchiato: As far as I'm aware, currently thats the only pallet that you interact with using a evm address so should be able to retrieve data using evm addresses

src/services/blocks/BlocksService.ts
filvecchiato: not sure, but I'll modify it!

andrew-ifrita (approved):

src/services/blocks/BlocksService.ts
```diff
@@ -238,9 +240,52 @@ export class BlocksService extends AbstractService {
...
+				};
+				// }
+				// return ext;
+			});
+		}
+
 		return response;
 	}
 
+	private convertDataToEvmAddress(data: AnyJson): AnyJson {
```
TarikGul, line 269: Is this for all addresses in the block, or should we be doing addresses that are specifically tied to an event using the `revive` pallet?

src/services/blocks/BlocksService.ts
```diff
@@ -238,9 +240,52 @@ export class BlocksService extends AbstractService {
 			finalized,
 			decodedXcmMsgs,
 		};
+
+		// const convertToEvm =
```
TarikGul, line 244: Reminder to clean this up.

TarikGul (changes requested):
Just a question, and a small nit.

src/services/blocks/BlocksService.ts
filvecchiato: Reverted to pallet-revive only!

TarikGul (approved):

PR #1732, opened 2025-08-25
feat: /rc/ endpoints for pallets staking

filvecchiato:


src/controllers/controllerInjection.spec.ts
```diff
@@ -65,7 +65,7 @@ const chainsToNode: Record<string, string> = {
 	karura: 'wss://karura-rpc.dwellir.com',
 	// manta: 'wss://ws.manta.systems',
 	kilt: 'wss://kilt.ibp.network',
-	'asset-hub-polkadot': 'wss://asset-hub-polkadot-rpc.dwellir.com',
+	// 'asset-hub-polkadot': 'wss://asset-hub-polkadot-rpc.dwellir.com',
```
andrew-ifrita, line 68: Why comment this out? Was this intentional?

filvecchiato: Yes the current tests for injection controllers need updating and would not fully work with composite controllers using data from both RC and Asset Hub

TarikGul (approved):
👍 Thanks :)

TarikGul:
Let's try and get this in tomorrow so it can be included in Monday's release! Thanks for the PR @filvecchiato

PR #1740, opened 2025-08-30
add ink v6 contract dry run

BurnWW:
The current contract interface requires transferring a large ABI file with each call. I am now submitting a new dry api for more lightweight contract queries.

### DIFF
--- a/src/controllers/contracts/ContractsInkController.ts
+++ b/src/controllers/contracts/ContractsInkController.ts
@@ -15,14 +15,19 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { ContractPromise } from '@polkadot/api-contract';
+import { hexToU8a } from '@polkadot/util';
 import { RequestHandler } from 'express';
 import { BadRequest } from 'http-errors';
 
 import { validateAddress } from '../../middleware';
 import { ContractsInkService } from '../../services';
-import { IBodyContractMetadata, IContractDryParams, IContractQueryParam, IPostRequestHandler } from '../../types/requests';
+import {
+	IBodyContractMetadata,
+	IContractDryParams,
+	IContractQueryParam,
+	IPostRequestHandler,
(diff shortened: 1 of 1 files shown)


src/controllers/contracts/ContractsInkController.ts
```diff
@@ -62,4 +69,26 @@ export default class ContractsInkController extends AbstractController<Contracts
...
+	 * Send a message call to a	dry run contract. It defaults to get if nothing is inputted.
+	 *
+	 * @param _req
+	 * @param res
+	 */
+	private dryRun: IPostRequestHandler<IBodyContractMetadata, IContractDryParams> = async (
+		{ params: { address }, query: { caller, payValue = '0', inputData } },
+		res,
+	): Promise<void> => {
+		const dryRunResult = await this.api.call.reviveApi.call(
```
andrew-ifrita: This needs a check if `reviveApi` is available as well as graceful handling if it is not.

src/types/requests.ts
```diff
@@ -70,6 +70,12 @@ export interface IContractQueryParam extends Query {
 	storageDepositLimit: string;
 }
 
+export interface IContractDryParams extends Query {
+	caller: string;
+	payValue: string;
+	inputData: string;
```
andrew-ifrita: I don't see any of these values being validated anywhere which can lead to some bugs and poor UX / hard to debug issues.

src/types/requests.ts
```diff
@@ -70,6 +70,12 @@ export interface IContractQueryParam extends Query {
 	storageDepositLimit: string;
 }
 
+export interface IContractDryParams extends Query {
+	caller: string;
+	payValue: string;
```
andrew-ifrita: I think this is meant to be optional, based on the default values that is has above.

andrew-ifrita:
@BurnWW Any update here or shall we consider this stale?

PR #1742, opened 2025-09-02
docs: v2

TarikGul:


filvecchiato (changes requested):
Overall they look great! When I try to test some endpoints after selecting a public sidecar instance, and then try to generate a cUrl endpoint, but it always shows as http://localhost/endpoint_selected

filvecchiato (approved):

andrew-ifrita (approved):

Issue #1725, opened 2025-08-08
Error when trying to get balance info from account


gflow33:
**Description**
After upgrading sidecar from version `v20.4.0` to `v20.5.0`. I get an error when I try to get the balance info of an account it and it returns a TypeError.

```
OS: Linux, Ubuntu 22.04
RAM: 32gb
```
Error returned
```
{"code":500,"message":"api.rpc.chain.subscribeNewHeads is not a function","stack":"TypeError: api.rpc.chain.subscribeNewHeads is not a function\n    at /usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/api-derive/cjs/chain/bestNumber.js:15:84\n    at /usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/api-derive/cjs/chain/util.js:8:72\n    at Observable._subscribe (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/rpc-core/cjs/util/memo.js:11:30)\n    at Observable._trySubscribe (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/Observable.js:41:25)\n    at /usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/Observable.js:35:31\n    at Object.errorContext (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/util/errorContext.js:22:9)\n    at Observable.subscribe (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/Observable.js:26:24)\n    at /usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/operators/catchError.js:12:27\n    at OperatorSubscriber.<anonymous> (/usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/util/lift.js:14:28)\n    at /usr/lib/node_modules/@substrate/api-sidecar/node_modules/rxjs/dist/cjs/internal/Observable.js:30:30","level":"error"}
```
**Steps to Reproduce**
`curl -sS   -H "Accept: application/json"   "http://<IP>:8080/accounts/{account-id}/balance-info"
`

**Expected vs. Actual Behavior**
We should get the info as such
```
[...]

PR #1745, opened 2025-09-04
fix: rewrite the PJS balances derive for transferable balances

TarikGul:
### Summary

  Rewrites the PJS balances derive to properly calculate transferable balances in the AccountsBalanceInfoService, replacing the built-in derive with a custom implementation.

### Changes

  - Replaced derive calls: Removed dependency on this.api.derive.balances.all() which was unreliable
  - Added custom calculation: Implemented calculateTransferable() method that properly handles frozen/reserved balance logic
  - Added lock checking: New isAllLocked() method to determine if all funds are locked based on current block height
  - Improved accuracy: Uses existential deposit and max functions to correctly compute available transferable balance
  
  rel: https://github.com/paritytech/substrate-api-sidecar/issues/1725

### DIFF
--- a/src/services/accounts/AccountsBalanceInfoService.ts
+++ b/src/services/accounts/AccountsBalanceInfoService.ts
@@ -16,8 +16,9 @@
 
 import { ApiDecoration } from '@polkadot/api/types';
 import { Vec } from '@polkadot/types';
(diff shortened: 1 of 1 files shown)


bee344 (approved):
lgmt

PR #1753, opened 2025-09-08
fix: remove redundant isApiAssetHub specName check

TarikGul:
`isApiAssetHub` may be undefined when the multi chain input doesn't set assethubs specName. It is not required to have the specName as part of the multi chain input, but also the logic was redundant internally, and can be removed.

### DIFF
--- a/src/services/pallets/PalletsStakingProgressService.ts
+++ b/src/services/pallets/PalletsStakingProgressService.ts
@@ -34,9 +34,7 @@ export class PalletsStakingProgressService extends AbstractService {
 	 * @param hash `BlockHash` to make call at
 	 */
 	async derivePalletStakingProgress(hash: BlockHash): Promise<IPalletStakingProgress> {
-		const { api, specName } = this;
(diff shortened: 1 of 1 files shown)


filvecchiato (commented):
I'm not sure its redundant. This is to pick up if a user is querying staking from an instance running assetHub, but using the /rc endpoints, where we need to find out if its actually asset hub or not (the api instance is injected so could be relay chain and still show as asset hub from the initialisation)

TarikGul:
> I'm not sure its redundant. This is to pick up if a user is querying staking from an instance running assetHub, but using the /rc endpoints, where we need to find out if its actually asset hub or not (the api instance is injected so could be relay chain and still show as asset hub from the initialisation)

I totally get that but it breaks the non`/rc` because the `chainType` wont exist when assetHub is the main chain. `assetHubInfo` already checks if `this.api` has a specName that is an assetHub specName.

Let me know if I am missing something though. I will double check the `/rc`!

PR #1764, opened 2025-09-10
fix: added mapping to bad staking blocks in westend AH

filvecchiato:


src/services/pallets/PalletsStakingValidatorsService.ts
```diff
@@ -47,6 +48,21 @@ export class PalletsStakingValidatorsService extends AbstractService {
...
+				case 'westmint':
+					chainName = 'Westend Asset Hub';
+					break;
+				case 'statemint':
+					chainName = 'Polkadot Asset Hub';
+					break;
+			}
+
+			throw new Error(
+				`Block ${number.unwrap().toString(10)} is in the list of known bad staking blocks in ${chainName}`,
```
TarikGul, line 63: Maybe we can mention the post migration here.

PR #1776, opened 2025-09-18
feat: `/paras/{number}/inclusion`

TarikGul:
## Summary

  Add `/paras/{number}/inclusion` endpoint to find relay chain inclusion information for parachain blocks.

  ## Changes

  - **New endpoint**: `GET /paras/{number}/inclusion?depth=10`
  - **Controller**: `ParasInclusionController` with depth parameter validation
  - **Service**: `ParasInclusionService` with batched parallel search optimization
  - **Documentation**: Complete OpenAPI specs in both docs directories

  ## Performance

  - Batched parallel search (5 blocks at a time) for 50% performance improvement
  - Configurable depth parameter (must be divisible by 5, max 100)
  - Early termination when inclusion found
  
  ## Requirements
  
  - A multi chain connection is required to use this endpoint, the main connection must be asset hub, and the secondary connection the relay chain.


src/services/paras/ParasInclusionService.ts
```diff
@@ -0,0 +1,161 @@
...
+						if (record.event.section === 'paraInclusion' && record.event.method === 'CandidateIncluded') {
+							return (record.event.data[0].toJSON() as unknown as IInclusionData).descriptor.paraId === paraId;
+						}
+						return false;
+					});
+
+					const foundInclusion = inclusionEvents.find((record) => {
+						const header = rcApiAt.registry.createType('Header', record.event.data[1]);
+						return header.number.toString() === parachainBlockNumber;
+					});
```
filvecchiato, line 143: to make this more performant we could merge this filter and find functions into a reduce so that we can short circuit

filvecchiato (approved):
One minor nitpick, but overall it looks good!

TarikGul:
@filvecchiato Nice suggestion, I actually did what you mentioned but just used a `find` instead to make it a single loop

PR #1787, opened 2025-09-30
docs: fix broken description in open api spec

TarikGul:


andrew-ifrita (approved):

TarikGul:
@filvecchiato I think we should get rid of those controllerInjection tests or just keep them to reliable parity based nodes.

Issue #1790, opened 2025-10-07
Polkadot Asset Hub metadata endpoint for v14 returns error


Curu24:
**When connecting Sidecar with Polkadot Asset Hub and trying to access one of the following endpoints:**

/runtime/metadata/v14
/runtime/metadata (which defaults to v14)

`curl -X GET 'http://127.0.0.1:8080/runtime/metadata'`

the following error is returned

`{"code":500,"message":"createType(PalletUtilityCall):: Unable to create Enum via index 80, in batch, as_derivative, batch_all, dispatch_as, force_batch, with_weight, if_else, dispatch_as_fallible","stack":"Error: createType(PalletUtilityCall):: Unable to create Enum via index 80, in batch, as_derivative, batch_all, dispatch_as, force_batch, with_weight, if_else, dispatch_as_fallible\n    at createTypeUnsafe (/usr/local/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types-create/cjs/create/type.js:54:22)\n    at TypeRegistry.createType (/usr/local/lib/node_modules/@substrate/api-sidecar/node_modules/@polkadot/types/cjs/create/registry.js:224:52)\n    at sanitizeMetadataExceptionsV14 (/usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:238:52)\n    at sanitizeMetadataExceptions (/usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:194:13)\n    at /usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:56:17\n    at Array.reduce (<anonymous>)\n    at sanitizeCodec (/usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:45:30)\n    at sanitizeNumbers (/usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:125:16)\n    at /usr/local/lib/node_modules/@substrate/api-sidecar/build/src/sanitize/sanitizeNumbers.js:96:35\n    at Type.map    (<anonymous>)","level":"error"}`

Following endpoints work as expected

runtime/metadata/v15
runtime/metadata/v16

reference error on Westend - https://github.com/paritytech/substrate-api-sidecar/issues/1668

PR #1793, opened 2025-10-08
fix: add migration boundaries for kusama

TarikGul:
Adds the migration boundaries for `/ahm-info`. This ensures once the runtime is updated and the `ahMigrator`, and `rcMigrator` are removed that we still have the values.


TarikGul:
The e2e tests are related to: https://github.com/paritytech/substrate-api-sidecar/issues/1790

So its not a non issue for this PR.

Shatnerz (approved):

PR #1805, opened 2025-10-23
fix: use all params for cache key

amacar:
Requesting head block with `finalized=true` sometimes returns block with `finalized=false` property.
Adding all params to cache key should prevent this.

Maybe you can consider increasing LRU cache a bit, default is only storing 2 items.

Test script which seems to be working fine when including fix from this PR
```
SAS_SUBSTRATE_URL=wss://rpc-shadow.crust.network/
SAS_EXPRESS_PORT=8070
```

```
#!/usr/bin/env bash

set -euo pipefail

while true; do
    nonfinalized=$(curl -s "http://localhost:8070/blocks/head?finalized=false" | jq -r .finalized)
    echo "Finalized block: $nonfinalized"
    finalized=$(curl -s "http://localhost:8070/blocks/head?finalized=true" | jq -r .finalized)

    if [[ "$finalized" == "false" ]]; then
        echo "❌ Error: Block is not finalized!"
        exit 1
    elif [[ "$finalized" == "true" ]]; then
        echo "✅ Block is finalized."
    else
        echo "⚠️ Unexpected response: $finalized"
    fi

    sleep 5s
done
```

### DIFF
--- a/src/controllers/blocks/BlocksController.ts
+++ b/src/controllers/blocks/BlocksController.ts
@@ -234,15 +234,7 @@ export default class BlocksController extends AbstractController<BlocksService> 
 				};
 
 				// Create a key for the cache that is a concatenation of the block hash and all the query params included in the request
-				const cacheKey =
-					ahHash.toString() +
-					Number(options.eventDocs) +
-					Number(options.extrinsicDocs) +
-					Number(options.checkFinalized) +
-					Number(options.noFees) +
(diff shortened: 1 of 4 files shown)


andrew-ifrita (approved):
Looks good to me. Thanks for your contribution.

Let me get 1-2 more pair of eyes to check things out and then we can merge and release, hopefully.

filvecchiato (approved):
Looks good to me!

PR #1814, opened 2025-10-29
fix: add validatorSet when forceNone for /staking/progress if it exists

TarikGul:
This ensures that if there is a validator set available during a ForceNone era, we display it in the base response.


bee344 (approved):
LGTM

filvecchiato (approved):

